PR #10, opened 2024-05-28
Trigger go vet, gofmt, and goimports during the actions workflow.

wbxyz:
I also had to fix a go vet finding in our benchmark generation.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -26,3 +26,6 @@ jobs:
 
     - name: Test
       run: go test -v ./...
+
+    - name: Vet
+      run: go vet -v ./...
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -32,3 +32,6 @@ jobs:
 
     - name: Format
       run: '[ ! "$(gofmt -s -l .)" ]'
+
+    - name: Imports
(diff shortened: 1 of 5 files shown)


wbxyz:
I added the gofmt changes here too, which also includes a cleanup commit.

.github/workflows/go.yml
```diff
@@ -26,3 +26,9 @@ jobs:
 
     - name: Test
       run: go test -v ./...
+
+    - name: Vet
+      run: go vet -v ./...
+
+    - name: Format
+      run: '[ ! "$(gofmt -s -l .)" ]'
```
msuozzo, line 34: let's also use goimports which i think is just called with `-l .`

tools/benchmark/generate/main.go
```diff
@@ -723,7 +723,7 @@ func main() {
 		}
 		log.Printf("Generating %s...", b.Filename)
 		todo <- nil
-		go func() {
+		go func(b *RebuildBenchmark) {
```
msuozzo, line 726: ugh. this is no longer necessary when we raise the minimum supported go version but it's fine for now.

msuozzo (approved):

.github/workflows/go.yml
wbxyz: Added.

PR #21, opened 2024-06-03
Fix local binary rebuilding.

wbxyz:
The TUI was building binaries that were dynamically linked, and that broken when they were copied into the rebuilder container.

This was really annoying to debug because the error message didn't mention anything about how the binary was built, it just said `exec ./rebuilder: no such file or directory` even though the binary had been sucessfully copied.


.gitignore
```diff
@@ -1,3 +1,6 @@
+# Locally built binaries
```
msuozzo, line 1: i don't think we can realistically keep this in sync. i think it's preferable to set the output location for the build to something like /tmp so it doesn't create the binary in the repo at all.

wbxyz: I think the only thing we currently build is this binary, so it's fine for now? But yeah I agree that it's not great to pollute the local dir and we should eventually clean that up.

msuozzo: i'm not crazy about adding application-specific names to the gitignore. what would the process be for changing the write location for the binary?

wbxyz: I tweaked the build libraries to use a `./bin/` dir instead.

build/binary/binary.go
```diff
@@ -18,20 +18,22 @@ package binary
...
 	"log"
+	"os"
 	"os/exec"
 	"path/filepath"
 )
 
 // Build constructs a binary for one of the project's microservices.
 func Build(ctx context.Context, name string) (path string, err error) {
-	cmd := exec.CommandContext(ctx, "go", "build", "-o", name, "./cmd/"+name)
+	cmd := exec.CommandContext(ctx, "go", "build", "-o", "./bin/"+name, "./cmd/"+name)
```
msuozzo, line 28: filepath.Join?

msuozzo (approved):

build/binary/binary.go
wbxyz: Done.

msuozzo (approved):

PR #22, opened 2024-06-03
Support "edit and run local" command.

wbxyz:
This allows users to edit the existing strategy and trigger a new run.

Because I figured a user might need to re-edit the strategy a few times, the most recently edited strategy is stored in the local asset store. If "edit and run local" is invoked a second time, TUI will detect that a local strategy already exists, and it will use that as the initial contents.


tools/ctl/ide/rebuilder.go
```diff
@@ -194,15 +195,20 @@ func (rb *Rebuilder) Restart(ctx context.Context) {
...
 	if err != nil {
 		log.Println(err)
 		return
 	}
 	log.Printf("Calling the rebuilder for %s\n", r.ID())
 	id := time.Now().UTC().Format(time.RFC3339)
-	cmd := exec.CommandContext(ctx, "curl", "--silent", "-d", fmt.Sprintf("ecosystem=%s&pkg=%s&versions=%s&id=%s", r.Ecosystem, r.Package, r.Version, id), "-X", "POST", "127.0.0.1:8080/smoketest")
+	var extra string
+	if len(extraParams) > 0 {
+		extra = "&" + strings.Join(extraParams, "&")
```
msuozzo, line 209: these would need to be url-escaped depending on how you'd intend the interface to work. might be better to have this be a url.Values at the interface level.

tools/ctl/ide/rebuilder.go
```diff
@@ -194,15 +195,20 @@ func (rb *Rebuilder) Restart(ctx context.Context) {
...
+	}
+	cmd := exec.CommandContext(ctx, "curl", "--silent", "-d", fmt.Sprintf("ecosystem=%s&pkg=%s&versions=%s&id=%s%s", r.Ecosystem, r.Package, r.Version, id, extra), "-X", "POST", "127.0.0.1:8080/smoketest")
```
msuozzo, line 211: just to reiterate, would be great if we changed this to use native golang. shelling out to curl for a simple HTTP request smells a bit

tools/ctl/ide/ui.go
```diff
@@ -216,6 +227,78 @@ func (e *explorer) showLogs(ctx context.Context, example firestore.Rebuild) {
...
 }
 
+func (e *explorer) editAndRun(ctx context.Context, example firestore.Rebuild) {
+	localAssets, err := localAssetStore(ctx, example.Run)
+	if err != nil {
+		log.Println(errors.Wrap(err, "failed to create local asset store"))
+		return
+	}
+	buildDefAsset := rebuild.Asset{Type: rebuild.BuildDef, Target: example.Target()}
+	var currentStratYaml []byte
```
msuozzo, line 237: I'd change this type to be a strategyoneof. you can encode it to the asset writer below. instead of marshalling it here.

tools/ctl/ide/ui.go
```diff
@@ -216,6 +227,78 @@ func (e *explorer) showLogs(ctx context.Context, example firestore.Rebuild) {
...
+				return
+			}
+			var err error
+			if currentStratYaml, err = yaml.Marshal(oneof); err != nil {
+				log.Println(errors.Wrap(err, "failed to marshal strategy"))
+				return
+			}
+		}
+	}
+	var newStratYaml []byte
```
msuozzo, line 260: and for parity, I'd also make this a oneof so this "solicit input from user" step can fail atomically if the user inputs an invalid strat.

tools/ctl/ide/rebuilder.go
msuozzo: come to think of it, maybe taking an opts is better with the full strategyoneof as a param

tools/ctl/ide/ui.go
```diff
@@ -216,6 +227,78 @@ func (e *explorer) showLogs(ctx context.Context, example firestore.Rebuild) {
...
+		}
+	}
+	log.Println("New strategy: " + string(newStratYaml))
+	var oneof schema.StrategyOneOf
+	if err := yaml.Unmarshal(newStratYaml, &oneof); err != nil {
+		log.Println(errors.Wrap(err, "failed to parse new strategy"))
+		return
+	}
+	log.Printf("Decoded oneof: %+v", oneof)
+	newStratJsonBytes, err := json.Marshal(oneof)
```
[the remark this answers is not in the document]
wbxyz: No longer needed with the opts.

tools/ctl/ide/rebuilder.go
```diff
@@ -194,27 +197,47 @@ func (rb *Rebuilder) Restart(ctx context.Context) {
...
 	}
-	cmd := exec.CommandContext(ctx, "curl", "--silent", "-d", fmt.Sprintf("ecosystem=%s&pkg=%s&versions=%s&id=%s%s", r.Ecosystem, r.Package, r.Version, id, extra), "-X", "POST", "127.0.0.1:8080/smoketest")
-	rllog := logWriter(log.New(log.Default().Writer(), logPrefix("runlocal"), 0))
-	cmd.Stdout = rllog
-	cmd.Stderr = rllog
-	log.Println(cmd.String())
-	if err := cmd.Start(); err != nil {
-		log.Println(err)
+	client := http.DefaultClient
+	resp, err := client.PostForm("http://127.0.0.1:8080/smoketest", vals)
```
msuozzo, line 228: optional: log the URL?

tools/ctl/ide/rebuilder.go
```diff
@@ -194,27 +197,47 @@ func (rb *Rebuilder) Restart(ctx context.Context) {
...
+	if err != nil {
+		log.Println(err.Error())
+		return
+	}
+	var smkResp schema.SmoketestResponse
+	if err := json.NewDecoder(resp.Body).Decode(&smkResp); err != nil {
+		log.Println(errors.Wrap(err, "failed to decode smoketest response").Error())
+	}
+	// TODO: Is there a more usefull way to print this response?
```
msuozzo, line 237: sp: useful

tools/ctl/ide/rebuilder.go
```diff
@@ -194,27 +197,47 @@ func (rb *Rebuilder) Restart(ctx context.Context) {
...
+	log.Println(smkResp)
```
msuozzo, line 238: can we json-encode it?

tools/ctl/ide/rebuilder.go
```diff
@@ -194,27 +197,47 @@ func (rb *Rebuilder) Restart(ctx context.Context) {
...
+	if len(smkResp.Verdicts) == 0 && smkResp.Verdicts[0].Message == "" {
```
msuozzo, line 239: `len(smkResp.Verdicts) != 0` ? this predicate seems odd

msuozzo (approved):

tools/ctl/ide/rebuilder.go
wbxyz: Fixed.

tools/ctl/ide/rebuilder.go
wbxyz: It already was json encoded in the response, I was json decoding it to inspect the message but I can just keep a copy of the pre-decoded response and print that.

tools/ctl/ide/rebuilder.go
wbxyz: Whoops.

msuozzo:
maybe SUCCEEDED?

tools/ctl/ide/rebuilder.go
```diff
@@ -193,23 +198,50 @@ func (rb *Rebuilder) Restart(ctx context.Context) {
...
+		return
+	}
+	var smkRespBytes bytes.Buffer
+	var smkResp schema.SmoketestResponse
+	if err := json.NewDecoder(io.TeeReader(resp.Body, &smkRespBytes)).Decode(&smkResp); err != nil {
+		log.Println(errors.Wrap(err, "failed to decode smoketest response").Error())
+	}
+	msg := "FAILED"
+	if json.Unmarshal(smkRespBytes.Bytes(), &smkResp) == nil && len(smkResp.Verdicts) == 1 && smkResp.Verdicts[0].Message == "" {
+		msg = "SUCCESS"
```
msuozzo, line 242: maybe succeed?

wbxyz: I think it would have to be SUCCEEDED to match FAILED but that's a mouthful so I'm gonna stick with SUCCESS for now.

PR #25, opened 2024-06-05
Add the `missing` command to ctl.

wbxyz:
This command helps find targets that should have an attestation but don't. It does this by inspecting all firestore results for a RunID and querying the attestation bucket to see if they're missing.

### DIFF
--- a/tools/ctl/ctl.go
+++ b/tools/ctl/ctl.go
@@ -42,6 +42,7 @@ import (
 	"github.com/google/oss-rebuild/tools/ctl/ide"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+	"google.golang.org/api/option"
 	yaml "gopkg.in/yaml.v3"
 )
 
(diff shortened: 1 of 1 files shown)


tools/ctl/ctl.go
```diff
@@ -216,6 +217,94 @@ var getResults = &cobra.Command{
...
+			}
+			for _, r := range rebuilds {
+				if r.Success {
+					passing = append(passing, r)
+				}
+			}
+		}
+		var missing []firestore.Rebuild
+		{
+			ctx := cmd.Context()
```
msuozzo, line 248: hoist to the first line and reuse in the cmd.Context() refs above

tools/ctl/ctl.go
```diff
@@ -216,6 +217,94 @@ var getResults = &cobra.Command{
...
+			fireClient, err := firestore.NewClient(cmd.Context(), *project)
+			if err != nil {
+				log.Fatal(err)
+			}
+			rebuilds, err := fireClient.FetchRebuilds(cmd.Context(), req)
+			if err != nil {
+				log.Fatal(err)
+			}
+			for _, r := range rebuilds {
+				if r.Success {
```
msuozzo, line 241: you should be able to do this using buildFetchRebuildRequest. If you change the filter param to a ptr to string, you can provide an empty string which will select all successes.

tools/ctl/ctl.go
```diff
@@ -216,6 +217,94 @@ var getResults = &cobra.Command{
 	},
 }
 
+var missingAttestations = &cobra.Command{
```
msuozzo, line 220: As discussed, I'm dubious on this having much long-term value but I'm fine using it as a util for now. We should look to doing this diff on firestore results for smoketest vs actual rebuilds. This procedure could be used to assert consistency between the two data sources but I suspect we'd want that running as a cron and not in an adhoc CLI.

To reflect this, do you want to add a TODO to implement this diff in a command like get-result instead? diff-result?

tools/ctl/ctl.go
```diff
@@ -216,6 +217,94 @@ var getResults = &cobra.Command{
...
+	Use:   "missing -project <ID> -run <ID> -attestation-bucket <name> -format <list|bench>",
```
msuozzo, line 221: nit: maybe `find-missing`? or `find-missing-attestations`?

tools/ctl/ctl.go
```diff
@@ -504,15 +593,16 @@ var (
...
-	runFlag      = flag.String("run", "", "the run(s) from which to fetch results")
-	bench        = flag.String("bench", "", "a path to a benchmark file. if provided, only results from that benchmark will be fetched")
-	format       = flag.String("format", "summary", "the format to be printed. Options: summary, bench")
-	filter       = flag.String("filter", "", "a verdict message (or prefix) which will restrict the returned results")
-	sample       = flag.Int("sample", -1, "if provided, only N results will be displayed")
-	project      = flag.String("project", "", "the project from which to fetch the Firestore data")
-	clean        = flag.Bool("clean", false, "whether to apply normalization heuristics to group similar verdicts")
-	debugBucket  = flag.String("debug-bucket", "", "the gcs bucket to find debug logs and artifacts")
-	strategyPath = flag.String("strategy", "", "the strategy file to use")
+	runFlag           = flag.String("run", "", "the run(s) from which to fetch results")
```
msuozzo, line 596: do you want to re-organize these by command?

tools/ctl/ctl.go
```diff
@@ -216,6 +217,94 @@ var getResults = &cobra.Command{
...
+					idx = len(ps.Packages) - 1
+				}
+				ps.Packages[idx].Versions = append(ps.Packages[idx].Versions, r.Version)
+			}
+			ps.Updated = time.Now()
+			b, err := json.MarshalIndent(ps, "", "  ")
+			if err != nil {
+				log.Fatal(errors.Wrap(err, "marshalling benchmark"))
+			}
+			fmt.Println(string(b))
```
msuozzo, line 301: use the cmd output

tools/ctl/ctl.go
```diff
@@ -216,6 +217,94 @@ var getResults = &cobra.Command{
...
+				}
+				bar.Increment()
+			}
+			bar.Finish()
+		}
+		slices.SortFunc(missing, func(a firestore.Rebuild, b firestore.Rebuild) int { return strings.Compare(a.ID(), b.ID()) })
+		switch *format {
+		case "list":
+			for _, rb := range missing {
+				fmt.Println(rb.Target())
```
msuozzo, line 277: use the cmd output

tools/ctl/ctl.go
wbxyz: Currently filter is a prefix, so I don't think changing to a string ptr would be enough. We would also have to special case the empty string to be an exact match. I also don't think filtering for an empty string is as readable as selecting based on r.Success.

We could maybe change filter to be a function type instead but I think that's probably outside the scope of this change.

tools/ctl/ctl.go
wbxyz: I had missing-attestations originally but I figured that was too long. I can change it back to `find-missing-attestations`

tools/ctl/ctl.go
wbxyz: Copied from get-results, but I can change it.

PR #26, opened 2024-06-05
Fix pages github actions workflow

msuozzo:


.github/workflows/jekyll-gh-pages.yml
```diff
@@ -1,45 +1,44 @@
-# Sample workflow for building and deploying a Jekyll site to GitHub Pages
-name: Deploy Jekyll with GitHub Pages dependencies preinstalled
+name: Build Jekyll GitHub Pages site
 
 on:
-  # Runs on pushes targeting the default branch
   push:
     branches: ["main"]
```
wbxyz, line 5: Should this also be limited to certain paths?

wbxyz (approved):

.github/workflows/jekyll-gh-pages.yml
msuozzo: I considered it. I guess the only difference is that we would be rebuilding and deploying more often. I don't think there's any quota concern so it would just be a matter of preference. I'm fine either way.

msuozzo: eh i'll just add it for parity with our other workflows

PR #28, opened 2024-06-07
 Refactor python repo inference.

wbxyz:
After we added support for unknown repos, this heuristic needed some
love. I changed to a prioritized system with the following order:
1. Homepage if it's using a uri pointing to a common repo host.
1. Project links using a common "source" name, with common uri host.
1. Project links using a common "source" name, with uncommon uri host.
1. Description links using a common host uri.
1. Project links outside the common "source" name, with common uri host.

### DIFF
--- a/internal/uri/uri.go
+++ b/internal/uri/uri.go
@@ -27,11 +27,16 @@ var (
 	githubRE    = re.MustCompile(`(?i)\bgithub(\.com)?[:/]([\w-]+/[\w-\.]+)`)
 	gitlabRE    = re.MustCompile(`(?i)\bgitlab(\.com)?[:/]([\w-]+/[\w-\.]+)`)
 	bitbucketRE = re.MustCompile(`(?i)\bbitbucket(\.org)?[:/]([\w-]+/[\w-\.]+)`)
+	commonRepos = []*re.Regexp{
+		githubRE,
+		gitlabRE,
+		bitbucketRE,
+	}
 )
 
 var errUnsupportedRepo = errors.Errorf("unsupported repo type")
 
-// CanonicalizeRepoURI parses the supported repos into canonical HTTP.
+// CanonicalizeRepoURI parses repos into a canonical HTTP. This changes
 func CanonicalizeRepoURI(uri string) (string, error) {
 	if uri == "" {
 		return "", errors.New("No repo URL")
@@ -59,4 +64,19 @@ func CanonicalizeRepoURI(uri string) (string, error) {
 	}
 	u.RawQuery = ""
 	return u.String(), nil
+}
+
+// FindARepo attempts to find something that looks like a repo in the text. It will return empty string when no repo is found.
+func FindARepo(text string) string {
+	for _, pattern := range commonRepos {
+		if repo := pattern.FindString(text); repo != "" {
+			return repo
+		}
+	}
+	return ""
+}
+
+// SmellsLikeARepo returns true if the uri matches a well known URI pattern.
+func SmellsLikeARepo(uri string) bool {
+	return FindARepo(uri) != ""
 }
--- a/internal/uri/uri.go
+++ b/internal/uri/uri.go
@@ -75,8 +75,3 @@ func FindCommonRepo(text string) string {
 	}
 	return ""
 }
-
-// SmellsLikeARepo returns true if the uri matches a well known URI pattern.
-func SmellsLikeARepo(uri string) bool {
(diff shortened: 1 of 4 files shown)


pkg/rebuild/pypi/infer.go
```diff
@@ -66,14 +80,33 @@ func (Rebuilder) InferRepo(t rebuild.Target, mux rebuild.RegistryMux) (string, e
...
 		}
-		c, err := uri.CanonicalizeRepoURI(url)
-		if err == nil {
-			return c, nil
+		repoLinks = append(repoLinks, url)
+	}
+	// Four priority levels:
+	// 1. project source link | known repo
+	for _, url := range repoLinks {
+		if uri.SmellsLikeARepo(url) {
```
msuozzo, line 88: instead of having this helper function, why not just return the outcome of Find and, if non-empty, pass that to Canonicalize?

internal/uri/uri.go
```diff
@@ -60,3 +65,18 @@ func CanonicalizeRepoURI(uri string) (string, error) {
 	u.RawQuery = ""
 	return u.String(), nil
 }
+
+// FindARepo attempts to find something that looks like a repo in the text. It will return empty string when no repo is found.
+func FindARepo(text string) string {
```
msuozzo, line 70: Maybe "FindCommonRepo"?

pkg/rebuild/pypi/infer.go
```diff
@@ -66,14 +80,33 @@ func (Rebuilder) InferRepo(t rebuild.Target, mux rebuild.RegistryMux) (string, e
...
+		return uri.CanonicalizeRepoURI(repoLinks[0])
+	}
+	// 3. description         | known repo
+	r := uri.FindARepo(project.Description)
+	if r != "" && !strings.Contains(r, "sponsors") {
+		return uri.CanonicalizeRepoURI(r)
+	}
+	// 4. other project links | known repo
+	for _, url := range project.ProjectURLs {
+		// Exclude sponsor URLs.
```
msuozzo, line 103: comment not adding any additional info or context

pkg/rebuild/pypi/infer.go
```diff
@@ -41,22 +41,36 @@ import (
 
 // These are commonly used in PyPi metadata to point to the project git repo, using a map as a set.
 // Some people capitalize these differently, or add/remove spaces. We normalized to lower, no space.
+// This list is ordered, we will choose the first occurance.
```
msuozzo, line 44: this is a map and maps are not ordered :)

i guess we can switch to a slice?

pkg/rebuild/pypi/infer.go
```diff
@@ -41,40 +41,69 @@ import (
...
+		for _, commonName := range commonRepoLinks {
+			if strings.ToLower(name) == commonName {
+				repoLinks = append(repoLinks, url)
+				break
+			}
 		}
-		// Exclude sponsor URLs.
+	}
+	// Four priority levels:
+	// 1. project source link | known repo
```
msuozzo, line 83: nit: the pipe here isn't making this super clear. maybe a comma e.g. "project source link, known repo" or "AND"?

pkg/rebuild/pypi/infer.go
```diff
@@ -41,40 +41,69 @@ import (
...
+			return uri.CanonicalizeRepoURI(repo)
+		}
+	}
+	// 2. project source link | not known repo
+	if len(repoLinks) != 0 {
+		return uri.CanonicalizeRepoURI(repoLinks[0])
+	}
+	// 3. description         | known repo
+	r := uri.FindCommonRepo(project.Description)
+	if r != "" && !strings.Contains(r, "sponsors") {
```
msuozzo, line 95: want to add a TODO to revisit this "sponsors" logic? i'm not convinced enough by this as-is and we should document the intended target URLs in more detail.

tools/ctl/ctl.go
```diff
@@ -209,7 +211,97 @@ var getResults = &cobra.Command{
...
-			fmt.Println(string(b))
+			cmd.OutOrStdout().Write(b)
+			cmd.OutOrStderr().Write([]byte("\n"))
+		default:
+			log.Fatalf("Unknown --format type: %s", *format)
+		}
+	},
+}
+
+var missingAttestations = &cobra.Command{
```
msuozzo: hmm this is a different PR, right? i think this should be separable from this change.

msuozzo (approved):

PR #34, opened 2024-06-11
Support run-bench targeting local instance of rebuilder.

wbxyz:
Change the run-bench to detecting a url that looks like a cloud run url (the same way run-one does). Based on this switch, run-bench will only conditionally upgrade to https or include an oauth token.

When the `--build-local` bool flag is provided, run-bench will skip the runid creation process, and it will also skip the warmup checks.

To help with running the local builder, the user agent is now optional.

### DIFF
--- a/tools/ctl/ctl.go
+++ b/tools/ctl/ctl.go
@@ -373,7 +373,7 @@ var runBenchmark = &cobra.Command{
 						}
 						if resp.StatusCode != 200 {
 							totalErrors++
-							aggErrors = append(aggErrors, errors.Wrapf(errors.New(resp.Status), "request: %s", req.URL.String()).Error())
+							aggErrors = append(aggErrors, errors.Wrapf(errors.New(resp.Status), "requesting %s", req.URL.String()).Error())
 						}
 					}
 					bar.Increment()
--- a/tools/ctl/ctl.go
+++ b/tools/ctl/ctl.go
@@ -278,7 +278,7 @@ var runBenchmark = &cobra.Command{
 		if *buildLocal {
 			// The build-local service does not support creating a new run-id.
 			// If we're talking to build-local directly, then we skip run-id generation.
-			run = "build-local-run"
+			run = time.Now().UTC().Format(time.RFC3339)
 		} else {
 			u := apiURL.JoinPath("runs")
 			values := url.Values{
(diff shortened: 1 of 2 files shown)


tools/ctl/ctl.go
```diff
@@ -523,6 +534,7 @@ var (
 func init() {
 	runBenchmark.Flags().AddGoFlag(flag.Lookup("api"))
 	runBenchmark.Flags().AddGoFlag(flag.Lookup("max-concurrency"))
+	runBenchmark.Flags().AddGoFlag(flag.Lookup("build-local"))
```
msuozzo, line 537: what about just `--local`

tools/ctl/ctl.go
```diff
@@ -269,7 +275,11 @@ var runBenchmark = &cobra.Command{
 			log.Fatal(err)
 		}
 		var run string
-		{
+		if *buildLocal {
+			// The build-local service does not support creating a new run-id.
+			// If we're talking to build-local directly, then we skip run-id generation.
+			run = "build-local-run"
```
msuozzo, line 281: might still want to generate a timestamp-like ID

tools/ctl/ctl.go
```diff
@@ -252,11 +252,17 @@ var runBenchmark = &cobra.Command{
 		if err != nil {
 			log.Fatal(errors.Wrap(err, "parsing API endpoint"))
 		}
-		apiURL.Scheme = "https"
 		ctx := cmd.Context()
-		idclient, err := oauth.AuthorizedUserIDClient(ctx)
-		if err != nil {
-			log.Fatal(errors.Wrap(err, "creating authorized HTTP client"))
+		var idclient *http.Client
+		if strings.Contains(apiURL.Host, "run.app") {
```
msuozzo, line 257: HasSuffix? and that should probably be `.run.app`

msuozzo (approved):

msuozzo (changes requested):
Can you also format the PR titles in the imperative? -> "Support local run-bench"

tools/ctl/ctl.go
wbxyz: Resolved in https://github.com/google/oss-rebuild/pull/35

tools/ctl/ctl.go
```diff
@@ -363,7 +373,7 @@ var runBenchmark = &cobra.Command{
 						}
 						if resp.StatusCode != 200 {
 							totalErrors++
-							aggErrors = append(aggErrors, errors.Wrap(errors.New(resp.Status), "request").Error())
+							aggErrors = append(aggErrors, errors.Wrapf(errors.New(resp.Status), "request: %s", req.URL.String()).Error())
```
msuozzo, line 376: nit: "requesting %s"

msuozzo (approved):

PR #35, opened 2024-06-11
Refactor run bench to print verdicts

wbxyz:
run-bench only collected the HTTP results but didn't have any visibility when a 200 happened. This PR adds much richer logging on run-bench to keep track of the verdicts being returned.

As part of this PR I refactored the run-bench goroutines into a first class struct called a WorkerPool, and the different types of runs (smoketest vs attest) are each an implementation of a PackageWorker.

### DIFF
--- a/tools/ctl/ctl.go
+++ b/tools/ctl/ctl.go
@@ -222,13 +222,13 @@ type PackageWorker interface {
 	ProcessOne(ctx context.Context, p benchmark.Package, out chan schema.Verdict)
 }
 
-type WorkerPool struct {
-	Size      int
-	Worker    PackageWorker
-	Increment func()
+type Executor struct {
+	Concurrency int
+	Worker      PackageWorker
+	Increment   func()
 }
 
-func (pool *WorkerPool) Process(ctx context.Context, out chan schema.Verdict, packages []benchmark.Package) {
+func (ex *Executor) Process(ctx context.Context, out chan schema.Verdict, packages []benchmark.Package) {
 	jobs := make(chan benchmark.Package)
 	go func() {
 		for _, p := range packages {
@@ -237,14 +237,14 @@ func (pool *WorkerPool) Process(ctx context.Context, out chan schema.Verdict, pa
 		close(jobs)
 	}()
 	var wg sync.WaitGroup
-	for i := 0; i < pool.Size; i++ {
+	for i := 0; i < ex.Concurrency; i++ {
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
 			for p := range jobs {
-				pool.Worker.ProcessOne(ctx, p, out)
-				if pool.Increment != nil {
-					pool.Increment()
+				ex.Worker.ProcessOne(ctx, p, out)
+				if ex.Increment != nil {
+					ex.Increment()
 				}
 			}
 		}()
@@ -481,19 +481,19 @@ var runBenchmark = &cobra.Command{
 		}
 		bar := pb.StartNew(len(set.Packages))
 		bar.ShowTimeLeft = true
-		pool := WorkerPool{Size: *maxConcurrency, Increment: func() { bar.Increment() }}
+		ex := Executor{Concurrency: *maxConcurrency, Increment: func() { bar.Increment() }}
 		if mode == firestore.SmoketestMode {
-			pool.Worker = &SmoketestWorker{
+			ex.Worker = &SmoketestWorker{
 				WorkerConfig: wrkConf,
 				warmup:       urlIsCloudRun(apiURL),
 			}
 		} else {
-			pool.Worker = &AttestWorker{
+			ex.Worker = &AttestWorker{
 				WorkerConfig: wrkConf,
 			}
 		}
 		verdictChan := make(chan schema.Verdict)
-		go pool.Process(ctx, verdictChan, set.Packages)
+		go ex.Process(ctx, verdictChan, set.Packages)
 		var verdicts []schema.Verdict
 		for v := range verdictChan {
 			verdicts = append(verdicts, v)
--- a/tools/ctl/ctl.go
+++ b/tools/ctl/ctl.go
@@ -217,6 +217,47 @@ var getResults = &cobra.Command{
 	},
 }
 
+type PackageWorker interface {
+	Run(ctx context.Context, bar *pb.ProgressBar, out chan schema.Verdict, in chan benchmark.Package)
+}
+
+type WorkerPool struct {
+	Size   int
+	Worker PackageWorker
+}
+
+func (pool *WorkerPool) Process(ctx context.Context, packages []benchmark.Package) []schema.Verdict {
+	jobs := make(chan benchmark.Package, *maxConcurrency)
+	results := make(chan schema.Verdict)
+	bar := pb.StartNew(len(packages))
+	bar.ShowTimeLeft = true
+	go func() {
+		for _, p := range packages {
+			jobs <- p
+		}
+		close(jobs)
+	}()
+	var wg sync.WaitGroup
+	for i := 0; i < *maxConcurrency; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			pool.Worker.Run(ctx, bar, results, jobs)
+		}()
+	}
+	var verdicts []schema.Verdict
+	for v := range results {
+		verdicts = append(verdicts, v)
+	}
+	wg.Wait()
+	bar.Finish()
+	log.Printf("Completed rebuilds for %d artifacts...\n", len(packages))
+	sort.Slice(verdicts, func(i, j int) bool {
+		return fmt.Sprint(verdicts[i].Target) > fmt.Sprint(verdicts[j].Target)
+	})
+	return verdicts
+}
+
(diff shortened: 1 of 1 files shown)


tools/ctl/ctl.go
```diff
@@ -229,34 +271,181 @@ func makeHTTPRequest(ctx context.Context, u *url.URL, msg schema.Message) *http.
...
+			log.Fatalf("Failed to decode smoketest response: %v", err)
+		}
+		for _, v := range r.Verdicts {
+			out <- v
+		}
+	}
+}
+
+func (w *SmoketestWorker) Run(ctx context.Context, bar *pb.ProgressBar, out chan schema.Verdict, in chan benchmark.Package) {
+	if w.warmup {
```
msuozzo, line 371: I'd probably move this to a separate Init() method so then Run() can be replaced with the processOne impl (i.e. made scalar).

tools/ctl/ctl.go
```diff
@@ -229,34 +271,181 @@ func makeHTTPRequest(ctx context.Context, u *url.URL, msg schema.Message) *http.
...
+		var errMsg string
+		if err != nil {
+			errMsg = errors.Wrap(err, "sending request").Error()
+		}
+		if resp.StatusCode != 200 {
+			errMsg = errors.Wrapf(errors.New(resp.Status), "sending request").Error()
+		}
+		var verdict schema.Verdict
+		if errMsg != "" {
+			verdict = schema.Verdict{
```
msuozzo, line 296: wdyt about assining the Target above and conditionally set verdict.Message? Would cut down on the vertical space here

tools/ctl/ctl.go
```diff
@@ -216,6 +217,47 @@ var getResults = &cobra.Command{
...
+		}()
+	}
+	var verdicts []schema.Verdict
+	for v := range results {
+		verdicts = append(verdicts, v)
+	}
+	wg.Wait()
+	bar.Finish()
+	log.Printf("Completed rebuilds for %d artifacts...\n", len(packages))
+	sort.Slice(verdicts, func(i, j int) bool {
```
msuozzo: Not sure this belongs here. it'd simplify things to return a channel instead

tools/ctl/ctl.go
```diff
@@ -216,6 +217,47 @@ var getResults = &cobra.Command{
...
+
+type WorkerPool struct {
+	Size   int
+	Worker PackageWorker
+}
+
+func (pool *WorkerPool) Process(ctx context.Context, packages []benchmark.Package) []schema.Verdict {
+	jobs := make(chan benchmark.Package, *maxConcurrency)
+	results := make(chan schema.Verdict)
+	bar := pb.StartNew(len(packages))
```
msuozzo, line 232: I think the progress bar logic should be moved out of this abstraction and a callback func() be added to the struct to be called on each iteration. the bar would then be managed by the caller and updated via that callback

tools/ctl/ctl.go
wbxyz: Done

tools/ctl/ctl.go
wbxyz: Done.

tools/ctl/ctl.go
```diff
@@ -229,8 +230,55 @@ func makeHTTPRequest(ctx context.Context, u *url.URL, msg schema.Message) *http.
...
+				Message: sharedMsg,
+			})
+		}
+		return errorVerdicts
+	}
+	var decodedVerdicts []rebuild.Verdict
+	d := json.NewDecoder(resp.Body)
+	for {
+		var v rebuild.Verdict
+		if err := d.Decode(&v); err != nil {
```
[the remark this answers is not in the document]
wbxyz: I think that was outdated?

tools/ctl/ctl.go
```diff
@@ -229,8 +230,55 @@ func makeHTTPRequest(ctx context.Context, u *url.URL, msg schema.Message) *http.
...
+	// don't expect missing verdicts often though, so fatal is ok.
+	if len(decodedVerdicts) != len(t.Versions) {
+		log.Fatalf("Unexpected number of decoded verdicts for %v (got %v)", t, decodedVerdicts)
+	}
+	return decodedVerdicts
+}
+
 var runBenchmark = &cobra.Command{
-	Use:   "run-bench smoketest|attest -api <URI> <benchmark.json>",
+	Use:   "run-bench smoketest|attest -api <URI> [-build-local] <benchmark.json> [-format=csv]",
```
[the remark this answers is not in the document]
wbxyz: I think that's from a child PR?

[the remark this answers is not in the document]
wbxyz: Oh wait I thought you were just talking about build-local, but I am also adding the format flag. I can group them.

tools/ctl/ctl.go
wbxyz: Done.

tools/ctl/ctl.go
```diff
@@ -229,34 +268,172 @@ func makeHTTPRequest(ctx context.Context, u *url.URL, msg schema.Message) *http.
...
+		"pypi":  time.Tick(time.Second),
+		"npm":   time.Tick(2 * time.Second),
+		"maven": time.Tick(2 * time.Second),
+		// NOTE: cratesio needs to be especially slow given our registry API
+		// constraint of 1QPS. At minimum, we expect to make 4 calls per test.
+		"cratesio": time.Tick(8 * time.Second),
+	}
+}
+
+func urlIsCloudRun(u *url.URL) bool {
```
msuozzo, line 394: remove "url"

tools/ctl/ctl.go
```diff
@@ -216,6 +217,42 @@ var getResults = &cobra.Command{
 	},
 }
 
+type PackageWorker interface {
+	Initialize(ctx context.Context)
+	ProcessOne(ctx context.Context, p benchmark.Package, out chan schema.Verdict)
+}
+
+type WorkerPool struct {
```
msuozzo, line 225: Maybe "Executor"

msuozzo (approved):

tools/ctl/ctl.go
```diff
@@ -294,88 +472,38 @@ var runBenchmark = &cobra.Command{
 			}
 			run = string(runBytes)
 		}
-		log.Printf("Triggering rebuilds on executor version '%s' with ID=%s...\n", executor, run)
-		jobs := make(chan benchmark.Package, *maxConcurrency)
+		wrkConf := WorkerConfig{
```
msuozzo, line 475: nit: conf

tools/ctl/ctl.go
```diff
@@ -294,88 +472,38 @@ var runBenchmark = &cobra.Command{
...
-		log.Printf("Completed rebuilds for %d artifacts...\n", set.Count)
-		log.Printf("Total errors: %d\n", totalErrors)
-		for _, e := range aggErrors {
-			log.Println(e)
+		sort.Slice(verdicts, func(i, j int) bool {
+			return fmt.Sprint(verdicts[i].Target) > fmt.Sprint(verdicts[j].Target)
+		})
+		log.Printf("Triggering rebuilds on executor version '%s' with ID=%s...\n", executor, run)
+		for _, v := range verdicts {
+			fmt.Printf("%s,%s", fmt.Sprint(v.Target), v.Message)
```
msuozzo, line 506: is --format=none gonna be an option? might it be the default? we might not always want potentially thousands of lines of output, right?

msuozzo, line 506: maybe leave a TODO to support more optional (?) data? we may want other parts of the verdict here e.g. strategy.

msuozzo (approved):

tools/ctl/ctl.go
wbxyz: I added support for the flag default of "summary" which is what other usages of that flag support.

tools/ctl/ctl.go
```diff
@@ -502,8 +502,22 @@ var runBenchmark = &cobra.Command{
...
+				cmd.OutOrStdout().Write([]byte(fmt.Sprintf("%s,%s\n", v.Target, v.Message)))
+			}
+		case "summary":
+			var successes int
+			for _, v := range verdicts {
+				if v.Message == "" {
+					successes++
+				}
+			}
+			cmd.OutOrStdout().Write([]byte(fmt.Sprintf("Successes: %d/%d\n", successes, len(verdicts))))
```
msuozzo, line 518: io.WriteString here and above

tools/ctl/ctl.go
```diff
@@ -502,8 +502,22 @@ var runBenchmark = &cobra.Command{
 			return fmt.Sprint(verdicts[i].Target) > fmt.Sprint(verdicts[j].Target)
 		})
 		log.Printf("Triggering rebuilds on executor version '%s' with ID=%s...\n", executor, run)
-		for _, v := range verdicts {
-			fmt.Printf("%s,%s", fmt.Sprint(v.Target), v.Message)
+		switch *format {
+		// TODO: Maybe add more format options, or include more data in the csv?
+		case "csv":
+			for _, v := range verdicts {
+				cmd.OutOrStdout().Write([]byte(fmt.Sprintf("%s,%s\n", v.Target, v.Message)))
```
msuozzo: ugh this actually isn't safe since there are cases where Message has commas. this needs to use encoding/csv :/

PR #38, opened 2024-06-14
Fetch manual strategies from a repo

msuozzo:


tools/ctl/ctl.go
```diff
@@ -556,6 +556,9 @@ var runOne = &cobra.Command{
 		}
 		var strategy *schema.StrategyOneOf
 		if *strategyPath != "" {
+			if mode == firestore.AttestMode {
+				log.Fatal("Use --strategy-from-repo")
```
wbxyz, line 560: Maybe mention that flag is required because of the attest mode?

"attest does not support providing a strategy, use --stragey-from-repo instead"?

pkg/rebuild/schema/schema.go
```diff
@@ -225,16 +226,12 @@ func NewRebuildPackageRequest(form url.Values) (*RebuildPackageRequest, error) {
...
+	if fromRepo := form.Get("strategyFromRepo"); fromRepo != "" {
+		strategyFromRepo, err := strconv.ParseBool(fromRepo)
 		if err != nil {
 			return nil, err
 		}
-		if _, err := oneof.Strategy(); err != nil {
-			return nil, err
-		}
-		req.StrategyOneof = &oneof
+		req.StrategyFromRepo = strategyFromRepo
```
wbxyz, line 234: nit: Maybe just assign directly to req.StrategyFromRepo?

pkg/builddef/set.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/go-git/go-git/v5/storage/memory"
+	"github.com/google/oss-rebuild/pkg/rebuild/rebuild"
+	"github.com/pkg/errors"
+	"gopkg.in/yaml.v3"
+)
+
+// BuildDefinitionSet represents a collection of build definitions.
+type BuildDefinitionSet interface {
+	Get(ctx context.Context, target rebuild.Target) (rebuild.Strategy, error)
```
wbxyz, line 32: Is rebuild.Strategy sufficient for all our future build definition needs? Maybe it would make sense to have a BuildDefinition type that includes a strategy, but leaves the door open for adding other build definition info (information about who originally authored this definition and if it was manual or inferred, we also talked about the build definition having a "most recent status" field).

There are probably other fields that a build def might need that I can't think of right now, but making container with a strategy inside would make it much easier to add those things in the future, without costing much in the short term.

internal/verifier/attestation.go
```diff
@@ -136,6 +136,7 @@ func CreateAttestations(ctx context.Context, input rebuild.Input, finalStrategy
 			return nil, nil, errors.Wrap(err, "marshalling manual strategy")
 		}
 		externalParams["manualBuild"] = string(raw)
+		rd = append(rd, slsa1.ResourceDescriptor{Name: "git+" + buildDefRepo.Repo, Digest: gitDigestSet(buildDefRepo)})
```
wbxyz, line 139: Should this have a prefix to differentiate it from the other resolved "git+" descriptor (inst.Location.Repo)? IIUC both entries will be bundled in the "ResolvedDependencies" section, but there's no structural way to know which git repo is the source code vs the build def but maybe I'm missing something about the structure.

wbxyz (approved):

pkg/builddef/set.go
msuozzo: We've discussed this previously. I think it's fine for now (notably, we don't yet have any use-cases to add metadata here) and this interface could easily be updated to return the new construct if/when we migrate.

pkg/rebuild/schema/schema.go
msuozzo: it was gonna cost an extra line regardless since `err` isn't in scope. either we need a forward declaration or a re-assignment. didn't feel one was particularly better than the other.

internal/verifier/attestation.go
msuozzo: Yeah so after some research, it seems like the emerging pattern is to create a "buildConfigSource" param that includes repo/ref/path fields. I bodged together a way to get the path out of the build defn set interface so we should be able to populate that construct appropriately.

cmd/api/main.go
```diff
@@ -381,19 +388,45 @@ func HandleRebuildPackage(rw http.ResponseWriter, req *http.Request) {
...
+				ReferenceName: plumbing.Main,
+				Depth:         1,
+				NoCheckout:    true,
+			},
+			RelativePath: dir,
+			// TODO: Limit this further to only the target's path we want.
+			SparseCheckoutDirs: []string{dir},
+		})
+		if err != nil {
+			err = errors.Wrap(err, "parsing provided strategy")
```
wbxyz, line 418: Is this error message correct? IIUC err != nil means the git build def set failed to be constructed, but no strategies have been parsed yet?

cmd/api/main.go
```diff
@@ -381,19 +388,45 @@ func HandleRebuildPackage(rw http.ResponseWriter, req *http.Request) {
...
+			http.Error(rw, err.Error(), 400)
+			return
+		}
+		pth, _ := defs.Path(ctx, t)
+		buildDefLoc = rebuild.Location{
+			Repo: repo,
+			Ref:  defs.Ref().String(),
+			Dir:  pth,
+		}
+		manualStrategy, err := defs.Get(ctx, t)
```
wbxyz, line 429: Related, does this error need to be checked?

wbxyz (approved):

docs/builds/Rebuild@v0.1.md
```diff
@@ -34,12 +34,16 @@ Example:
...
+| ------------------------------ | --------------------------------------------------------------------------------------------- |
+| `ecosystem`                    | The ecosystem identifier associated with the artifact.                                        |
+| `package`                      | The package whose artifact is to be rebuilt.                                                  |
+| `version`                      | The package version whose artifact is to be rebuilt.                                          |
+| `artifact`                     | The file name of the artifact to be rebuilt.                                                  |
+| `buildConfigSource`            | The location from which the build was read. _NOTE: Only for user-generated build definitions_ |
+| `buildConfigSource.repository` | The repo URL from which the build was read.                                                   |
+| `buildConfigSource.ref`        | The repo ref from which the build was read.                                                   |
+| `buildConfigSource.path`       | The repo relpath from which the build was read.                                               |
```
wbxyz, line 46: "from which the build was read" should this be "build config" or "build def" to be more precise?

wbxyz (approved):

PR #45, opened 2024-06-25
Move handler definitions to dedicated packages.

msuozzo:


wbxyz:
Nice! I think this is a really clean improvement on the infra.

wbxyz (approved):

PR #46, opened 2024-06-25
Create a serialization lib for url.Values

msuozzo:
While the use of reflection is sub-optimal, we can end up removing a great deal of the boilerplate in our handlers and unblock the use of the upcoming rpc library.

### DIFF
--- a/pkg/rebuild/schema/schema.go
+++ b/pkg/rebuild/schema/schema.go
@@ -17,11 +17,6 @@
 package schema
 
 import (
-	"encoding/json"
-	"net/url"
-	"strconv"
-	"strings"
-
 	"github.com/google/oss-rebuild/pkg/rebuild/cratesio"
 	"github.com/google/oss-rebuild/pkg/rebuild/npm"
 	"github.com/google/oss-rebuild/pkg/rebuild/pypi"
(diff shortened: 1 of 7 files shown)


pkg/rebuild/schema/form.go
```diff
@@ -0,0 +1,112 @@
...
+			if err != nil {
+				return nil, err
+			}
+			v.Set(opt.name, string(jsonv))
+		}
+	}
+	return v, nil
+}
+
+func FromURLValues(v url.Values, out any) error {
```
wbxyz, line 74: Swap argument order to match (dest, src)?

pkg/rebuild/schema/schema.go
```diff
@@ -197,112 +140,28 @@ type SmoketestResponse struct {
...
-	vals.Set("strategyFromRepo", strconv.FormatBool(req.StrategyFromRepo))
-	return vals, nil
-}
-
 // InferenceRequest is a single request to the inference endpoint.
 type InferenceRequest struct {
-	Ecosystem    rebuild.Ecosystem
-	Package      string
-	Version      string
-	LocationHint *rebuild.LocationHint
```
wbxyz, line 254: Can we keep this limited to LocationHint for now? IIUC we don't have any guardrails to detect unsupported Strategy types in inference, and I don't think we have near-term plans to support other strategy types either.

wbxyz (approved):

pkg/rebuild/schema/form.go
msuozzo: this is kind of based off of `json.Unmarshal(reader, &out)`. in fact that might be a better name....

msuozzo: yeah ok so i actually changed this. Marshal and Unmarshal with a dedicated package gets closer to behavior that should be more familiar.

pkg/rebuild/schema/schema.go
msuozzo: Okay so I've left the logic in and have hooked it back in to the parsing logic in #47

PR #48, opened 2024-06-26
Firestore logging for RebuildRemote

wbxyz:
Add firestore logging for attestation based executions.

I split the RebuildPackage function into a handful of smaller functions. The motivation is that I wanted to capture most failure types that do an early return with an error. By pushing the "meat" of execution into these new functions, there's only one or two return values I need to capture and write to firestore, rather than lots of early return errors.

### DIFF
--- a/internal/api/apiservice/rebuild.go
+++ b/internal/api/apiservice/rebuild.go
@@ -149,7 +149,7 @@ func getStrategy(ctx context.Context, deps *RebuildPackageDeps, t rebuild.Target
 			SparseCheckoutDirs: []string{deps.BuildDefRepo.Dir},
 		})
 		if err != nil {
-			return nil, nil, api.AsStatus(codes.Internal, errors.Wrap(err, "creating build definition repo reader"))
+			return nil, nil, errors.Wrap(err, "creating build definition repo reader")
 		}
 		pth, _ := defs.Path(ctx, t)
 		rstrat = &repoStrategy{
@@ -161,7 +161,7 @@ func getStrategy(ctx context.Context, deps *RebuildPackageDeps, t rebuild.Target
 		}
 		rstrat.RepoStrategy, err = defs.Get(ctx, t)
 		if err != nil {
-			return nil, nil, api.AsStatus(codes.Internal, errors.Wrap(err, "accessing build definition"))
+			return nil, nil, errors.Wrap(err, "accessing build definition")
 		}
 		if hint, ok := rstrat.RepoStrategy.(*rebuild.LocationHint); ok && hint != nil {
 			ireq.StrategyHint = &schema.StrategyOneOf{LocationHint: hint}
@@ -173,11 +173,11 @@ func getStrategy(ctx context.Context, deps *RebuildPackageDeps, t rebuild.Target
 		s, err := deps.InferStub(ctx, ireq)
 		if err != nil {
 			// TODO: Surface better error than Internal.
-			return nil, nil, api.AsStatus(codes.Internal, errors.Wrap(err, "fetching inference"))
+			return nil, nil, errors.Wrap(err, "fetching inference")
 		}
 		strat, err = s.Strategy()
 		if err != nil {
-			return nil, nil, api.AsStatus(codes.Internal, errors.Wrap(err, "reading strategy"))
+			return nil, nil, errors.Wrap(err, "reading strategy")
 		}
 	}
 	return strat, rstrat, nil
@@ -187,7 +187,7 @@ func buildAndAttest(ctx context.Context, deps *RebuildPackageDeps, mux rebuild.R
 	id := uuid.New().String()
 	metadata, err := deps.MetadataBuilder(ctx, id)
 	if err != nil {
-		return api.AsStatus(codes.Internal, errors.Wrap(err, "creating metadata store"))
+		return errors.Wrap(err, "creating metadata store")
 	}
 	hashes := []crypto.Hash{crypto.SHA256}
 	opts := rebuild.RemoteOptions{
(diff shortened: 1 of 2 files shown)


internal/api/apiservice/rebuild.go
```diff
@@ -114,35 +120,20 @@ type RebuildPackageDeps struct {
...
-		} else if exists {
-			return nil, api.AsStatus(codes.AlreadyExists, errors.New("conflict with existing attestation bundle"))
-		}
-	}
-	var manualStrategy, strategy rebuild.Strategy
-	var buildDefLoc rebuild.Location
+type StrategyAndOrigin struct {
+	Strategy       rebuild.Strategy
+	ManualStrategy rebuild.Strategy
+	buildDefLoc    rebuild.Location
```
msuozzo, line 126: public

internal/api/apiservice/rebuild.go
```diff
@@ -114,35 +120,20 @@ type RebuildPackageDeps struct {
...
-	if !deps.OverwriteAttestations {
-		if exists, err := a.BundleExists(ctx, t); err != nil {
-			return nil, api.AsStatus(codes.Internal, errors.Wrap(err, "checking existing bundle"))
-		} else if exists {
-			return nil, api.AsStatus(codes.AlreadyExists, errors.New("conflict with existing attestation bundle"))
-		}
-	}
-	var manualStrategy, strategy rebuild.Strategy
-	var buildDefLoc rebuild.Location
+type StrategyAndOrigin struct {
```
msuozzo, line 123: non-public

internal/api/apiservice/rebuild.go
```diff
@@ -158,38 +149,41 @@ func RebuildPackage(ctx context.Context, req schema.RebuildPackageRequest, deps
...
 		}
 		pth, _ := defs.Path(ctx, t)
-		buildDefLoc = rebuild.Location{
+		strat.buildDefLoc = rebuild.Location{
 			Repo: deps.BuildDefRepo.Repo,
 			Ref:  defs.Ref().String(),
 			Dir:  pth,
 		}
-		manualStrategy, err := defs.Get(ctx, t)
+		strat.ManualStrategy, err = defs.Get(ctx, t)
```
msuozzo, line 157: oh shoot i just realized this was a poor name choice in my original code since it collides with the actual "rebuild.ManualStrategy" type. maybe "RepoStrategy"?

internal/api/apiservice/rebuild.go
msuozzo, line 123: Let's actually tweak this. We want to unconditionally return the Strategy but only conditionally return the repo-based one. What if we narrow this construct to just the manual and the Location. Then the getStrategy return type can be `(rebuild.Strategy, *repoStrategy, error)`. wdyt?

internal/api/apiservice/rebuild.go
```diff
@@ -199,38 +193,105 @@ func RebuildPackage(ctx context.Context, req schema.RebuildPackageRequest, deps
...
+	v := schema.Verdict{
+		Target: t,
+	}
+	if strat != nil {
+		v.StrategyOneof = schema.NewStrategyOneOf(strat.Strategy)
+	}
+	if rberr != nil {
+		v.Message = rberr.Error()
+	}
+	var rawStrategy string
```
msuozzo, line 274: #52 obviates this conversion. i'm fine submitting it after this one, tho

msuozzo (commented):
Nice! real close already and the separation between the firestore and non-firestore parts will make it a lot easier to test.

internal/api/apiservice/rebuild.go
wbxyz: I like that split, done.

internal/api/apiservice/rebuild.go
```diff
@@ -199,38 +193,105 @@ func RebuildPackage(ctx context.Context, req schema.RebuildPackageRequest, deps
...
+	if !deps.OverwriteAttestations {
+		if exists, err := a.BundleExists(ctx, t); err != nil {
+			return nil, api.AsStatus(codes.Internal, errors.Wrap(err, "checking existing bundle"))
+		} else if exists {
+			return nil, api.AsStatus(codes.AlreadyExists, errors.New("conflict with existing attestation bundle"))
+		}
+	}
+	strat, err := getStrategy(ctx, deps, t, req.StrategyFromRepo)
+	if err != nil {
+		return nil, api.AsStatus(codes.Internal, errors.Wrap(err, "getting strategy"))
```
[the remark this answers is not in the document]
wbxyz: Yeah I didn't intend to overwrite the lower level's status codes. Removed the AsStatus call.

internal/api/apiservice/rebuild.go
```diff
@@ -199,38 +193,105 @@ func RebuildPackage(ctx context.Context, req schema.RebuildPackageRequest, deps
...
+	}
+	err = buildAndAttest(ctx, deps, mux, a, t, *strat)
+	if err != nil {
+		return strat, api.AsStatus(codes.Internal, errors.Wrap(err, "executing rebuild"))
+	}
+	return strat, nil
+}
+
+func RebuildPackage(ctx context.Context, req schema.RebuildPackageRequest, deps *RebuildPackageDeps) (*schema.Verdict, error) {
+	t := rebuild.Target{Ecosystem: req.Ecosystem, Package: req.Package, Version: req.Version}
```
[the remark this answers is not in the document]
wbxyz: Originally I did this logic at the top level because I wanted the error to be handled differently (an error populating the artifact should prevent firestore logging because theoretically firestore should use the artifact name when creating the record).

After pushing this logic inside doRebuildPackage, I made that specific error return an error from doRebuildPackage. Otherwise all other errors get pushed into the verdict being returned.

internal/api/apiservice/rebuild.go
msuozzo: let's submit this first

msuozzo (approved):

PR #50, opened 2024-06-27
Add ManualStrategy to oneof.

msuozzo:


wbxyz:
I think you also need to update NewStrategyOneOf() to handle ManualStrategy types.

msuozzo:
Yep good catch. Done

wbxyz (approved):

PR #51, opened 2024-06-27
Add test library for firestore emulator.

msuozzo:
No current plans about how/whether to hook this into our existing testing since it requires a `gcloud` installation with the firestore emulator component. For now, we can use it for local testing/experimentation with firestore.

### DIFF
--- a/internal/firestoretest/emulator.go
+++ b/internal/firestoretest/emulator.go
@@ -1,0 +1,108 @@
+package firestoretest
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"net/http"
+	"os"
+	"os/exec"
+	"testing"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+func getPort() (*int, error) {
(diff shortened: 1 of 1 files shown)


internal/firestoretest/emulator.go
```diff
@@ -0,0 +1,103 @@
...
+		go func() {
+			for {
+				c, err := net.DialTCP("tcp", nil, &net.TCPAddr{Port: *port})
+				if err == nil {
+					c.Close()
+					close(portReachable)
+					return
+				}
+				time.Sleep(300 * time.Millisecond)
+			}
```
wbxyz, line 68: If firestore emulator fails to start, will this goroutine ever be cleaned up? Should this also attempt to wait for ctx.Done()?

wbxyz (approved):

internal/firestoretest/emulator.go
msuozzo: sure although in that case, the program's gonna exit immediately so i see very little value in it

PR #52, opened 2024-06-28
Use native types in firestore schema.

msuozzo:


pkg/rebuild/schema/schema.go
```diff
@@ -222,18 +222,15 @@ type CreateRunResponse struct {
...
-	RunID             string  `firestore:"run_id,omitempty"`
-	Created           int64   `firestore:"created,omitempty"`
+	Ecosystem       string          `firestore:"ecosystem,omitempty"`
+	Package         string          `firestore:"package,omitempty"`
+	Version         string          `firestore:"version,omitempty"`
+	Artifact        string          `firestore:"artifact,omitempty"`
+	Success         bool            `firestore:"success,omitempty"`
+	Message         string          `firestore:"message,omitempty"`
+	Strategy        StrategyOneOf   `firestore:"strategy,omitempty"`
+	Timings         rebuild.Timings `firestore:"timings,omitempty"`
```
wbxyz, line 232: Will this break backwards compatibility? IIUC currently we store things in TimeSource TimeInfer and TimeBuild fields, but I think this change might push all of those into a single Timings field. And maybe this will fail to read timings from old firestore entries?

It's probably fine to break backwards compatibility, just wondering if you have thoughts.

msuozzo: Yeah this will break the existing fields. we could add the existing ones back to support the current data but I wasn't sure how important that was. i can hack it together and see how it looks.

wbxyz: Discussed offline, I think the latest version will ignore old formats gracefully, leaving the new fields empty (timings and strategy).

tools/ctl/ide/ui.go
```diff
@@ -180,10 +173,10 @@ func (e *explorer) showDetails(ctx context.Context, example firestore.Rebuild) {
...
 	enc.SetIndent(2)
 	err := enc.Encode(detailsStruct{
-		Success:  example.Success,
-		Message:  example.Message,
-		Timings:  example.Timings,
-		Strategy: stratOneof,
+		Success: example.Success,
+		Message: example.Message,
+		Timings: example.Timings,
+		//Strategy: example.Strategy,
```
wbxyz, line 179: Does this still need to be commented out?

wbxyz (approved):

PR #58, opened 2024-07-01
Split metadata bucket.

wbxyz:
Previously the metadata bucket was used for both the rebuilt artifacts and also the build info/docker file.

This gave cloud build unnecessary access to both the build info and the dockerfile which are both written and read by the orchestrator.

After this PR, cloud build only has Create access to the newly created rebuild-artifacts bucket. That bucket only stores:
  1. The rebuilt artifact.
  2. The docker image used for the build.

Further, cloud build no longer has access to the metadata bucket, which contains the build info and docker file. Because it's only used locally to the orchestrator, I considered switching it to a filesystem or in-memory asset store, except we want these assets available for debugging so I'm keeping it as a bucket for now.

One interesting affect of this PR, is that `verifier.SummarizeArtifacts` only requires `rebuildStore` and `verifier.CreateAttestations` only requires the metadata store, clearly distinguishing where each of those steps get their input.

This further reduces the already tightly restricted permissions afforded the the cloud build role.

### DIFF
--- a/cmd/api/main.go
+++ b/cmd/api/main.go
@@ -48,7 +48,7 @@ var (
 	buildLocalURL         = flag.String("build-local-url", "", "URL of the rebuild service")
 	inferenceURL          = flag.String("inference-url", "", "URL of the inference service")
 	signingKeyVersion     = flag.String("signing-key-version", "", "Resource name of the signing CryptoKeyVersion")
-	rebuildBucket         = flag.String("rebuild-bucket", "", "GCS bucket for rebuild artifacts")
+	metadataBucket        = flag.String("metadata-bucket", "", "GCS bucket for rebuild artifacts")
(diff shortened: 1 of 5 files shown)


pkg/rebuild/schema/schema.go
```diff
@@ -229,6 +229,7 @@ type SmoketestAttempt struct {
 	Success           bool    `firestore:"success,omitempty"`
 	Message           string  `firestore:"message,omitempty"`
 	Strategy          string  `firestore:"strategy,omitempty"`
+	Dockerfile        string  `firestore:"dockerfile,omitempty"`
```
msuozzo, line 232: wanna pull this out into a separate PR?

msuozzo (approved):

pkg/rebuild/schema/schema.go
wbxyz: Moved to https://github.com/wbxyz/oss-rebuild/tree/dockerfile-firestore

PR #59, opened 2024-07-16
Add git tag heuristic test.

msuozzo:


pkg/rebuild/rebuild/git_test.go
```diff
@@ -0,0 +1,125 @@
...
+	must(repo.CreateTag(tag, commit.Hash, nil))
+}
+
+func createAnnotatedTag(repo *git.Repository, tag, targetCommit string) {
+	commit := must(repo.CommitObject(plumbing.NewHash(targetCommit)))
+	must(repo.CreateTag(tag, commit.Hash, &git.CreateTagOptions{
+		Message: "Test annotated tag",
+		Tagger:  &object.Signature{Name: "Test Author", Email: "test@example.com"},
+	}))
+}
```
wbxyz, line 125: nit: Define before use?

pkg/rebuild/rebuild/git_test.go
```diff
@@ -0,0 +1,125 @@
...
+	c2 := createCommit(repo, "commit2")
+	createLightweightTag(repo, "v1.0.0", c1)
+	createLightweightTag(repo, "v1.1.0", c2)
+	createAnnotatedTag(repo, "v1.0.0-alpha", c1)
+
+	tests := []struct {
+		pkg     string
+		version string
+		want    string
+		wantErr bool
```
wbxyz, line 59: Is there a test case where error is expected? AFAICT they're all `false`

pkg/rebuild/rebuild/git_test.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+func TestFindTagMatch(t *testing.T) {
+	repo := must(git.Init(memory.NewStorage(), memfs.New()))
+
+	c1 := createCommit(repo, "commit1")
+	c2 := createCommit(repo, "commit2")
+	createLightweightTag(repo, "v1.0.0", c1)
+	createLightweightTag(repo, "v1.1.0", c2)
+	createAnnotatedTag(repo, "v1.0.0-alpha", c1)
```
wbxyz, line 53: Does this need to be on the same c1 for the tests? It would be helpful if this was a separate c3 to make it clear when TagMatch is using the annotated tag vs the lightweight tag.

pkg/rebuild/rebuild/git.go
```diff
@@ -92,27 +91,19 @@ func FindTagMatch(pkg, version string, repo *git.Repository) (commit string, err
...
 	}
-	ti, err := repo.TagObjects()
-	if err != nil {
-		return
-	}
-	err = ti.ForEach(func(t *object.Tag) error {
-		tags = append(tags, t.Name)
-		return nil
-	})
-	return
```
wbxyz, line 115: Was the previous version just including all tags twice?

pkg/rebuild/rebuild/git_test.go
msuozzo: eh for the test methods, i think stuffing them at the end is fine

pkg/rebuild/rebuild/git_test.go
msuozzo: it doesn't need to be. was just briefer. changed

pkg/rebuild/rebuild/git.go
msuozzo: it looks like it, yes. it wouldn't have resulted in any changes in inference but it may have increased runtime by a tiny amount.

pkg/rebuild/rebuild/git_test.go
msuozzo: eh i would like to cover the error cases but they turned out to be hard to create. i'll leave a TODO

pkg/rebuild/rebuild/git_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+func TestFindTagMatch(t *testing.T) {
+	repo := must(git.Init(memory.NewStorage(), memfs.New()))
+
+	c1 := createCommit(repo, "commit1")
+	c2 := createCommit(repo, "commit2")
+	c3 := createCommit(repo, "commit3")
+	createLightweightTag(repo, "v1.0.0", c1)
+	createLightweightTag(repo, "v1.1.0", c2)
+	createAnnotatedTag(repo, "v1.0.0-alpha", c3)
```
wbxyz, line 54: Would it be worth adding a test case where FindTagMatch finds an AnnotatedTag instead of just LightweightTags?  Because it doesn't return c3 in any of the cases, it's not clear from the test cases whether FindTagMatch actually reads Annotated Tag types.

wbxyz (approved):

PR #61, opened 2024-07-23
Add exhaustive indexscan matching strategy.

msuozzo:


internal/bitmap/bitmap.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	for i := range b.bits {
+		b.bits[i] |= other.bits[i]
+	}
+}
+
+// And performs a bitwise AND operation with another bitmap, modifying this bitmap.
+func (b *Bitmap) And(other *Bitmap) {
+	if b.size != other.size {
+		return
```
wbxyz, line 84: Should these operations return an error when the sizes don't match?

Silently not performing the operation seems like it could be really challenging to debug.

tools/indexscan/main.go
```diff
@@ -251,6 +252,96 @@ func (s CommitsNearPublishStrategy) Search(ctx context.Context, r *git.Repositor
...
+	// Find subset of file hashes that are present in the repo.
+	files := make(map[plumbing.Hash]int)
+	var i int
+	for _, h := range hashes {
+		_, err = r.BlobObject(plumbing.NewHash(h))
+		if err == plumbing.ErrObjectNotFound {
+			// Skip files not present in repo.
+			err = nil
+		} else if err != nil {
+			return
```
wbxyz, line 270: panic?

tools/indexscan/main.go
```diff
@@ -251,6 +252,96 @@ func (s CommitsNearPublishStrategy) Search(ctx context.Context, r *git.Repositor
...
+	// Recursively build set of matching files for each tree.
+	matches := bitmap.NewBatch(len(files), len(trees))
+	processed := bitmap.New(len(trees))
+	ti, _ := r.TreeObjects()
+	ti.ForEach(func(t *object.Tree) error {
+		processTree(t, files, trees, matches, processed)
+		return nil
+	})
+	// Search through all commits for the one whose tree has the most matches.
+	ci, _ := r.CommitObjects()
```
wbxyz, line 304: check err?

tools/indexscan/main.go
```diff
@@ -251,6 +252,96 @@ func (s CommitsNearPublishStrategy) Search(ctx context.Context, r *git.Repositor
...
+		case filemode.Dir:
+			if !processed.Get(trees[t.Hash]) {
+				st, _ := t.Tree(e.Name)
+				processTree(st, files, trees, matches, processed)
+			}
+			matches[trees[t.Hash]].Or(&matches[trees[e.Hash]])
+		case filemode.Submodule, filemode.Symlink:
+			continue
+		default:
+			matches[trees[t.Hash]].Set(files[e.Hash])
```
wbxyz, line 338: I think I must be missing something, why doesn't files[e.Hash] sometimes cause a key error? IIUC `files` only contains the files originally passed into Search as `hashes`, but e.Hash could be any hash of any file in the entire git history (presumably not in `hashes`).

tools/indexscan/main.go
```diff
@@ -251,6 +252,96 @@ func (s CommitsNearPublishStrategy) Search(ctx context.Context, r *git.Repositor
...
+
+// processTree recursively determines the presence of a set of files in the given git Tree and records them.
+func processTree(t *object.Tree, files, trees map[plumbing.Hash]int, matches []bitmap.Bitmap, processed *bitmap.Bitmap) {
+	if processed.Get(trees[t.Hash]) {
+		return
+	}
+	for _, e := range t.Entries {
+		switch e.Mode {
+		case filemode.Dir:
+			if !processed.Get(trees[t.Hash]) {
```
wbxyz, line 330: Isn't this conditional always true? There's a guard clause at the start of processTree that IIUC is making sure this is true.

Should this be `e.Hash` instead of `t.Hash`? Or maybe just delete and let the recursive call into `processTree` do that check?

msuozzo: eh i'd prefer to avoid the call to Tree.Tree() if possible so I'll just fix the bug.

tools/indexscan/main.go
msuozzo: there's no lookup error it just maps to the default `int` value (i.e. 0) if it's not found: [link](https://go.dev/blog/maps#:~:text=If%20the%20requested%20key%20doesn%E2%80%99t%20exist%2C%20we%20get%20the%20value%20type%E2%80%99s%20zero%20value.%20In%20this%20case%20the%20value%20type%20is%20int%2C%20so%20the%20zero%20value%20is%200%3A)

tools/indexscan/main.go
msuozzo: functionally, this really shouldn't happen if we've gotten this far (this is an iterator over local values) but I'll add handling

tools/indexscan/main.go
msuozzo: no this will just return with err set

tools/indexscan/main.go
```diff
@@ -251,6 +252,103 @@ func (s CommitsNearPublishStrategy) Search(ctx context.Context, r *git.Repositor
...
+	}
+	// Iterate over repo trees to create an ordering.
+	trees := make(map[plumbing.Hash]int)
+	i = 0
+	for ti, _ := r.TreeObjects(); ; {
+		t, err := ti.Next()
+		if err == io.EOF {
+			break
+		} else if err != nil {
+			panic(err)
```
wbxyz, line 284: If the above loop just returns err instead of panic, should this loop do the same?

tools/indexscan/main.go
wbxyz: Does that mean any tree with files that don't exist in `hashes` is marked as matching `hashes[0]`?

tools/indexscan/main.go
msuozzo: I think the way it's assigned (`t, err := ti.Next()`) means the function's err value won't be updated and the name will be shadowed instead🤷

wbxyz: Ah, gotcha.

tools/indexscan/main.go
wbxyz: Discussed offline, the previous behavior was ~always~ counting a match for hashes[0], and when trying to find a "best match" that's not a very obvious bug. The latest version checks for the hash in `files` before counting it as a match.

wbxyz (approved):

PR #63, opened 2024-07-23
Refactor GCB client to parameterize wait.

msuozzo:


internal/gcb/gcb.go
```diff
@@ -29,22 +29,43 @@ import (
...
+type clientImpl struct {
+	service      *cloudbuild.Service
+	pollInterval time.Duration
+}
+
+// NewClient creates a new Client with the given options.
+func NewClient(s *cloudbuild.Service) Client {
+	return &clientImpl{
+		service:      s,
+		pollInterval: 10 * time.Second,
```
wbxyz, line 45: Want to add a comment here about why we chose 10 seconds? I remember GCB was rate limiting us at one point but I don't remember exactly how we came to 10 seconds.

wbxyz (approved):

internal/gcb/gcb.go
msuozzo: i think i did the math based on GCB's default quota but I'd like to apply for a quota bump so I won't be too specific in the comment.

PR #67, opened 2024-07-23
Add test for primary rebuild endpoint.

msuozzo:


internal/api/apiservice/rebuild_test.go
```diff
@@ -0,0 +1,275 @@
...
+					if diff := cmp.Diff(expected, actual); diff != "" {
+						t.Errorf("URL mismatch: diff\n%v", diff)
+					}
+				},
+			}
+			d.Signer, _ = dsse.NewEnvelopeSigner(&FakeSigner{})
+			fs := memfs.New()
+			afs, _ := fs.Chroot("attestations")
+			d.AttestationStore = rebuild.NewFilesystemAssetStore(afs)
+			mfs, _ := fs.Chroot("metadata")
```
wbxyz, line 157: must?

wbxyz (approved):
Very cool that we can do a test like this!

internal/api/apiservice/rebuild_test.go
msuozzo: done

PR #72, opened 2024-07-24
Implement initial version of network proxy.

msuozzo:


wbxyz (approved):
After a quick glance  and discussion offline LGTM. We're submitting as-is to match the experimental version developed in a different repo. We'll do a followup review next week to see if there's anything we want to change.

msuozzo:
Let's follow up in a bit to work through localizing this code better. Still a lot of work to be done in logging and error-handling. Filed #73 to track

PR #74, opened 2024-07-26
Firestore should return error gracefully.

wbxyz:
This caused debugging to be really challenging, TUI would simply crash and no useful error would be logged because output was still being redirected to the TUI component.

I think it's reasonable for the TUI to keep running even if a firestore query fails.

I also added a success log message because otherwise empty queries are indistinguishable from hung queries.

### DIFF
--- a/tools/ctl/firestore/firestore.go
+++ b/tools/ctl/firestore/firestore.go
@@ -277,8 +277,10 @@ func (f *Client) FetchRebuilds(ctx context.Context, req *FetchRebuildRequest) (r
 		rebuilds[r.ID()] = r
 	}
 	if err := <-cerr; err != nil {
-		log.Fatal("query error", err.Error())
+		err = errors.Wrap(err, "query error")
+		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


tools/ctl/firestore/firestore.go
```diff
@@ -277,8 +277,10 @@ func (f *Client) FetchRebuilds(ctx context.Context, req *FetchRebuildRequest) (r
 		rebuilds[r.ID()] = r
 	}
 	if err := <-cerr; err != nil {
-		log.Fatal("query error", err.Error())
+		err = errors.Wrap(err, "query error")
+		return nil, err
 	}
+	log.Printf("Fetched %d rebuilds", len(rebuilds))
```
msuozzo, line 283: this log line might go better in the caller function. not necessary but i'd generally lean towards moving log messages as far up as possible.

msuozzo (approved):

tools/ctl/firestore/firestore.go
wbxyz: Moved.

PR #75, opened 2024-07-26
Use the CreateRun schema types.

wbxyz:
Without this the TUI was failing to actually run-bench due to an argument skew.

### DIFF
--- a/tools/ctl/ctl.go
+++ b/tools/ctl/ctl.go
@@ -407,10 +407,10 @@ var runBenchmark = &cobra.Command{
 		if mode != firestore.SmoketestMode && mode != firestore.AttestMode {
 			log.Fatalf("Unknown mode: %s. Expected one of 'smoketest' or 'attest'", string(mode))
 		}
-		if *api_uri == "" {
+		if *apiUri == "" {
 			log.Fatal("API endpoint not provided")
 		}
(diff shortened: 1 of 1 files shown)


tools/ctl/ctl.go
```diff
@@ -450,29 +450,22 @@ var runBenchmark = &cobra.Command{
...
-				"hash": []string{hex.EncodeToString(set.Hash(sha256.New()))},
-				"type": []string{string(mode)},
-			}
-			u.RawQuery = values.Encode()
-			req, err := http.NewRequestWithContext(ctx, http.MethodPost, u.String(), nil)
-			if err != nil {
-				log.Fatal(errors.Wrap(err, "creating API version request"))
-			}
-			resp, err := client.Do(req)
+			resp, err := client.Do(makeHTTPRequest(ctx, apiURL.JoinPath("runs"), &schema.CreateRunRequest{
```
msuozzo, line 453: Can you use a Stub here? I think that would get rid of the serialization boilerplate below as it could project all errors into the err return value on which you could unconditionally fail.

msuozzo (approved):

tools/ctl/ctl.go
wbxyz: Done.

PR #79, opened 2024-08-08
Json decode before encode in the CLI.

wbxyz:
When fetching the build or steps from our attestations the CLI was converting them back to base64 instead of adding intendation. I think this is because we were asking the json encoder to Encode() a bytes string that was already json, so it encoded those bytes into base64 and added quotes outside it.

### DIFF
--- a/cmd/oss-rebuild/main.go
+++ b/cmd/oss-rebuild/main.go
@@ -131,7 +131,7 @@ func (b *Bundle) Byproduct(name string) ([]byte, error) {
 }
 
 func writeIndentedJson(out io.Writer, b []byte) error {
-	var decoded interface{}
+	var decoded any
 	if err := json.NewDecoder(bytes.NewBuffer(b)).Decode(&decoded); err != nil {
 		return errors.Wrap(err, "decoding json")
 	}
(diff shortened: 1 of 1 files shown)


cmd/oss-rebuild/main.go
```diff
@@ -210,19 +223,16 @@ The ecosystem is one of npm, pypi, or cratesio. For npm the artifact is the <pac
 			if err != nil {
 				log.Fatal(errors.Wrap(err, "getting build.json"))
 			}
-			e := json.NewEncoder(cmd.OutOrStdout())
-			e.SetIndent("", "  ")
-			if err := e.Encode(build); err != nil {
+			cmd.OutOrStdout().Write(build)
```
msuozzo, line 226: remove

cmd/oss-rebuild/main.go
```diff
@@ -130,6 +130,19 @@ func (b *Bundle) Byproduct(name string) ([]byte, error) {
 	return nil, fmt.Errorf("byproduct named %s not found", name)
 }
 
+func writeIndentedJson(out io.Writer, b []byte) error {
+	var decoded interface{}
```
msuozzo, line 134: any

msuozzo (approved):
would be nice to have a test (we are using cmd.OutOrStdout to enable that, after all) but feel free to punt

wbxyz:
I can add some tests in a followup PR, submitting for now to un-break CLI.

PR #81, opened 2024-08-16
Add support for network coercion in docker proxy

msuozzo:


internal/proxy/docker/docker_test.go
```diff
@@ -687,6 +687,88 @@ func TestAddBinding(t *testing.T) {
...
+func TestSetNetwork(t *testing.T) {
+	testCases := []struct {
+		name       string
+		imageSpec  []byte
+		newNetwork string
+		wantSpec   []byte
+		wantErr    bool
+	}{
+		{
+			name:       "Set new network",
```
wbxyz, line 740: It also might be worth testing when a network wasn't provided in the input (like getNetwork, but I think this time it's not an error)?

internal/proxy/docker/docker.go
```diff
@@ -384,29 +419,46 @@ type patchSet struct {
...
-	EnvVars     []string
-	JavaEnvVar  bool
-	proxySocket string
-	patchMap    map[string]*patchSet
-	m           sync.Mutex
-	created     atomic.Uint32
+	cert            x509.Certificate
+	envVars         []string
+	javaEnvVar      bool
+	requiredNetwork string // TODO: Not a good fit for this abstraction
```
wbxyz, line 425: nit: is requiredNetwork the right name for this? Required makes it sound like we're checking if it's already present, wdyt of patchedNetwork, or networkOverride, or maybe just ContainerTruststorePatcher.network?

wbxyz (approved):

internal/proxy/docker/docker.go
msuozzo: i didn't want patched because it's not unpatched. "network" is tricky because it's not clear from the name of the abstraction how a "network" would relate to it. Override is closer to what I was aiming for and i do think it's a better option than required. Done.

PR #82, opened 2024-08-16
Collapse docker build and run into a single GCB step

msuozzo:


pkg/rebuild/rebuild/rebuildremote.go
```diff
@@ -97,11 +97,7 @@ func makeBuild(t Target, dockerfile, imageUploadPath, rebuildUploadPath string,
 		Steps: []*cloudbuild.BuildStep{
 			{
 				Name:   "gcr.io/cloud-builders/docker",
-				Script: "cat <<'EOS' | docker buildx build --tag=img -\n" + dockerfile + "\nEOS",
-			},
-			{
-				Name: "gcr.io/cloud-builders/docker",
-				Args: []string{"run", "--name=container", "img"},
+				Script: "cat <<'EOS' | docker buildx build --tag=img -\n" + dockerfile + "\nEOS\ndocker run --name=container img",
```
loosebazooka, line 100: just curious how failures happen here. Do you need to set the failure mode in bash?

msuozzo: https://cloud.google.com/build/docs/build-config-file-schema#allowfailure

I believe it takes the exit code of the last command. So in this case, if docker build fails, docker run **theoretically** won't find `img`. You bring up a good point, though, that img still would reach out to docker hub, i believe. So we could either run `--pull=never` or make the shell fail if docker build fails (or both).

msuozzo: Ended up just setting the script to terminate after a failed build.

PR #85, opened 2024-08-21
Use npx -c to ensure compatibility before npm@8.0.0

msuozzo:
Comparing the two npm versions, `--` wasn't present in v7 while `-c` remains supported:
- v8: https://docs.npmjs.com/cli/v8/commands/npx
- v7: https://docs.npmjs.com/cli/v7/commands/npx

This manifested as failures during dependency install:

```
Error: Invalid tag name "cd . && npm install --force": Tags may not have any characters that encodeURIComponent encodes.
```


loosebazooka (approved):
seems fine, if we don't intend on any sort of shell expansion would it make sense to just use single quotes instead of double after the `-c`? <-- I don't care either way, your call.

msuozzo:
> seems fine, if we don't intend on any sort of shell expansion would it make sense to just use single quotes instead of double after the `-c`? <-- I don't care either way, your call.

ugh yeah i actually had that in this PR at first and took it out to reduce the sprawl of the diff. i'll send something separate to address this.

PR #90, opened 2024-08-28
Separate out AssetStore path API

msuozzo:
Previously, The reader and writer apis had an ad hoc additional return value which was the ill-defined "resource path" for that asset. This change breaks that out into a more formal API and uses URLs as the medium of exchange.

As for the design choice to split out the URL() method into a separate interface: I found it useful to have users of this function declare in their signatures exactly where they want to "break out" of accessing assets exclusively through the AssetStore.

### DIFF
--- a/pkg/builddef/set.go
+++ b/pkg/builddef/set.go
@@ -43,7 +43,7 @@ func NewFilesystemBuildDefinitionSet(fs billy.Filesystem) *FilesystemBuildDefini
 
 func (s *FilesystemBuildDefinitionSet) Get(ctx context.Context, t rebuild.Target) (rebuild.Strategy, error) {
 	definitions := rebuild.NewFilesystemAssetStore(s.fs)
-	r, _, err := definitions.Reader(ctx, rebuild.Asset{Type: rebuild.BuildDef, Target: t})
+	r, err := definitions.Reader(ctx, rebuild.Asset{Type: rebuild.BuildDef, Target: t})
 	if err != nil {
 		if errors.Is(err, rebuild.ErrAssetNotFound) {
 			return nil, nil // Return nil strategy if definition is not found
@@ -59,9 +59,9 @@ func (s *FilesystemBuildDefinitionSet) Get(ctx context.Context, t rebuild.Target
 }
 
 func (s *FilesystemBuildDefinitionSet) Path(ctx context.Context, t rebuild.Target) (string, error) {
-	definitions := rebuild.NewFilesystemAssetStore(s.fs)
-	_, pth, err := definitions.Reader(ctx, rebuild.Asset{Type: rebuild.BuildDef, Target: t})
-	return pth, err
+	defs := rebuild.NewFilesystemAssetStore(s.fs)
+	url := defs.URL(rebuild.Asset{Type: rebuild.BuildDef, Target: t})
+	return url.Path, nil
 }
 
 type GitBuildDefinitionSet struct {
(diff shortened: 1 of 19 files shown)


loosebazooka:
any chance you can add a short summary to the PR description?

msuozzo:
> any chance you can add a short summary to the PR description?

Yeah good call. Done.

msuozzo:
> any chance you can add a short summary to the PR description?

Yeah good call. Done.

loosebazooka:
I'm little hesitant to review this, wondering if you're okay waiting for William to get back on this one?

msuozzo:
> I'm little hesitant to review this, wondering if you're okay waiting for William to get back on this one?

Happy to reassign. It touches a bunch of files but really only meaningfully in ~3 spots (`remoterebuild.go`, `ui.go`, and the no-op interface change in `internal/api/apiservice/rebuild.go`).

pkg/rebuild/rebuild/rebuildremote.go
```diff
@@ -185,22 +185,14 @@ func RebuildRemote(ctx context.Context, input Input, id string, opts RemoteOptio
...
-	if err != nil {
-		return errors.Wrap(err, "creating dummy writer for container image")
-	}
-	_, rebuildUploadPath, err := opts.RemoteMetadataStore.Writer(ctx, Asset{Target: t, Type: RebuildAsset})
-	if err != nil {
-		return errors.Wrap(err, "creating dummy writer for rebuild")
-	}
+	imageUploadPath := opts.RemoteMetadataStore.URL(Asset{Target: t, Type: ContainerImageAsset}).String()
+	rebuildUploadPath := opts.RemoteMetadataStore.URL(Asset{Target: t, Type: RebuildAsset}).String()
 	build := makeBuild(t, dockerfile, imageUploadPath, rebuildUploadPath, opts)
```
wbxyz, line 190: The LocatableAssetStore is, by design, implementation agnostic. However I think makeBuild is implicitly expecting these paths to be GCS paths.

Should we add a check here that the URL schema is GCS? I suspect that might break some tests that are using a filesystem assetstore for convenience, but it feels wrong to use the generic LocatableAssetStore in places where we actually want the specific GCSAssetStore

pkg/builddef/set.go
```diff
@@ -59,9 +59,9 @@ func (s *FilesystemBuildDefinitionSet) Get(ctx context.Context, t rebuild.Target
 }
 
 func (s *FilesystemBuildDefinitionSet) Path(ctx context.Context, t rebuild.Target) (string, error) {
-	definitions := rebuild.NewFilesystemAssetStore(s.fs)
-	_, pth, err := definitions.Reader(ctx, rebuild.Asset{Type: rebuild.BuildDef, Target: t})
-	return pth, err
+	defs := rebuild.NewFilesystemAssetStore(s.fs)
+	url := defs.URL(rebuild.Asset{Type: rebuild.BuildDef, Target: t})
+	return url.Path, nil
```
wbxyz, line 64: Why not return in one line?

pkg/rebuild/rebuild/rebuildremote.go
msuozzo: So gsutil does, in fact, work with `file://` URLs but it's true that the interface is not necessarily enforcing the exact behavior we believe will make the system work correctly. Directionally, though, I think it does provide a clearer indication of where we're breaking out of the normal AssetStore usage pattern and allows us to think about encoding more in the type system in the future.

wbxyz: Yeah this change is definitely an improvement, adding structure to the URLs allows the caller to at least inspect what kind of location is being returned.

The previous implementation was only providing an unstructured string which made it difficult for the caller to understand what they can do with the resource.

wbxyz (approved):

pkg/builddef/set.go
msuozzo: figured it was marginally easier to read this way. that said, i think it might be a good idea to revisit this API after this change.

PR #91, opened 2024-08-28
Add dedent utility

msuozzo:
this mirrors python's textwrap.dedent to make string literals a bit easier to read in our go source files.

### DIFF
--- a/internal/textwrap/textwrap.go
+++ b/internal/textwrap/textwrap.go
@@ -1,0 +1,57 @@
+package textwrap
+
+import (
+	"strings"
+)
+
+// Dedent removes common leading whitespace from each line.
+//
+// Adapted from Python's textwrap.dedent.
+//
+// Note: Only tabs and spaces are considered whitespace.
+// Note: Blank lines are normalized to a newline character.
+func Dedent(text string) string {
+	isSpaceOrTab := func(r rune) bool { return r == ' ' || r == '\t' }
+	lines := strings.Split(text, "\n")
+
+	// Find the text's common indent.
+	var commonIndent string
+	var foundIndent bool
+	for _, line := range lines {
+		if len(line) == 0 {
+			continue // Skip blank lines
+		}
+		content := strings.TrimLeftFunc(line, isSpaceOrTab)
+		indent := line[:len(line)-len(content)]
+		if !foundIndent {
+			commonIndent = indent
+		} else if strings.HasPrefix(indent, commonIndent) {
+			// More indented -> no change
+			continue
+		} else if strings.HasPrefix(commonIndent, indent) {
+			// Less indented -> update to this indent
+			commonIndent = indent
+		} else {
+			// Mismatched indent -> update to largest common prefix
+			for i := 0; i < len(commonIndent) && i < len(indent); i++ {
+				if commonIndent[i] != indent[i] {
+					commonIndent = commonIndent[:i]
+					break
+				}
+			}
+		}
+		foundIndent = true
+	}
+
+	// Remove the common indent from each line.
+	var result []string
+	for _, line := range lines {
+		if strings.TrimSpace(line) == "" {
+			result = append(result, strings.TrimLeftFunc(line, isSpaceOrTab))
+		} else {
+			result = append(result, strings.TrimPrefix(line, commonIndent))
+		}
+	}
+
+	return strings.Join(result, "\n")
+}
--- a/internal/textwrap/textwrap_test.go
+++ b/internal/textwrap/textwrap_test.go
@@ -1,0 +1,228 @@
+package textwrap
+
+import (
+	"strings"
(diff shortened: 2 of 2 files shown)


internal/textwrap/textwrap_test.go
```diff
@@ -0,0 +1,213 @@
...
+		{
+			name: "Very long lines",
+			input: `
+\t\t\t\tThis is a very long line that goes on and on and on and on and on and on and on and on and on and on and on and on and on and on and on
+\t\t\t\t\tThis is another very long line that goes on and on and on and on and on and on and on and on and on and on and on and on and on and on
+\t\t\t\tThis is a third very long line that goes on and on and on and on and on and on and on and on and on and on and on and on and on and on and on
+\t\t\t\t`,
+			expected: `
+This is a very long line that goes on and on and on and on and on and on and on and on and on and on and on and on and on and on and on
+\tThis is another very long line that goes on and on and on and on and on and on and on and on and on and on and on and on and on and on
```
loosebazooka, line 165: what's up with this `\t`?

msuozzo: Was a pain in the ass to figure out which were tabs and spaces in my editor but it's pretty important this test include combinations of. So I added an escape expander to the test harness.

loosebazooka: is there another way that can be explained to me :confused: ? just looking at this test, I wouldn't expect

`\t\t\t\t\tThis is another...` to become `\tThis is another`

msuozzo: Ah it's a little hard to see with wrapped lines but this line has 5 tabs while the least common indent is 4 tabs.

loosebazooka: oooooooooooh. Gotcha. dedent, not ltrim or whatever.

msuozzo: righhht yeah it's aimed at making in-code string literals line up better with the indentation of the surrounding code

internal/textwrap/textwrap_test.go
```diff
@@ -0,0 +1,213 @@
...
+\t\t\t`,
+			expected: `
+Hello
+World
+Foo
+Bar
+`,
+		},
+		{
+			name: "Mixed tabs and spaces",
```
loosebazooka, line 60: I think I might like to see a test like (and maybe I missed this)
```
\tHello
 World
 ```
 
 to define if tabs and spaces are treated equivalently.

Comment (msuozzo) [reply internal/textwrap/textwrap_test.go:60]: Yeah I think this case should have a bare 3 tab and 3 tab+1 space lines for completeness.

Comment (msuozzo) [reply internal/textwrap/textwrap_test.go:60]: Done

Comment (loosebazooka) [reply internal/textwrap/textwrap_test.go:60]: I'm not sure this covers what I had in mind. This still just strips 3x \t. I was curious if \t and ' ' are treated equivalent and this, from what I can see, may not answer that question?

like does 
```
\tHello
 World
 ```
 dedent to
 ```
 Hello
 World
 ```
 or remain
 ```
 \tHello
  World
  ```

msuozzo: They're not equivalent so they'd remain not indented. I'll add a test for this.

msuozzo: Done

msuozzo:
Okay I've reduced the number of indents at the start of most test cases from 3 to 1 so the changes after the initial indents are clearer.

internal/textwrap/textwrap.go
```diff
@@ -0,0 +1,57 @@
...
+//
+// Note: Only tabs and spaces are considered whitespace.
+// Note: Blank lines are normalized to a newline character.
+func Dedent(text string) string {
+	isSpaceOrTab := func(r rune) bool { return r == ' ' || r == '\t' }
+	lines := strings.Split(text, "\n")
+
+	// Find the text's common indent.
+	var commonIndent string
+	var foundIndent bool
```
loosebazooka, line 19: just curious if we can get rid of this flag and just work off of `commonIdent != ""`

loosebazooka (approved):

loosebazooka:
There might actually be a missing test case for
```
\t \tHello
\t World
```
where commonIndent is a mixed tab space situation `\t `?

internal/textwrap/textwrap.go
msuozzo: Nah because the commonIndent can be empty e.g. `"A\n B\n C"`. We don't want `" "` to set `commonIndent` when we hit `B` because the true common indent is, indeed, empty.

loosebazooka: aaaaaah. right.

loosebazooka (approved):

PR #95, opened 2024-09-04
Use template to construct asset upload script

msuozzo:


pkg/rebuild/rebuild/rebuildremote.go
```diff
@@ -60,36 +60,67 @@ var rebuildContainerTpl = template.Must(
...
+				 {{.Instructions.Deps | indent}}
+				EOF
+				RUN cat <<'EOF' >build
+				 set -eux
+				 {{.Instructions.Build | indent}}
+				 mkdir /out && cp /src/{{.Instructions.OutputPath}} /out/
+				EOF
+				WORKDIR "/src"
+				ENTRYPOINT ["/bin/sh","/build"]
+				`)[1:], // remove leading newline
```
wbxyz, line 91: Should textwrap.Dedent auto-handle the leading newline, like in python?

wbxyz (approved):

pkg/rebuild/rebuild/rebuildremote.go
msuozzo: so python doesn't but the idiom is to include a backslash just after the open quotes to remove it. since there's no equivalent in golang string literals, we're a bit stuck. i'd like to obviate this but i'm also not crazy about diverging from the python API. i don't care very much either way, though

wbxyz: Ah right, I forgot about the backslash idiom.

Either way seems good to me.

PR #97, opened 2024-09-05
Add TransparentProxyService constructor

### LINKED ISSUE #89: Initialize TransparentProxyService.NetworkLog within a constructor
Managing the construction and synchronization of that call is now an implementation detail of the new Service abstraction. We'll need to add a constructor, move the mutex onto a private attr of the struct, and initialize the netlog to the CaptureAcivityLog value in that constructor.

andusy:
- Create constructor with mutex as a private member.
 - Initialize the network log inside the constructor.

Closes #89

### DIFF
--- a/cmd/proxy/main.go
+++ b/cmd/proxy/main.go
@@ -36,7 +36,8 @@ func main() {
 	if *verbose {
 		log.Printf("Server starting up! - configured to listen on http interface %s and https interface %s", *httpProxyAddr, *tlsProxyAddr)
 	}
-	proxyServer := proxy.NewTransparentProxyService(*verbose, ca)
+	p := proxy.NewTransparentProxyServer(*verbose)
+	proxyServer := proxy.NewTransparentProxyService(p, ca)
 	// Administrative endpoint.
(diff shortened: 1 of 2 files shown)


cmd/proxy/main.go
```diff
@@ -38,15 +36,9 @@ func main() {
 	if *verbose {
 		log.Printf("Server starting up! - configured to listen on http interface %s and https interface %s", *httpProxyAddr, *tlsProxyAddr)
 	}
-	p := proxy.NewTransparentProxyServer(*verbose)
```
msuozzo, line 41: i don't love this choice. I think I'd probably lean towards keeping the proxy _server_ construction exposed in the service constructor. mainly i'd be concerned if that constructor had to have more configuration, where would the extra params go. one answer could be a ServerOpts struct which could be added later but I'm not sure that's clearer than or preferable to exposing the server object itself.

andusy: I think that's a fair point. I left the initialization of the proxy server outside of the service.

msuozzo (approved):

PR #99, opened 2024-09-08
Implement Policy for network proxy

andusy:
- Add policy package that provides a Policy struct describing a set of rules applied to each incoming request.
 - Add flag to toggle policy modes to determine whether or not a given policy should be applied to an incoming request.

### DIFF
--- a/cmd/proxy/main.go
+++ b/cmd/proxy/main.go
@@ -26,7 +26,7 @@ var (
 	dockerJavaEnvVar  = flag.Bool("docker_java_truststore", false, "whether to patch containers with Java proxy cert truststore file and env var")
 	dockerProxySocket = flag.Bool("docker_recursive_proxy", false, "whether to patch containers with a unix domain socket which proxies docker requests from created containers")
 	// TODO: Implement flag for reading a policy file.
-	policyMode = flag.String("policy_mode", "disabled", "mode to run the proxy in. Options: monitor, enforce")
+	policyMode = flag.String("policy_mode", "disabled", "mode to run the proxy in. Options: disabled, enforce")
 )
 
 func main() {
--- a/cmd/proxy/main.go
+++ b/cmd/proxy/main.go
@@ -40,10 +40,10 @@ func main() {
 		log.Printf("Server starting up! - configured to listen on http interface %s and https interface %s", *httpProxyAddr, *tlsProxyAddr)
 	}
 	p := proxy.NewTransparentProxyServer(*verbose)
-	proxyService := proxy.NewTransparentProxyService(p, ca, *proxyMode)
+	proxyService := proxy.NewTransparentProxyService(p, ca, proxy.ProxyMode(*proxyMode))
 	proxyService.Proxy.OnRequest().DoFunc(
 		func(req *http.Request, ctx *goproxy.ProxyCtx) (*http.Request, *http.Response) {
-			return proxyService.CheckNetworkPolicy(req, ctx)
+			return proxyService.ApplyNetworkPolicy(req, ctx)
 		})
 	// Administrative endpoint.
 	go proxyService.ServeMetadata(*ctrlAddr)
--- a/cmd/proxy/main.go
+++ b/cmd/proxy/main.go
@@ -4,9 +4,11 @@ package main
 import (
 	"flag"
 	"log"
+	"net/http"
 	"strings"
 	"time"
 
+	"github.com/elazarl/goproxy"
 	"github.com/google/oss-rebuild/pkg/proxy/cert"
 	"github.com/google/oss-rebuild/pkg/proxy/docker"
 	"github.com/google/oss-rebuild/pkg/proxy/proxy"
@@ -23,6 +25,7 @@ var (
 	dockerEnvVars     = flag.String("docker_truststore_env_vars", "", "comma-separated env vars to populate with the proxy cert and patch into containers")
 	dockerJavaEnvVar  = flag.Bool("docker_java_truststore", false, "whether to patch containers with Java proxy cert truststore file and env var")
 	dockerProxySocket = flag.Bool("docker_recursive_proxy", false, "whether to patch containers with a unix domain socket which proxies docker requests from created containers")
+	proxyMode         = flag.String("proxy_mode", "monitor", "mode to run the proxy in. Options: monitor, enforce")
 )
 
 func main() {
@@ -37,12 +40,16 @@ func main() {
 		log.Printf("Server starting up! - configured to listen on http interface %s and https interface %s", *httpProxyAddr, *tlsProxyAddr)
 	}
 	p := proxy.NewTransparentProxyServer(*verbose)
-	proxyServer := proxy.NewTransparentProxyService(p, ca)
+	proxyService := proxy.NewTransparentProxyService(p, ca, *proxyMode)
+	proxyService.Proxy.OnRequest().DoFunc(
+		func(req *http.Request, ctx *goproxy.ProxyCtx) (*http.Request, *http.Response) {
+			return proxyService.CheckNetworkPolicy(req, ctx)
+		})
 	// Administrative endpoint.
-	go proxyServer.ServeMetadata(*ctrlAddr)
+	go proxyService.ServeMetadata(*ctrlAddr)
 	// Start proxy server endpoints.
-	go proxyServer.ProxyTLS(*tlsProxyAddr)
-	go proxyServer.ProxyHTTP(*httpProxyAddr)
+	go proxyService.ProxyTLS(*tlsProxyAddr)
+	go proxyService.ProxyHTTP(*httpProxyAddr)
 	if len(*dockerAddr) > 0 {
 		vars := strings.Split(*dockerEnvVars, ",")
 		ctp, err := docker.NewContainerTruststorePatcher(*ca.Leaf, docker.ContainerTruststorePatcherOpts{
--- a/cmd/proxy/main.go
+++ b/cmd/proxy/main.go
@@ -25,7 +25,8 @@ var (
 	dockerEnvVars     = flag.String("docker_truststore_env_vars", "", "comma-separated env vars to populate with the proxy cert and patch into containers")
 	dockerJavaEnvVar  = flag.Bool("docker_java_truststore", false, "whether to patch containers with Java proxy cert truststore file and env var")
 	dockerProxySocket = flag.Bool("docker_recursive_proxy", false, "whether to patch containers with a unix domain socket which proxies docker requests from created containers")
-	proxyMode         = flag.String("proxy_mode", "monitor", "mode to run the proxy in. Options: monitor, enforce")
+	// TODO: Implement flag for reading a policy file.
+	policyMode = flag.String("policy_mode", "disabled", "mode to run the proxy in. Options: monitor, enforce")
 )
 
 func main() {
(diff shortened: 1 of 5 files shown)


pkg/proxy/proxy/transparent.go
```diff
@@ -67,21 +68,48 @@ func NewTransparentProxyServer(verbose bool) *goproxy.ProxyHttpServer {
...
+	Mode       ProxyMode
+
+	mx *sync.Mutex
 }
 
 // NewTransparentProxyService creates a new TransparentProxyService.
-func NewTransparentProxyService(p *goproxy.ProxyHttpServer, ca *tls.Certificate) TransparentProxyService {
+func NewTransparentProxyService(p *goproxy.ProxyHttpServer, ca *tls.Certificate, mode string) TransparentProxyService {
 	m := new(sync.Mutex)
+	pm := StringToProxyMode(mode)
```
msuozzo, line 104: okay so I think here it could be cool to do something closer to this: https://www.davidagood.com/go-enum-pattern/

pkg/proxy/proxy/transparent.go
```diff
@@ -175,6 +203,55 @@ func (t *TransparentProxyService) ServeMetadata(addr string) {
...
+}
+
+type MatchingStrategy int
+
+const (
+	PathPrefix MatchingStrategy = iota
+	FullPath
+)
+
+type NetworkPolicy struct {
```
msuozzo, line 234: I feel like this stuff should actually go in a separate package or, at the very least, file. Either one is fine with me but something like policy.Rule or netpolicy.Rule would probably be cleaner than long names here.

pkg/proxy/proxy/transparent.go
```diff
@@ -175,6 +203,55 @@ func (t *TransparentProxyService) ServeMetadata(addr string) {
...
+				return req, nil
+			}
+		}
+	}
+	log.Printf("Request to %s blocked by network policy", url.String())
+	errorMessage := fmt.Sprintf("Access to %s is blocked by the proxy's network policy", url.String())
+	return req, goproxy.NewResponse(req, goproxy.ContentTypeText, http.StatusForbidden, errorMessage)
+}
+
+type MatchingStrategy int
```
msuozzo, line 227: nit: this is more just a local quirk of the project but we use "Strategy" as a somewhat important interface elsewhere so it might be nice to find another descriptor instead. "Type"?

pkg/proxy/proxy/transparent.go
```diff
@@ -175,6 +203,55 @@ func (t *TransparentProxyService) ServeMetadata(addr string) {
...
+const (
+	PathPrefix MatchingStrategy = iota
+	FullPath
+)
+
+type NetworkPolicy struct {
+	Rules []NetworkPolicyRule
+}
+
+type NetworkPolicyRule struct {
```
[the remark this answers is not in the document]
andusy: I think the existing goproxy methods for filtering are not quite as flexible as we may need them to be. My initial thought is that it may not necessarily make the code simpler/cleaner since we will potentially need custom handlers for things like redirection in the future anyways, but this could be something to look into.

pkg/proxy/proxy/transparent.go
```diff
@@ -175,6 +203,55 @@ func (t *TransparentProxyService) ServeMetadata(addr string) {
...
+	}
+	log.Printf("Request to %s blocked by network policy", url.String())
+	errorMessage := fmt.Sprintf("Access to %s is blocked by the proxy's network policy", url.String())
+	return req, goproxy.NewResponse(req, goproxy.ContentTypeText, http.StatusForbidden, errorMessage)
+}
+
+type MatchingStrategy int
+
+const (
+	PathPrefix MatchingStrategy = iota
```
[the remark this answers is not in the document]
msuozzo: my main concern would be that prefix is a more permissive default so the failure/misuse case is worse than with full match.

pkg/proxy/proxy/transparent.go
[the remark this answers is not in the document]
msuozzo: yeah I sort of suspected that but i figured it was worth throwing out there. thanks!

pkg/proxy/proxy/transparent_test.go
```diff
@@ -4,26 +4,28 @@ import (
...
 	proxyService := TransparentProxyService{}
 	tests := []struct {
 		name     string
 		mode     ProxyMode
-		policy   NetworkPolicy
+		policy   policy.Policy
 		url      string
 		wantResp int
 	}{
 		{
```
msuozzo, line 20: hmm i think these tests might belong in the policy package now. we can have a few tests here to make sure the check is working properly but the exhaustive rule tests belong alongside their defining code.

pkg/proxy/proxy/transparent.go
```diff
@@ -101,8 +98,8 @@ type TransparentProxyService struct {
 // NewTransparentProxyService creates a new TransparentProxyService.
 func NewTransparentProxyService(p *goproxy.ProxyHttpServer, ca *tls.Certificate, mode string) TransparentProxyService {
 	m := new(sync.Mutex)
-	pm := StringToProxyMode(mode)
-	if pm == UnkownMode {
+	pm := ProxyMode(mode)
```
msuozzo, line 101: since this is part of the public pkg interface, might it make sense to change the type of the param?

pkg/proxy/proxy/transparent.go
```diff
@@ -175,6 +200,14 @@ func (t *TransparentProxyService) ServeMetadata(addr string) {
 	}
 }
 
+// Check that the requested url is allowed by the network policy.
+func (proxy TransparentProxyService) CheckNetworkPolicy(req *http.Request, ctx *goproxy.ProxyCtx) (*http.Request, *http.Response) {
```
msuozzo, line 204: nit: since this is actually modifying the request/response, I'd probably call this "ApplyNetworkPolicy." something called "check" i expect to return a bool or an error.

pkg/proxy/policy/policy.go
```diff
@@ -0,0 +1,61 @@
...
+	PathPrefix MatchingType = iota
+	FullPath
+)
+
+type Policy struct {
+	Rules []Rule
+}
+
+func (p Policy) EnforcePolicy(req *http.Request, ctx *goproxy.ProxyCtx) (*http.Request, *http.Response) {
+	url := req.URL
```
msuozzo, line 25: nit: I'd probably inline this var

pkg/proxy/proxy/transparent_test.go
andusy: I added these tests to the policy package and also changed the tests in the proxy package to just test if the proxy mode is working properly.

pkg/proxy/proxy/transparent.go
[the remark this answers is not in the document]
andusy: Thinking about it again I think if the user gets to this point they are probably intentionally trying to enforce a policy, in which case I agree it should probably be more restrictive than not. 

I've changed this to a string enum instead (`MatchingType` in policy.go) so that the type must explicitly match one of these options. wdyt?

pkg/proxy/policy/policy.go
```diff
@@ -0,0 +1,61 @@
...
+	for _, rule := range p.Rules {
+		if rule.IsCompliant(req) {
```
msuozzo, line 19: hmm I actually think it's tricky to understand what to expect with these names. A "policy" with many "rule"s which evaluate "compliance" makes me expect that a request must pass all rules to satisfy the policy. Is there another way we could assign names here to make this less surprising?

Maybe just renaming "Policy.Rules" to "Policy.Allowlist"? Or maybe "Policy.AllowAny"? Or "Policy.AnyOf"

pkg/proxy/policy/policy.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+	}
+	log.Printf("Request to %s blocked by network policy", url.String())
+	errorMessage := fmt.Sprintf("Access to %s is blocked by the proxy's network policy", url.String())
+	return req, goproxy.NewResponse(req, goproxy.ContentTypeText, http.StatusForbidden, errorMessage)
+}
+
+// Rule interface with method to check compliance of incoming http(s) requests.
+type Rule interface {
+	IsCompliant(req *http.Request) bool
```
msuozzo, line 30: nit: maybe change this to "Allows"? or "Accepts"?

pkg/proxy/policy/policy.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+type URLMatchRule struct {
+	Host string
+	Path string
+	Type MatchingType
```
msuozzo, line 43: oh i missed this the first time but maybe change the names of MatchingType values to be "FullMatch" and "PrefixMatch" and then this struct member to be "PathMatch"

That way MatchingType is more reusable if it needs to be applied to some other part of the URL and the URLMatchRule member has a clearer affinity.

pkg/proxy/policy/policy.go
```diff
@@ -0,0 +1,60 @@
...
+	"strings"
+
+	"github.com/elazarl/goproxy"
+)
+
+type Policy struct {
+	Rules []Rule
+}
+
+func (p Policy) EnforcePolicy(req *http.Request, ctx *goproxy.ProxyCtx) (*http.Request, *http.Response) {
```
msuozzo, line 17: nit: Since it's a method  on the Policy, maybe just "Enforce" or, better yet, "Apply"

pkg/proxy/proxy/transparent.go
```diff
@@ -175,6 +199,14 @@ func (t *TransparentProxyService) ServeMetadata(addr string) {
 	}
 }
 
+// Check that the requested url is allowed by the network policy.
+func (proxy TransparentProxyService) ApplyNetworkPolicy(req *http.Request, ctx *goproxy.ProxyCtx) (*http.Request, *http.Response) {
+	if proxy.Mode != EnforcementMode {
```
msuozzo, line 204: I'm now realizing that "monitor"ing a policy as implemented is actually ignoring that policy. do you think it's worth having a dry-run / audit of the policy? If not, maybe  just tweak the wording to be something like "disabled" (and we could add in a "monitor" at a later point, if necessary).

cmd/proxy/main.go
```diff
@@ -23,6 +25,7 @@ var (
 	dockerEnvVars     = flag.String("docker_truststore_env_vars", "", "comma-separated env vars to populate with the proxy cert and patch into containers")
 	dockerJavaEnvVar  = flag.Bool("docker_java_truststore", false, "whether to patch containers with Java proxy cert truststore file and env var")
 	dockerProxySocket = flag.Bool("docker_recursive_proxy", false, "whether to patch containers with a unix domain socket which proxies docker requests from created containers")
+	proxyMode         = flag.String("proxy_mode", "monitor", "mode to run the proxy in. Options: monitor, enforce")
```
msuozzo, line 28: What's the rough plan for loading the policy at the CLI? do you want to provide a quick-and-dirty json- or file-based flag to load the policy, if provided?

msuozzo (commented):
Alright very nice. I think we're basically there.

pkg/proxy/policy/policy.go
```diff
@@ -0,0 +1,63 @@
...
+
+// Rule interface with method to check compliance of incoming http(s) requests.
+type Rule interface {
+	Allows(req *http.Request) bool
+}
+
+type MatchingType string
+
+const (
+	FullMatch   MatchingType = "full_path"
```
msuozzo, line 39: nit: full_path -> full

cmd/proxy/main.go
msuozzo: yeah that sounds good. want to add a TODO here for the flag?

pkg/proxy/proxy/transparent.go
msuozzo: hmm okay so in that case another option could be to rename this to PolicyMode and have "disabled", "monitor/dry-run/audit", "enforce". If we want to add that middle option, i think prefer that but if we don't, i think i could work with the current naming scheme

msuozzo (approved):
nice!

pkg/proxy/proxy/transparent.go
andusy: I've renamed it to PolicyMode and only added disabled/enforce for now and I'll create a new bug to add a dry-run mode.

PR #100, opened 2024-09-09
Add network proxy support to remote rebuilds

msuozzo:


pkg/rebuild/rebuild/rebuildremote.go
```diff
@@ -110,13 +176,56 @@ var assetUploadTpl = template.Must(
...
 func makeBuild(t Target, dockerfile string, opts RemoteOptions) (*cloudbuild.Build, error) {
+	var buildScript bytes.Buffer
+	uploads := []upload{
+		{From: "/workspace/image.tgz", To: opts.RemoteMetadataStore.URL(Asset{Target: t, Type: ContainerImageAsset}).String()},
+		{From: path.Join("/workspace", t.Artifact), To: opts.RemoteMetadataStore.URL(Asset{Target: t, Type: RebuildAsset}).String()},
+	}
+	if opts.UseNetworkProxy {
+		err := proxyBuildTpl.Execute(&buildScript, map[string]any{
+			"UtilPrebuildBucket": opts.UtilPrebuildBucket,
+			"DockerIPTemplate":   `{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}`,
```
wbxyz, line 187: I think this is maybe a bug? It results in non-evaluated template code in the final build steps.

pkg/rebuild/rebuild/rebuildremote.go
```diff
@@ -91,6 +92,71 @@ var rebuildContainerTpl = template.Must(
...
+// TODO: Support IPv6.
+var proxyBuildTpl = template.Must(
+	template.New(
+		"proxy build",
+	).Funcs(template.FuncMap{
+		"join": func(sep string, s []string) string { return strings.Join(s, sep) },
+	}).Parse(
+		textwrap.Dedent(`
+				set -eux
+				docker run --volume=/workspace:/workspace gcr.io/cloud-builders/gsutil cp -P gs://{{.UtilPrebuildBucket}}/proxy /workspace
```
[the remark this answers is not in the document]
msuozzo: Per a previous conversation, I think the plan was to keep these bootstrap binaries until we move to either in-build or release-based builds of these components. For now, I think we should just maintain this pattern.

pkg/rebuild/rebuild/rebuildremote.go
msuozzo: nah i didn't know of a better way to embed templates in templates. this works but the printf solution you suggested offline works better. thanks!

pkg/rebuild/rebuild/rebuildremote.go
```diff
@@ -91,6 +92,71 @@ var rebuildContainerTpl = template.Must(
...
+						-docker_addr=:{{.DockerPort}} \
+						-docker_socket=/var/run/docker.sock \
+						-docker_truststore_env_vars={{join "," .CertEnvVars}} \
+						-docker_network=container:build \
+						-docker_java_truststore=true"
+				'
+				proxyIP=$(docker inspect -f '{{.DockerIPTemplate}}' proxy)
+				docker network connect cloudbuild proxy
+				docker run --detach --name=build --network=proxynet --entrypoint=/bin/sh gcr.io/cloud-builders/docker -c 'sleep infinity'
+				docker exec --privileged build /bin/sh -euxc '
```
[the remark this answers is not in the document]
msuozzo: Yep exactly!

pkg/rebuild/rebuild/rebuildremote.go
```diff
@@ -91,6 +92,71 @@ var rebuildContainerTpl = template.Must(
...
+					echo 1 | sudo tee /proc/sys/net/ipv4/ip_forward > /dev/null
+					iptables -t nat -A OUTPUT -p tcp --dport {{.HTTPPort}} -j ACCEPT
+					iptables -t nat -A OUTPUT -p tcp --dport {{.TLSPort}} -j ACCEPT
+					iptables -t nat -A OUTPUT -p tcp -m owner --uid-owner '$uid' -j ACCEPT
+					iptables -t nat -A OUTPUT -p tcp --dport 80 -j DNAT --to-destination '$proxyIP':{{.HTTPPort}}
+					iptables -t nat -A OUTPUT -p tcp --dport 443 -j DNAT --to-destination '$proxyIP':{{.TLSPort}}
```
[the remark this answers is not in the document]
msuozzo: nope these need to be the http and https ports respectively. they correspond to the configurable ones we serve from the proxy which do make sense to be encoded in templates.

pkg/rebuild/rebuild/rebuildremote.go
```diff
@@ -91,6 +92,71 @@ var rebuildContainerTpl = template.Must(
...
+					chown {{.User}} /var/run/docker.sock
+					su - {{.User}} -c "/workspace/proxy \
+						-verbose=true \
+						-http_addr=:{{.HTTPPort}} \
+						-tls_addr=:{{.TLSPort}} \
+						-ctrl_addr=:{{.CtrlPort}} \
+						-docker_addr=:{{.DockerPort}} \
+						-docker_socket=/var/run/docker.sock \
+						-docker_truststore_env_vars={{join "," .CertEnvVars}} \
+						-docker_network=container:build \
```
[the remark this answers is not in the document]
msuozzo: `docker_network` is the one that the proxy coerces all incoming requests to use while `network` is just the network that this docker container is connected to (we also connect to the cloudbuild network below so the proxy interface becomes reachable from the top-level script).

pkg/rebuild/rebuild/rebuildremote.go
```diff
@@ -91,6 +92,71 @@ var rebuildContainerTpl = template.Must(
...
+						-docker_java_truststore=true"
+				'
+				proxyIP=$(docker inspect -f '{{.DockerIPTemplate}}' proxy)
+				docker network connect cloudbuild proxy
```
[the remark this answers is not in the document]
msuozzo: PTAL. Most of the weird design bits should be covered there. To answer your question about "http://proxy", docker's DNS resolver will route the container's name to the container when they're connected to the same network.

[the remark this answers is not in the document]
wbxyz: Awesome, thanks. The comment is very helpful.

wbxyz (approved):

pkg/rebuild/rebuild/rebuildremote.go
```diff
@@ -91,6 +93,110 @@ var rebuildContainerTpl = template.Must(
...
+//     network which joins both the same network AND the same network
+//     namespace. So we use the proxy to enforce that all new containers use
+//     the "container:build" network.
+//   - User setup: We use the iptables' owner module feature of redirecting
+//     traffic based on the UID of the originating process. To do so here,
+//     though, we need a user/uid that's shared across the proxy and build
+//     containers. The user namespaces are not shared between the containers so
+//     we cannot refer to the user by name from one if it's created by the
+//     other. Creating a user at the top-level, recreating it in the proxy, and
+//     using the uid in the build satisfies our constraints.
```
sethkoehler, line 136: Are you worried at all that someone may create a user inside the container that isn't caught by the iptables rules?

pkg/rebuild/rebuild/rebuildremote.go
```diff
@@ -91,6 +93,110 @@ var rebuildContainerTpl = template.Must(
...
+				'
+				proxyIP=$(docker inspect -f '{{printf "%s" "{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}"}}' proxy)
+				docker network connect cloudbuild proxy
+				docker run --detach --name=build --network=proxynet --entrypoint=/bin/sh gcr.io/cloud-builders/docker -c 'sleep infinity'
+				docker exec --privileged build /bin/sh -euxc '
+					iptables -t nat -A OUTPUT -p tcp --dport {{.HTTPPort}} -j ACCEPT
+					iptables -t nat -A OUTPUT -p tcp --dport {{.TLSPort}} -j ACCEPT
+					iptables -t nat -A OUTPUT -p tcp -m owner --uid-owner '$uid' -j ACCEPT
+					iptables -t nat -A OUTPUT -p tcp --dport 80 -j DNAT --to-destination '$proxyIP':{{.HTTPPort}}
+					iptables -t nat -A OUTPUT -p tcp --dport 443 -j DNAT --to-destination '$proxyIP':{{.TLSPort}}
```
sethkoehler, line 184: Is there any concern that folks will work around the proxy by using other ports (may not be an issue for your use case, but wanted to make sure what the threat vectors were here)?

pkg/rebuild/rebuild/rebuildremote.go
msuozzo: So this rule means all users except the proxy get redirected to the proxy so I think the only viable attack would be if the build was able to escalate privilege and execute code as the proxy user. That's definitely a concern but many other parts of this security model break down at that point.

pkg/rebuild/rebuild/rebuildremote.go
msuozzo: Absolutely. As you've noted before, I think the main challenge here will be SSH with DNS also presenting difficulty. Once those are dealt with, we plan on blocking all other egress via similar rules.

pkg/rebuild/rebuild/rebuildremote.go
sethkoehler: Ah sorry, I'd missed that, that addresses my concern!

pkg/rebuild/rebuild/rebuildremote.go
sethkoehler: Gotcha, blocking all other egress (once a few other pieces are in place) sgtm!

PR #101, opened 2024-09-09
Migrate uses of os errors to fs equivalents

msuozzo:
The os errors are intended for use in OS I/O operations, not as generic errors for fs-like abstractions. See guidance: https://pkg.go.dev/os#IsNotExist

### DIFF
--- a/pkg/proxy/docker/docker.go
+++ b/pkg/proxy/docker/docker.go
@@ -29,7 +29,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"io/fs"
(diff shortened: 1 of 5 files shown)


pkg/proxy/docker/docker.go
```diff
@@ -30,6 +30,7 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
+	iofs "io/fs"
```
wbxyz, line 33: Should the non-aliased import be removed?

wbxyz (approved):

PR #104, opened 2024-09-11
Improve docker proxy logging and error-handling

msuozzo:


pkg/proxy/docker/docker.go
```diff
@@ -617,35 +617,35 @@ func (d *ContainerTruststorePatcher) proxyRequest(clientConn, serverConn net.Con
 		certBytes := cert.ToPEM(&d.cert)
 		// NOTE: This doesn't need to be cleaned up due to the enclosing volume
 		// binding made at creation time.
-		if err := createFile(fs, id, certBytes, proxyCertPath); err != nil {
-			log.Println(err.Error())
+		if err := createFile(fs, certBytes, proxyCertPath); err != nil {
+			log.Printf("Creating proxy cert: %v", err)
```
wbxyz, line 621: Any particular reason to prefer `log.Printf` instead of `log.Println(error.Wrap(err, "bla"))`?

wbxyz (approved):

pkg/proxy/docker/docker.go
msuozzo: yeah i think they're pretty equivalent. the two reasons we might prefer this here is that Wrap grabs and adds a stacktrace to the error and the printf form allows a one character change if we want to print out the existing stack trace (`%v` -> `%+v`)

PR #109, opened 2024-09-25
Implement policy update methods

andusy:
- Expose a /policy endpoint to update the network policy.
- Add a flag to read json files that set the network policy on startup.

### DIFF
--- a/pkg/proxy/policy/policy.go
+++ b/pkg/proxy/policy/policy.go
@@ -2,6 +2,7 @@
 package policy
 
 import (
+	"encoding/json"
 	"fmt"
 	"log"
 	"net/http"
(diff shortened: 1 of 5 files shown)


pkg/proxy/proxy/transparent.go
```diff
@@ -165,7 +179,7 @@ func (t TransparentProxyService) ProxyTLS(addr string) {
 	}
 }
 
-func (t *TransparentProxyService) ServeMetadata(addr string) {
+func (t *TransparentProxyService) ServeAdminEndpoint(addr string) {
```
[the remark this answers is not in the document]
andusy: Hmm, if we're speaking about the KoRBE use case then I don't think we need to secure the endpoint, since the user could bypass the proxy in other ways if they broke out of the docker container. This would probably just add some overhead in maintenance on how we call the proxy.

What type of auth mechanism did you have in mind?

pkg/proxy/policy/policy.go
```diff
@@ -10,10 +11,60 @@ import (
 	"github.com/elazarl/goproxy"
 )
 
+var ruleRegistry = map[string]func() Rule{}
+
+// RegisterRule adds a rule to the rule registry.
+// The rule registry contains implementations of the Rule interface recognized
+// by the proxy. Updates to the policy will try to unmarshal json structs into the
+// registered rules. Expects the rule name and a constructor returning the Rule type.
+func RegisterRule(rulename string, constructor func() Rule) {
```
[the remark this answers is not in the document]
msuozzo: oh yep I didn't think that through. i guess the only way i can think of to avoid this would be to push the implementation cost into the Rule interface by requiring a "Name" or something for the string ID and a "Make" or "New" method (gross, I know) to provide the constructor. Other than that, looks like reflection would be required :/

all in all, I think the current design is fine

pkg/proxy/policy/policy.go
```diff
@@ -24,51 +24,57 @@ func RegisterRule(rulename string, constructor func() Rule) {
...
+	if err != nil {
+		return err
+	}
+	p.AnyOf = rules
+	return nil
+}
 
-	for _, rule := range policywrapper.Policy.AnyOf {
-		var tmpmap map[string]interface{}
+// unmarshalRule tries to unmarshal a rule json into the registered rule types.
```
msuozzo, line 49: fix docs

pkg/proxy/policy/policy.go
```diff
@@ -24,51 +24,57 @@ func RegisterRule(rulename string, constructor func() Rule) {
...
 		Policy struct {
-			AnyOf []json.RawMessage `json:"AnyOf"`
-		} `json:"Policy"`
+			AnyOf []json.RawMessage
+		}
 	}
 	if err := json.Unmarshal(data, &policywrapper); err != nil {
 		return err
 	}
+	rules, err := unmarshalRuleType(policywrapper.Policy.AnyOf)
```
msuozzo, line 41: can you make this function scalar? that way you can just progressively build AnyOf instead of assigning to it at the end

pkg/proxy/policy/policy.go
```diff
@@ -95,7 +101,7 @@ const (
 type URLMatchRule struct {
 	Host      string       `json:"host"`
 	Path      string       `json:"path"`
-	PathMatch MatchingType `json:"matching_type"`
+	PathMatch MatchingType `json:"matchingType"`
```
msuozzo, line 104: hmm it's not clear what matching this controls. I think the json key should say "path" somehow. Maybe "matchPath" or "matchPathBy" and then "prefix" and "full" would be a bit clearer

pkg/proxy/proxy/transparent.go
```diff
@@ -98,28 +97,17 @@ type TransparentProxyService struct {
 }
 
 // NewTransparentProxyService creates a new TransparentProxyService.
-func NewTransparentProxyService(p *goproxy.ProxyHttpServer, ca *tls.Certificate, mode PolicyMode, policyFile string) TransparentProxyService {
+func NewTransparentProxyService(p *goproxy.ProxyHttpServer, ca *tls.Certificate, mode PolicyMode, pl *policy.Policy) TransparentProxyService {
```
msuozzo, line 100: might be nice to make this a bit more defensive. if PolicyMode isn't Disabled, we should validate that the policy isn't nil. Alternatively, you could accept the policy by value instead of ptr. The former is probably sufficient.

msuozzo (approved):
awesome!

pkg/proxy/proxy/transparent.go
andusy: I was thinking that since the behaviour of enforcement mode is to have an allowlist rather than a blocklist, a nil policy should be treated as valid and just block everything.

pkg/proxy/policy/policy.go
andusy: I just renamed the function to make it more clear what it is trying to do.

pkg/proxy/proxy/transparent.go
msuozzo: well it'll currently crash since it'll try to call Apply on nil so we need to fix it some way. i think i'd still lean towards doing the validation/defaulting here so if you'd like to keep the param nil, it'd need to create an empty instance.

andusy: Oh good point. Added a check

pkg/proxy/proxy/transparent.go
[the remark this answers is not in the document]
andusy: Will follow up on this in the future if we need to add auth

pkg/proxy/policy/policy.go
msuozzo: wait I don't think this got added. the loadRules still operates on the array, right? I think it should just be loadRule and take the RawMessage and return the Rule.

PR #112, opened 2024-10-08
Add graceful shutdown to network proxy

andusy:


pkg/proxy/proxy/transparent.go
```diff
@@ -117,60 +118,104 @@ func NewTransparentProxyService(p *goproxy.ProxyHttpServer, ca *tls.Certificate,
...
+	var inflightRequests sync.WaitGroup
+	t.Proxy.OnResponse().DoFunc(func(resp *http.Response, ctx *goproxy.ProxyCtx) *http.Response {
+		defer inflightRequests.Done()
+		return resp
+	})
+	done := make(chan bool)
+	go func() {
+		for {
+			select {
+			case <-done:
```
msuozzo, line 153: I think in the steady state, this thread will block at line 157's `ln.Accept()` so you could detect net.ErrClosed as the error return instead to terminate the thread. That would remove the select and the extra channel creation which should simplify things

pkg/proxy/proxy/transparent.go
```diff
@@ -117,60 +118,104 @@ func NewTransparentProxyService(p *goproxy.ProxyHttpServer, ca *tls.Certificate,
...
+					log.Println("Connection from ", c.RemoteAddr())
+					conn, hello, err := handshake.PeekClientHello(c)
+					if err != nil {
+						log.Printf("Error accepting new connection - %v", err)
+						return
+					}
+					host := hello.ServerName
+					if host == "" {
+						log.Printf("Cannot support non-SNI enabled clients")
+						c.Close()
```
msuozzo, line 173: inflightRequests won't get updated here or in the block above. I think you should add a `defer inflightRequests.Done()` at the top of the function to be sure it's resolved. (we'd need to make sure this code is respectful of some deadline/timeout to be entirely confident in the shutdown not hanging but I think it's fine as-is)

andusy: Good point. I originally had `defer inflightRequests.Done()` at the top but realized this doesn't wait for the request to actually complete. Instead I added it in the `OnResponse` callback, but looking at the code goproxy actually returns a response before the request is complete. So I think a graceful shutdown is not possible unless we make upstream changes 🙃 

I just put `defer inflightRequests.Done()` for now so it doesn't hang if there's an error somewhere.

pkg/proxy/proxy/transparent.go
```diff
@@ -115,31 +120,82 @@ func NewTransparentProxyService(p *goproxy.ProxyHttpServer, ca *tls.Certificate,
...
+		ch := make(chan struct{})
+		errChan := make(chan error)
+		go func() {
+			if err := ln.Close(); err != nil {
+				errChan <- err
+				return
+			}
+			log.Println("Waiting for in-flight requests to complete...")
+			inflightRequests.Wait()
+			close(ch)
```
msuozzo, line 176: I think you can just send nil to the errChan here instead and get rid of ch. as a result you could rename if you wanted.

msuozzo (approved):

pkg/proxy/proxy/transparent.go
```diff
@@ -115,31 +120,82 @@ func NewTransparentProxyService(p *goproxy.ProxyHttpServer, ca *tls.Certificate,
...
+			return err
+		}
+	}
+	return nil
+}
+
 // ProxyHTTP serves an endpoint that transparently redirects HTTP connections to the proxy server.
 // This endpoint also explicitly (i.e. non-transparently) proxies HTTP and TLS connections.
-func (t TransparentProxyService) ProxyHTTP(addr string) {
+func (t *TransparentProxyService) ProxyHTTP(addr string) {
```
msuozzo, line 144: hmm actually it comes to mind that multiple calls to this function would fail to tear down the prior versions as it would override the shutdown func member. you could make it a slice and append entries instead

andusy: Created a slice and just had all the shutdown functions appended to this

PR #113, opened 2024-10-08
Add option to disable capturing network activity log

andusy:
Add TransparentProxyServiceOpts struct to disable network activity logging in order to avoid unbounded memory usage.

### DIFF
--- a/pkg/proxy/proxy/transparent.go
+++ b/pkg/proxy/proxy/transparent.go
@@ -125,17 +125,6 @@ func NewTransparentProxyService(p *goproxy.ProxyHttpServer, ca *tls.Certificate,
 	}
 }
 
-// RotateLog clears the activity log if logging is enabled. Otherwise starts capturing network activity.
-func (t *TransparentProxyService) RotateLog() {
-	t.mx.Lock()
-	defer t.mx.Unlock()
-	if t.networkLog == nil {
-		t.networkLog = netlog.CaptureActivityLog(t.Proxy, t.mx)
-	} else {
-		t.networkLog.HTTPRequests = []netlog.HTTPRequestLog{}
-	}
-}
-
 // ProxyHTTP serves an endpoint that transparently redirects HTTP connections to the proxy server.
 // This endpoint also explicitly (i.e. non-transparently) proxies HTTP and TLS connections.
 func (t TransparentProxyService) ProxyHTTP(addr string) {
--- a/pkg/proxy/proxy/transparent.go
+++ b/pkg/proxy/proxy/transparent.go
@@ -98,29 +98,41 @@ type TransparentProxyService struct {
 
 // TransparentProxyServiceOpts defines the optional parameters for creating a TransparentProxyService.
 type TransparentProxyServiceOpts struct {
-	SkipLogInit bool
+	Policy      *policy.Policy
+	SkipLogging bool
 }
 
 // NewTransparentProxyService creates a new TransparentProxyService.
-func NewTransparentProxyService(p *goproxy.ProxyHttpServer, ca *tls.Certificate, mode PolicyMode, pl *policy.Policy, opts TransparentProxyServiceOpts) TransparentProxyService {
+func NewTransparentProxyService(p *goproxy.ProxyHttpServer, ca *tls.Certificate, mode PolicyMode, opts TransparentProxyServiceOpts) TransparentProxyService {
 	m := new(sync.Mutex)
 	if !mode.IsValid() {
 		log.Fatalf("Invalid proxy mode specified: %v", mode)
 	}
-	if mode != DisabledMode && pl == nil {
-		log.Fatalf("Invalid policy: %v", pl)
+	if mode != DisabledMode && opts.Policy == nil {
+		log.Fatalf("Invalid policy: %v", opts.Policy)
 	}
 	networkLog := &netlog.NetworkActivityLog{}
-	if !opts.SkipLogInit {
+	if !opts.SkipLogging {
 		networkLog = netlog.CaptureActivityLog(p, m)
 	}
 	return TransparentProxyService{
 		Proxy:      p,
 		Ca:         ca,
 		Mode:       mode,
-		Policy:     pl,
+		Policy:     opts.Policy,
 		mx:         m,
 		networkLog: networkLog,
+	}
+}
+
+// RotateLog clears the activity log if logging is enabled. Otherwise starts capturing network activity.
+func (t *TransparentProxyService) RotateLog() {
+	t.mx.Lock()
+	defer t.mx.Unlock()
+	if t.networkLog == nil {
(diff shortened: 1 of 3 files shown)


pkg/proxy/proxy/transparent.go
```diff
@@ -108,13 +108,18 @@ func NewTransparentProxyService(p *goproxy.ProxyHttpServer, ca *tls.Certificate,
 	return TransparentProxyService{
 		Proxy:      p,
 		Ca:         ca,
-		NetworkLog: netlog.CaptureActivityLog(p, m),
+		NetworkLog: &netlog.NetworkActivityLog{},
```
msuozzo, line 111: I think I might prefer this being nil on creation. Thoughts?

pkg/proxy/proxy/transparent.go
```diff
@@ -108,13 +108,18 @@ func NewTransparentProxyService(p *goproxy.ProxyHttpServer, ca *tls.Certificate,
...
 		Mode:       mode,
 		Policy:     pl,
 		mx:         m,
 	}
 }
 
+// LogActivity starts recording network activity in memory.
+func (t *TransparentProxyService) LogActivity() {
+	t.NetworkLog = netlog.CaptureActivityLog(t.Proxy, t.mx)
```
msuozzo, line 120: are you sure calling this again will free the underlying memory of the replaced object? won't the prior goroutine still be running, populating the object? some means of preempting that process (a context?) might also be required.

pkg/proxy/proxy/transparent.go
andusy: I think either works here. I just added an empty object here in case the `/summary` gets called with this uninitialized. Seemed nicer to have something like 
```
{
  "HTTPRequests": null
}
```
rather than just having
```
null
```
But I'm ok either way.

pkg/proxy/proxy/transparent.go
```diff
@@ -108,13 +108,18 @@ func NewTransparentProxyService(p *goproxy.ProxyHttpServer, ca *tls.Certificate,
...
-		NetworkLog: netlog.CaptureActivityLog(p, m),
+		NetworkLog: &netlog.NetworkActivityLog{},
 		Mode:       mode,
 		Policy:     pl,
 		mx:         m,
 	}
 }
 
+// LogActivity starts recording network activity in memory.
+func (t *TransparentProxyService) LogActivity() {
```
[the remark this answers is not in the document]
andusy: Tbh I don't really like having a function like this so I'm open to changing it. My original thought was to add another constructor like
```
func NewTransparentProxyServiceWithLogging(p *goproxy.ProxyHttpServer, ca *tls.Certificate, mode PolicyMode, pl *policy.Policy) TransparentProxyService {}
```
which is essentially the same thing excepts it initializes the logger and the original constructor wouldn't. Wasn't sure if it was worth it to add another constructor though. wdyt?

pkg/proxy/proxy/transparent.go
msuozzo: ah right we just fixed that bug in the first place... Okay yeah I think initializing it sounds good.

Would it be possible to make this a private member though? Exposing the log through a method should avoid exposing the implementation details too much now that we're supporting "resetting".

one bigger change would be to move the aggregation up to this level. CaptureActivityLog could just be responsible for providing a channel of log entries and the summary construct could be populated here. If we want to add logic related to memory usage monitoring, doing so here would be a better fit than in the netlog package.

pkg/proxy/proxy/transparent.go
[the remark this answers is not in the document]
andusy: I just added a `TransparentProxyServiceOpts` struct for now. Not sure there's a use case for resetting the activity log currently so I'll maybe add that in the future. I've made NetworkLog private anyways since the mutex is also private so it doesn't make sense to create it outside of this class.

pkg/proxy/proxy/transparent.go
andusy: Made networkLog private.

pkg/proxy/proxy/transparent.go
andusy: I think this is obsolete now that there is no new function.

pkg/proxy/proxy/transparent.go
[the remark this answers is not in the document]
msuozzo: hmm but without this method, there would be no way to start logging after creation (you'd need to name it "SkipLogging" instead). Given the current change, I think having a "RotateLog" method would fit nicely.

pkg/proxy/proxy/transparent.go
```diff
@@ -87,31 +87,40 @@ func (m PolicyMode) IsValid() bool {
...
+}
+
+// TransparentProxyServiceOpts defines the optional parameters for creating a TransparentProxyService.
+type TransparentProxyServiceOpts struct {
+	SkipLogInit bool
 }
 
 // NewTransparentProxyService creates a new TransparentProxyService.
-func NewTransparentProxyService(p *goproxy.ProxyHttpServer, ca *tls.Certificate, mode PolicyMode, pl *policy.Policy) TransparentProxyService {
+func NewTransparentProxyService(p *goproxy.ProxyHttpServer, ca *tls.Certificate, mode PolicyMode, pl *policy.Policy, opts TransparentProxyServiceOpts) TransparentProxyService {
```
msuozzo, line 105: i was also imagining the policy configuration moving to the Opt struct but we could also consider that a follow-on.

pkg/proxy/proxy/transparent.go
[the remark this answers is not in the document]
andusy: I renamed it to SkipLogging and added a RotateLog function.

pkg/proxy/proxy/transparent.go
andusy: I moved the policy inside the opt

pkg/proxy/proxy/transparent.go
```diff
@@ -87,31 +87,52 @@ func (m PolicyMode) IsValid() bool {
...
 		mx:         m,
+		networkLog: networkLog,
+	}
+}
+
+// RotateLog clears the activity log if logging is enabled. Otherwise starts capturing network activity.
+func (t *TransparentProxyService) RotateLog() {
+	t.mx.Lock()
+	defer t.mx.Unlock()
+	if t.networkLog == nil {
```
msuozzo, line 132: we don't expect this, do we? can we just blindly assign to CaptureActivityLog?

msuozzo (approved):

pkg/proxy/proxy/transparent.go
andusy: Filed https://github.com/google/oss-rebuild/issues/116 and removed this function for now.

msuozzo (approved):

PR #115, opened 2024-10-10
Use benchmark.Run for ctl run-one

wbxyz:
This removes the need for schema.MakeHttpRequest

### DIFF
--- a/tools/ctl/ctl.go
+++ b/tools/ctl/ctl.go
@@ -370,8 +370,7 @@ var runOne = &cobra.Command{
 		}
 		set := benchmark.PackageSet{
 			Metadata: benchmark.Metadata{
-				Count:   1,
-				Updated: time.Now(),
+				Count: 1,
 			},
 			Packages: []benchmark.Package{
 				{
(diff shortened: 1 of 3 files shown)


tools/ctl/ctl.go
```diff
@@ -368,40 +368,34 @@ var runOne = &cobra.Command{
...
-			})
-			if err != nil {
-				log.Fatal(err)
-			}
-		}
-		resp, err := client.Do(req)
+		set := benchmark.PackageSet{
+			Metadata: benchmark.Metadata{
+				Count:   1,
+				Updated: time.Now(),
```
wbxyz, line 374: I was thinking, maybe this updated time should be some default constant. That way the on-off benchmarks always have the same hash if they're referring to the same target.

msuozzo: Yeah you should be able to just omit it, right? To be more confident it'll serialize properly, we could omitempty on the struct (although I'm not sure that's implemented in the form schema).

msuozzo (approved):

tools/ctl/ctl.go
wbxyz: I think the zero value for datetime should probably be good enough? Deleted for now, if it bothers us in the future we can investigate more.

PR #119, opened 2024-10-10
Add form tests for bool fields

msuozzo:


pkg/rebuild/schema/form/form_test.go
```diff
@@ -11,6 +11,7 @@ type TestStruct struct {
 	IntField      int      `form:"int_field"`
 	SliceField    []string `form:"slice_field"`
 	IntSliceField []int    `form:""`
+	BoolField     bool     `form:"bool_field"`
```
wbxyz, line 14: Is `false` just omitted when marshaled to url.Values? It might be illustrative to have an example of `false` somewhere.

msuozzo: `false` is no different from `true` here as they're both spelled the same in json and golang. not sure of the value of testing it in particular but I've added it as a case in the Unmarshal test. omitting the field will default it to its zero value and that case is tested in the one named `Pointer to struct`.

wbxyz (approved):

PR #126, opened 2024-10-14
Support a local debug asset store

wbxyz:
This change updates the flag names, and many of the handling locations to support both gs:// and file:// URI types.

One open question I still have is whether the local file assets store should consider the RunID when deciding the filepath for storage. I'm thinking it should, especially because TUI will download files for the same target across multiple runs. However maybe this could be implemented at the localfiles level (constructing the local files asset store takes a runid and each runid has a dedicated folder with a separate asset store).


pkg/rebuild/rebuild/storage.go
```diff
@@ -87,18 +89,24 @@ type LocatableAssetStore interface {
...
 		return errors.Wrap(err, "from.Reader failed")
 	}
-	defer r.Close()
+	defer logWrap(r.Close)
 	w, err := to.Writer(ctx, a)
 	if err != nil {
 		return errors.Wrap(err, "to.Writer failed")
 	}
-	defer w.Close()
+	defer logWrap(w.Close)
```
msuozzo, line 109: you could also actually return the error here, right?

tools/ctl/localfiles/localfiles.go
```diff
@@ -11,20 +11,23 @@ import (
 	"github.com/pkg/errors"
 )
 
-// These are the different folders used for local storage.
-// 		./assets/		// AssetStore used to cache logs, artifacts, and other assets manipulated by ctl.
-//		./rundex/       // The local metadata about runs and attempts, normally generated by local smoketets benchmarks.
-
 const (
-	TempRoot = "/tmp/oss-rebuild/"
+	tempRoot = "/tmp/oss-rebuild/" // The directory where all the local files are stored
```
msuozzo, line 15: again, this default should not be here. a better design would be to pass around a chroot with this abspath such that it's configured at as high a level as possible (i.e. injected at the program root). Having this be defined within a library is a code smell.

tools/docker/docker.go
```diff
@@ -27,13 +27,18 @@ import (
 
 // RunOptions defines optional arguments for RunServer.
 type RunOptions struct {
-	ID     chan<- string
-	Output io.Writer
+	ID         chan<- string
+	Output     io.Writer
+	DockerArgs []string
```
msuozzo, line 32: instead, I think adding a Mount member here is better. Keeping `Args` the way it is would be fine.

pkg/rebuild/rebuild/storage.go
```diff
@@ -87,18 +89,24 @@ type LocatableAssetStore interface {
...
+}
+
 // AssetCopy copies an asset from one store to another.
 func AssetCopy(ctx context.Context, to, from AssetStore, a Asset) error {
 	r, err := from.Reader(ctx, a)
 	if err != nil {
 		return errors.Wrap(err, "from.Reader failed")
 	}
-	defer r.Close()
+	defer logWrap(r.Close)
```
msuozzo, line 104: do we need this? I'd like to avoid fire-and-forget logging like this. would it be possible to tie this in to the return value?

tools/ctl/ctl.go
```diff
@@ -102,25 +102,27 @@ func buildFetchRebuildRequest(ctx context.Context, bench, run, filter string, cl
...
 		var fireClient rundex.Reader
 		if *benchmarkDir != "" {
 			fireClient = rundex.NewLocalClient()
 		} else {
-			if *debugBucket != "" {
-				bucket, prefix, _ := strings.Cut(strings.TrimPrefix(*debugBucket, "gs://"), string(filepath.Separator))
-				if prefix != "" {
-					log.Fatalf("--debug-bucket cannot have additional path elements, found %s", prefix)
+			if *debugStorage != "" {
+				if strings.HasPrefix(*debugStorage, "gs://") {
```
msuozzo, line 119: maybe parse as a URL instead? looking back at this, that seems like a cleaner solution that prefix matching

pkg/rebuild/rebuild/storage.go
```diff
@@ -111,6 +119,9 @@ func DebugStoreFromContext(ctx context.Context) (AssetStore, error) {
 		if strings.HasPrefix(uploadpath, "gs://") {
```
msuozzo, line 119: same comment as below. maybe use URL parsing here?

tools/ctl/localfiles/localfiles.go
wbxyz: I don't think I have super strong opinions on this, but I think I'd prefer to punt to a separate PR. That change would affect a few other usages of the localfiles and further affect the signatures in the library.

pkg/rebuild/rebuild/storage.go
wbxyz: Used URL parsing for the scheme but it seemed more straightforward to use uploadpath to actually build the AssetStore, using the URL elements would have required some concatenation and would silently drop params.

pkg/rebuild/rebuild/storage.go
```diff
@@ -102,15 +103,25 @@ func AssetCopy(ctx context.Context, to, from AssetStore, a Asset) error {
...
-		if strings.HasPrefix(uploadpath, "gs://") {
+		u, err := url.Parse(uploadpath)
+		if err != nil {
+			return nil, errors.Wrap(err, "parsing UploadArtifactsPathID as url")
+		}
+		if u.Scheme == "gs" {
 			storer, err := NewGCSStore(ctx, uploadpath)
 			return storer, errors.Wrapf(err, "Failed to create GCS storer")
+		} else if u.Scheme == "file" {
+			path, _ := strings.CutPrefix(uploadpath, "file://")
```
msuozzo, line 120: can you use u.Path?

tools/ctl/ctl.go
```diff
@@ -102,25 +102,31 @@ func buildFetchRebuildRequest(ctx context.Context, bench, run, filter string, cl
...
-				bucket, prefix, _ := strings.Cut(strings.TrimPrefix(*debugBucket, "gs://"), string(filepath.Separator))
-				if prefix != "" {
-					log.Fatalf("--debug-bucket cannot have additional path elements, found %s", prefix)
+			if *debugStorage != "" {
+				u, err := url.Parse(*debugStorage)
+				if err != nil {
+					log.Fatal(errors.Wrap(err, "parsing --debug-storage as url"))
+				}
+				if u.Scheme == "gs" {
+					prefix := strings.TrimPrefix(u.Path, string(filepath.Separator))
```
msuozzo, line 124: nit: this should actually be path.Separator since this shouldn't vary by OS

tools/ctl/ide/rebuilder.go
```diff
@@ -98,7 +99,18 @@ func (in *Instance) Run(ctx context.Context) {
 		in.state = running
 		idchan := make(chan string)
 		go func() {
-			err = docker.RunServer(ctx, "rebuilder", 8080, &docker.RunOptions{ID: idchan, Output: logWriter(rblog)})
+			localAssets := localfiles.AssetsPath()
```
msuozzo, line 102: nit: variable name-wise, maybe "assetPath" or "assetDir"?

tools/ctl/localfiles/localfiles.go
msuozzo: yeah separate PR is fine

msuozzo (approved):

pkg/rebuild/rebuild/storage.go
wbxyz: I thought I couldn't because I was doing "file://a/b/c" which cuts "a" as the host and "/b/c" as the path, but reading the spec I think the format I was using is incorrect. Instead I should have a leading slash on the path section for three slashes in a row: "file:///a/b/c" which should work.

tools/ctl/ctl.go
msuozzo: oops yeah that's perfect

PR #131, opened 2024-10-23
Filter local runs when loading rebuilds.

wbxyz:
The change looks bigger than it is, most of it is an indentation and for loop.

### DIFF
--- a/tools/ctl/rundex/rundex.go
+++ b/tools/ctl/rundex/rundex.go
@@ -382,16 +382,16 @@ func (f *LocalClient) FetchRebuilds(ctx context.Context, req *FetchRebuildReques
 	walkErr := make(chan error, 1)
 	all := make(chan Rebuild, 1)
 	go func() {
-		var pathsToExplore []string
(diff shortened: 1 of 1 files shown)


tools/ctl/rundex/rundex.go
```diff
@@ -382,31 +382,44 @@ func (f *LocalClient) FetchRebuilds(ctx context.Context, req *FetchRebuildReques
...
 	go func() {
-		walkErr <- util.Walk(f.fs, localRunsDir, func(path string, info fs.FileInfo, err error) error {
-			if err != nil {
-				return err
-			}
-			// TODO: Skip run directories that don't match req.Runs
-			// This is just an optimization, filterRebuilds will also drop Rebuilds that don't match req.Runs.
-			if info.IsDir() {
-				return nil
+		var pathsToExplore []string
```
msuozzo, line 385: what about `toWalk`?

msuozzo (approved):

Issue #39, opened 2024-06-16
Make normalization logic more accessible


hboutemy:
current attestations are done just after rebuild, on zip files that are de-facto not reproducible because they contain build timestamp

normalizing timestamp with `ziptime` for example (or any equivalent tool) would ease rebuilding locally and confirming results

rebuild instructions could even download reference archive from registries and apply the normalization tool then check that normalized rebuild output is exactly the same as normalized reference archive

msuozzo:
Hello!

To briefly summarize the current two-staged rebuild process:

1. **Faithful build** of the upstream package (note: this is encapsulated in the Dockerfile we provide)
2. **Normalization process** that takes in both upstream and rebuild and applies transforms to canonicalize the artifacts

The faithful build is intended to match how the upstream builds and, as such, will not have any normalizations (e.g. zip build time set, anything in `strip-nondeterminism`) applied. One reason for this separation is to run the build in an untrusted sandbox while the (security-critical) normalization can happen in trusted code.

---

To try to get at the root of the issue: I think you're (reasonably) asking "Why isn't the provided build deterministic/reproducible?"

To attempt an answer: The two stages of the rebuild process together _are_ deterministic but we don't currently provide an easy way to run the second normalization step.

To look forward to a solution: I think users should be able to easily run _both_ stage 1 alone _and_ stage 1+2 so we should take an FR to expose that.

---

Let me know if that answers your question and thanks for filing the issue!

hboutemy:
yes, this perfectly answers the question, thank you
and waiting for the feature, this does the minimum documentation that was necessary :)

Issue #83, opened 2024-08-21
Expose the exact normalizations required to achieve reproduction


msuozzo:
We currently apply all relevant normalizations to all artifacts meaning we neither know nor convey to the user which were the minimal set of normalizations that ended up being necessary to achieve an identical artifact.

Determining the minimal set of normalizations could be tricky but i think the best strategy would be to decompose our monolithic normalizer into a set of "passes" and, as we're normalizing each artifact, evaluate the similarity of the artifacts at each step. When passes 'do nothing', the degree of similarity should remain unchanged when the pass is applied.

I don't believe this approach would work for passes applied in a random order but I do think we could order them such that the similarity would monotonically increase.

Credit to @hboutemy for the idea 👍

PR #132, opened 2024-10-24
Separate out stabilization into configurable passes

msuozzo:
Working towards a solution for #39

### DIFF
--- a/internal/verifier/attestation.go
+++ b/internal/verifier/attestation.go
@@ -68,6 +68,7 @@ func CreateAttestations(ctx context.Context, input rebuild.Input, finalStrategy 
 		// TODO: Include build repository associated with this builder.
 	}
 	publicRebuildURI := path.Join("rebuild", buildInfo.Target.Artifact)
+	// TODO: Change from "normalized" to "stabilized".
 	publicNormalizedURI := path.Join("normalized", buildInfo.Target.Artifact)
 	// Create comparison attestation.
 	eqStmt := &in_toto.ProvenanceStatementSLSA1{
--- a/pkg/archive/tar.go
+++ b/pkg/archive/tar.go
@@ -20,10 +20,10 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"io"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"slices"
-	"sort"
 	"strings"
 	"time"
 
@@ -34,30 +34,6 @@ import (
 // Pick some arbitrary time to set all the time fields.
 // Source: https://github.com/npm/pacote/blob/main/lib/util/tar-create-options.js#L28
 var arbitraryTime = time.Date(1985, time.October, 26, 8, 15, 0, 0, time.UTC)
-
-func stabilizeTarHeader(h *tar.Header) (*tar.Header, error) {
-	switch h.Typeflag {
-	case tar.TypeGNUSparse, tar.TypeGNULongName, tar.TypeGNULongLink:
-		// NOTE: Non-PAX header type support can be added, if necessary.
-		return nil, errors.Errorf("Unsupported header type: %v", h.Typeflag)
-	default:
-		return &tar.Header{
-			Typeflag:   h.Typeflag,
-			Name:       h.Name,
-			ModTime:    arbitraryTime,
-			AccessTime: arbitraryTime,
-			// TODO: Surface presence/absence of execute bit as a comparison config.
-			Mode:  0777,
-			Uid:   0,
-			Gid:   0,
-			Uname: "",
-			Gname: "",
-			Size:  h.Size,
-			// TODO: Surface comparison config for TAR metadata (PAXRecords, Xattrs).
-			Format: tar.FormatPAX,
-		}, nil
-	}
-}
 
 // TarEntry represents an entry in a tar archive.
 type TarEntry struct {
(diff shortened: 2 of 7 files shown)


docs/builds/ArtifactEquivalence@v0.1.md
wbxyz, line 1: These doc changes should maybe be moved to the rebrand PR?

internal/verifier/attestation.go
wbxyz, line 1: Rebrand PR?

pkg/archive/tar.go
```diff
@@ -76,10 +48,88 @@ func (e TarEntry) WriteTo(tw *tar.Writer) error {
...
+
+var StableTarTime = TarEntryStabilizer{
+	Name: "tar-time",
+	Func: func(e *TarEntry) {
+		e.ModTime = time.UnixMilli(0)
+		e.AccessTime = time.UnixMilli(0)
+		e.ChangeTime = time.Time{}
+		// NOTE: Without a PAX record, the tar library will disregard this value
+		// and write a USTAR-formatted file. Setting 'atime' ensures at least one
+		// record exists which will cause tar to serialize and re-parse it as PAX.
```
wbxyz, line 89: I'm a little confused by the comment. The second half of the comment seems to describe that we must set atime as a way of forcing PAX but then why do we explicitly set FormatPAX? Also, it might be helpful to explain why we choose PAX instead of USTAR as the stable format.

pkg/archive/tar_test.go
```diff
@@ -24,6 +24,8 @@ import (
 	"github.com/google/go-cmp/cmp"
 )
 
+var epoch = time.UnixMilli(0)
```
wbxyz, line 27: const?

pkg/archive/zip.go
```diff
@@ -68,40 +70,156 @@ func (e ZipEntry) WriteTo(zw *zip.Writer) error {
...
+type MutableZipReader struct {
+	*zip.Reader
+	File []*MutableZipFile
+}
+
+func NewMutableReader(zr *zip.Reader) MutableZipReader {
+	mr := MutableZipReader{Reader: zr}
+	for _, zf := range zr.File {
+		mr.File = append(mr.File, &MutableZipFile{File: zf, FileHeader: zf.FileHeader})
+	}
```
wbxyz, line 101: Does the MutableReader need to keep the underlying reader if it already has an in-memory copy of all the files?

pkg/archive/zip.go
```diff
@@ -68,40 +70,156 @@ func (e ZipEntry) WriteTo(zw *zip.Writer) error {
...
+	Name: "zip-compression",
+	Func: func(zf *MutableZipFile) {
+		zf.Method = zip.Store
+	},
+}
+
+var StableZipDataDescriptor = ZipEntryStabilizer{
+	Name: "zip-data-descriptor",
+	Func: func(zf *MutableZipFile) {
+		zf.Flags = zf.Flags & ^uint16(0x8)
```
wbxyz, line 170: Could you add a comment for which flag this is changing?

pkg/archive/zip.go
```diff
@@ -68,40 +70,156 @@ func (e ZipEntry) WriteTo(zw *zip.Writer) error {
...
+		zf.ExternalAttrs = 0
+	},
+}
+
+var StableZipMisc = ZipEntryStabilizer{
+	Name: "zip-misc",
+	Func: func(zf *MutableZipFile) {
+		zf.Comment = ""
+		zf.ReaderVersion = 0
+		zf.Flags = zf.Flags & uint16(0x8)
```
wbxyz, line 199: Related to the similar line in StableZipDataDescriptor?

wbxyz (approved):
This is cool! It's really helpful to have the different considerations separated out and encapsulated with a description.

Do you think eventually we'll be able to tell which stabilizers actually made modifications? I was imagining the stabilizers could return a modified copy instead of modifying in place. That way the invocation of the stabilizer can compare before-after, but that would be expensive to copy the file on every stabilizer invocation.

I guess another option would be for the stabilizers to return a bool of whether they made a modification. It would be more efficient but potentially less robust and more complex.

pkg/archive/tar.go
msuozzo: > we must set a time as a way of forcing PAX

Kind of. A PAX-format with no records IS a ustar-format archive. We use this strategy as a way of making working with the library more consistent I.e. format will actually be respected when we set it.

> why we choose PAX instead of USTAR as the stable format

The underlying blocks are identical between the two. PAX specifies how to provide additional records. So really, we _can't_ use ustar if there are, or we need, additional records (notably, required for precise timestamps and GNU tar compatibility).

pkg/archive/zip.go
msuozzo: Yeah this just blanks out all flags except the data descriptor so they don't overlap in their functionality.

pkg/archive/zip.go
msuozzo: I mean, the third bit? the "data descriptor" flag? what do you mean?

msuozzo:
> Do you think eventually we'll be able to tell which stabilizers actually made modifications?

Yeah it's certainly possible but I don't think it's a particular priority. The current way, we provide the document the passes we do apply at a given time. Next, we would have to provide feedback from both the rebuilt and upstream artifacts, take the super-set of the applied passes, and report those. Even that could be confusing given we'd report the same set of passes for both while the changes would often just reflect variability in the upstream artifact's metadata.

Further, this wouldn't necessarily provide valuable feedback to our users or to upstream maintainers. If our argument is that these differences aren't security-critical, diverting time/resources to resolving them isn't productive.

pkg/archive/tar_test.go
msuozzo: ./prog.go:8:15: time.UnixMilli(0) (value of type time.Time) is not constant

pkg/archive/zip.go
wbxyz: Yeah, I mean giving the name of the flag. The rest of the behavior is readable in English but zeroing the third bit seems like a magic number.

pkg/archive/zip.go
wbxyz: It's probably worth adding that as a comment "Preserve the data descriptor flag to avoid overlap with zip-data-descriptor" or something?

wbxyz (approved):

wbxyz:
I think I was thinking about https://github.com/google/oss-rebuild/issues/83 but agreed that it's not high priority at the moment.

msuozzo:
> I think I was thinking about #83 but agreed that it's not high priority at the moment.

Yeah I read that, as well. I think that's more the "next step" as I was saying and less a near-term priority.

pkg/archive/zip.go
msuozzo: it's a bit tricky. While it doesn't need to, there are a number of methods that the original exports that we get to inherit by retaining the pointer. and given the name "MutableZipReader", departing from the actual supported API of the reader would be suboptimal and so would be better served by changing the name. MutableZipFile re-implements the original API in spite of not holding on to the original pointer.

all things considered, I don't think it's terribly valuable removing it but it's not strictly required by the stabilizers.

pkg/archive/zip.go
msuozzo: Done

pkg/archive/tar.go
msuozzo: tweaked this to make this more clear

pkg/archive/zip.go
msuozzo: eh actually naming it also enables us to do the cast once instead of twice. changed.

PR #134, opened 2024-10-25
Rebrand 'canonical' to 'stable' for artifacts

msuozzo:
Since we're in the process of separating out the individual operations we're applying to artifacts, there won't be a single 'canonical' form for the artifact. The set of operations necessary to achieve a match between the rebuild and upstream artifacts may not match that of another rebuild-upstream pair. As such, our terminology should be flexible enough to encompass a spectrum of outcomes and, with this constraint, 'canonical' is no longer fit for purpose.

'normalized' was a close second but 'stabilized' is more permissive of this multiple endpoint usage. This term still appears in the ArtifactEquivalence attestation but will be migrated later in #135.

### DIFF
--- a/internal/verifier/attestation.go
+++ b/internal/verifier/attestation.go
@@ -96,7 +96,7 @@ func CreateAttestations(ctx context.Context, input rebuild.Input, finalStrategy 
 					InvocationID: id,
 				},
 				Byproducts: []slsa1.ResourceDescriptor{
-					{Name: publicNormalizedURI, Digest: makeDigestSet(up.CanonicalHash...)},
+					{Name: publicNormalizedURI, Digest: makeDigestSet(up.StabilizedHash...)},
 				},
 			},
 		},
--- a/internal/verifier/summary_test.go
+++ b/internal/verifier/summary_test.go
@@ -68,11 +68,11 @@ func TestSummarizeArtifacts(t *testing.T) {
 				}
 			},
 		})
-		canonicalizedHash := hashext.NewMultiHash(crypto.SHA256)
-		canonicalizedZip := must(archivetest.ZipFile([]archive.ZipEntry{
+		stabilizedHash := hashext.NewMultiHash(crypto.SHA256)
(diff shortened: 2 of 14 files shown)


internal/verifier/summary_test.go
```diff
@@ -68,11 +68,11 @@ func TestSummarizeArtifacts(t *testing.T) {
 				}
 			},
 		})
-		canonicalizedHash := hashext.NewMultiHash(crypto.SHA256)
-		canonicalizedZip := must(archivetest.ZipFile([]archive.ZipEntry{
+		stabilizedHash := hashext.NewMultiHash(crypto.SHA256)
+		stableZip := must(archivetest.ZipFile([]archive.ZipEntry{
```
wbxyz, line 72: nit: Why use "stabilized" as an adjective sometimes vs "stable" other times?

internal/verifier/attestation.go
```diff
@@ -96,7 +96,7 @@ func CreateAttestations(ctx context.Context, input rebuild.Input, finalStrategy
 					InvocationID: id,
 				},
 				Byproducts: []slsa1.ResourceDescriptor{
-					{Name: publicNormalizedURI, Digest: makeDigestSet(up.CanonicalHash...)},
+					{Name: publicNormalizedURI, Digest: makeDigestSet(up.StabilizedHash...)},
```
wbxyz, line 99: Are we keeping normalize in the public attestation?

wbxyz (approved):

internal/verifier/summary_test.go
msuozzo: here this was due to "stable hash" having a separate meaning while it's less ambiguous than "stable zip."

internal/verifier/attestation.go
msuozzo: For now, I'd vote to keep it to reduce churn. We hope to change the full attestation in the near to incorporate the new 'stabilize' binary better so we should revisit it again then.

PR #137, opened 2024-10-29
Add gzip stabilizer

msuozzo:


pkg/archive/gzip_test.go
```diff
@@ -0,0 +1,88 @@
...
+		name        string
+		input       []byte
+		header      gzip.Header
+		stabilizers []any
+		wantHeader  gzip.Header
+	}{
+		{
+			name:  "all stabilizers",
+			input: []byte("hello world"),
+			header: gzip.Header{
```
wbxyz, line 24: Add a compression level to verify that it's set to gzip.NoCompression?

wbxyz (approved):

pkg/archive/gzip_test.go
msuozzo: Done.

PR #142, opened 2024-11-06
Add dockerfile format to infer command

wbxyz:
This adds the `--format=dockerfile` option on the `infer` command. It's helpful to see the full dockerfile when debugging inference, and easy to pipe this directly into docker locally. For example:

```
go run tools/ctl/ctl.go infer --ecosystem=debian --package=main/file --version=5.45-3+b1 --artifact=libmagic-mgc_5.45-3+b1_amd64.deb --format=dockerfile | docker buildx build -t tmp - && docker run -it -t tmp
```

### DIFF
--- a/tools/ctl/ctl.go
+++ b/tools/ctl/ctl.go
@@ -455,7 +455,7 @@ var listRuns = &cobra.Command{
 }
 
 var infer = &cobra.Command{
-	Use:   "infer --ecosystem <ecosystem> --package <name> --version <version> [--artifact <name>] [--api <URI>] ",
+	Use:   "infer --ecosystem <ecosystem> --package <name> --version <version> [--artifact <name>] [--api <URI>] [--format strategy|dockerfile]",
 	Short: "Run inference",
 	Args:  cobra.NoArgs,
 	Run: func(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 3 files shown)


tools/ctl/ctl.go
```diff
@@ -462,6 +467,68 @@ var listRuns = &cobra.Command{
...
+		defer f.Close()
+		if _, err := f.WriteString(dockerfile); err != nil {
+			log.Fatal(err)
+		}
+		log.Println("File name: ", f.Name())
+		if err := f.Close(); err != nil {
+			log.Fatal(err)
+		}
+		build := exec.CommandContext(cmd.Context(), "docker", "buildx", "build", "--tag", "local-gcb", dir)
+		build.Stdout = log.Writer()
```
msuozzo, line 518: use cobra IO

tools/ctl/ctl.go
```diff
@@ -462,6 +467,68 @@ var listRuns = &cobra.Command{
...
+		}, rebuild.RemoteOptions{})
+		if err != nil {
+			log.Fatal(err)
+		}
+		log.Println("Dockerfile: ", dockerfile)
+		dir, err := os.MkdirTemp("/tmp", "")
+		if err != nil {
+			log.Fatal(err)
+		}
+		f, err := os.Create(filepath.Join(dir, "dockerfile"))
```
msuozzo, line 505: you should be able to pipe the dockerfile content into the docker build command by specifying the docker context (last argument) as "-"

tools/ctl/ctl.go
```diff
@@ -462,6 +467,68 @@ var listRuns = &cobra.Command{
...
+			NPM:      npmreg.HTTPRegistry{Client: http.DefaultClient},
+			PyPI:     pypireg.HTTPRegistry{Client: http.DefaultClient},
+		}
+		t := rebuild.Target{
+			Ecosystem: rebuild.Ecosystem(*ecosystem),
+			Package:   *pkg,
+			Version:   *version,
+			Artifact:  *artifact,
+		}
+		s, err := r.InferStrategy(cmd.Context(), t, mux, nil, nil)
```
msuozzo, line 488: what if we support local execution (with inferenceservice) as well as remote execution (with an API handler). that way you could avoid managing the Rebuilder directly. that said, we may want to add the repo storage to InferDeps (tho the default of memfs should be fine in the general case here).

msuozzo (approved):

PR #143, opened 2024-11-06
Incrementally construct RebuildOne verdict

msuozzo:
Not a major change but one consequence of this is that the Strategy and Timings will now be stored in the db entry for this run. That should give us more context to resume/edit/debug runs on failure.

### DIFF
--- a/pkg/rebuild/rebuild/rebuildone.go
+++ b/pkg/rebuild/rebuild/rebuildone.go
@@ -29,28 +29,30 @@ import (
 )
 
 // RebuildOne runs a rebuild for the given package artifact.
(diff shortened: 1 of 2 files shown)


pkg/rebuild/rebuild/rebuildone.go
```diff
@@ -29,119 +29,109 @@ import (
 )
 
 // RebuildOne runs a rebuild for the given package artifact.
```
wbxyz, line 31: The signature only tells you that verdict and toUpload are non-nil, want to mention that they contain valid data even when err != nil?

wbxyz (approved):

PR #156, opened 2024-11-13
Extract packages from network log

msuozzo:


internal/netclassify/classify_test.go
```diff
@@ -0,0 +1,197 @@
...
+		},
+		{
+			name:    "egg file",
+			fname:   "package-1.0.0.egg",
+			wantErr: ErrUnclassified,
+		},
+		{
+			name:  "source distribution tar.gz",
+			fname: "package-1.0.0.tar.gz",
+			want:  "pkg:pypi/package@1.0.0",
```
wbxyz, line 167: optional: Differentiate source vs non-source dependencies

wbxyz (approved):

internal/netclassify/classify_test.go
msuozzo: i think it's valuable in the future but I'm not sure whether it's critical right now. I've added a TODO and if/when this comes back from downstream consumers, we should revisit.

PR #160, opened 2024-11-14
Wait for tetragon to start before building

wbxyz:


pkg/rebuild/rebuild/rebuildremote.go
```diff
@@ -138,14 +198,20 @@ var standardBuildTpl = template.Must(
...
 				touch /workspace/tetragon.jsonl
-				docker run --name=tetragon --detach --pid=host --cgroupns=host --privileged -v=/workspace/tetragon.jsonl:/workspace/tetragon.jsonl -v=/sys/kernel/btf/vmlinux:/var/lib/tetragon/btf quay.io/cilium/tetragon:v1.1.2 /usr/bin/tetragon --export-filename=/workspace/tetragon.jsonl
+				echo '{{.SyscallPolicy}}' > /workspace/tetragon_policy.yaml
+				export TID=$(docker run --name=tetragon --detach --pid=host --cgroupns=host --privileged -v=/workspace/tetragon.jsonl:/workspace/tetragon.jsonl -v=/workspace/tetragon_policy.yaml:/workspace/tetragon_policy.yaml -v=/sys/kernel/btf/vmlinux:/var/lib/tetragon/btf quay.io/cilium/tetragon:v1.1.2 /usr/bin/tetragon --tracing-policy=/workspace/tetragon_policy.yaml --export-filename=/workspace/tetragon.jsonl)
 				{{- end}}
 				cat <<'EOS' | docker buildx build --tag=img -
 				{{.Dockerfile}}
 				EOS
+				{{- if .UseSyscallMonitor}}
+				grep -q "Listening for events..." <(docker logs --follow $TID 2>&1)
```
msuozzo, line 212: this will fail the script when it fails so you can remove the line below.

per offline discussion, you can also output the logs in the case of failure with ` || docker logs $TID && exit 1`

pkg/rebuild/rebuild/rebuildremote.go
```diff
@@ -243,6 +310,10 @@ var proxyBuildTpl = template.Must(
 						docker buildx build --builder proxied --build-context certs=/etc/ssl/certs --secret id=PROXYCERT --load --tag=img -
 					{{.Dockerfile}}
 				EOS
+					{{- if .UseSyscallMonitor}}
```
msuozzo, line 313: let's also move this above the docker build so we capture any activity that happens there

PR #167, opened 2024-11-20
Remove obsolete comment

msuozzo:


tools/ctl/rundex/rundex.go
```diff
@@ -212,7 +212,6 @@ type FirestoreClient struct {
 	Client *firestore.Client
 }
 
-// FirestoreClient is only a Reader for now.
```
wbxyz, line 215: Can you change this to a TODO to add writer capability?

PR #176, opened 2024-11-20
Add a rebuild recovery tool

msuozzo:


tools/eval/main.go
```diff
@@ -0,0 +1,433 @@
...
+	ctx := context.Background()
+
+	// Initialize dependencies
+	client := must(genai.NewClient(ctx, *modelProject, "us-central1"))
+	model := client.GenerativeModel(*baseModel)
+	model = llm.WithSystemPrompt(*model, llm.NPMSystemPrompt)
+	apiURL := must(url.Parse(*apiURI))
+	idclient := must(oauth.AuthorizedUserIDClient(ctx))
+	dex := must(rundex.NewFirestore(ctx, *project))
+	debug := must(rebuild.NewGCSStore(context.WithValue(ctx, rebuild.RunID, *runFlag), *debugStorage))
```
wbxyz, line 237: optional: Maybe worth putting the RunID in the main-level ctx? IIUC that would match the ctx scope and the pertinance of the RunID.

On the other hand, I can see the argument for using WithValue as late as possible though, it minimizes the amount of code that can be dependent on the RunID value (mitigating the downsides of global state).

tools/eval/main.go
```diff
@@ -0,0 +1,433 @@
...
+	cache := rebuild.NewFilesystemAssetStore(osfs.New("/tmp/eval-cache"))
```
wbxyz, line 238: Maybe use the localfiles library to define the cache location?

wbxyz:
The filepath doesn't make it very obvious that this is an LLM recovery tool (tools/eval/main.go). WDYT of (tools/recovery/main.go) or maybe (tools/eval/recovery/main.go)

wbxyz (approved):

tools/eval/main.go
msuozzo: yeah i view it mostly as regrettable and easier to manage if we minimize the breadth of impact the ctx value has. i'd lean towards keeping it here

tools/eval/main.go
msuozzo: nah, that lib is in ctl and this is clear at the callsite imo

PR #180, opened 2024-11-22
Add ctl infer command

wbxyz:


tools/ctl/ctl.go
```diff
@@ -444,6 +445,34 @@ var listRuns = &cobra.Command{
 	},
 }
 
+var infer = &cobra.Command{
+	Use:   "infer --ecosystem <ecosystem> --package <name> --version <version> [--artifact <name>]",
```
msuozzo, line 449: Would it be possible to support both remote and local variants here? Would be nice to be able to dispatch to the hosted version, too

tools/ctl/ctl.go
```diff
@@ -444,6 +445,34 @@ var listRuns = &cobra.Command{
...
+	Run: func(cmd *cobra.Command, args []string) {
+		deps := &inferenceservice.InferDeps{
+			HTTPClient: http.DefaultClient,
+			GitCache:   nil,
+		}
+		req := schema.InferenceRequest{
+			Ecosystem: rebuild.Ecosystem(*ecosystem),
+			Package:   *pkg,
+			Version:   *version,
+			//Artifact:  *artifact,
```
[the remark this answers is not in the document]
wbxyz: It wasn't necessary for Debian to add support here, we can add support whenever it becomes necessary.

tools/ctl/ctl.go
wbxyz: Added

msuozzo (approved):

PR #185, opened 2024-11-25
refactor(debian): inference incrementally constructs strategy

wbxyz:
Construct the debian strategy in-place

Include DSC URL in inference failure error message

### DIFF
--- a/pkg/rebuild/debian/infer.go
+++ b/pkg/rebuild/debian/infer.go
@@ -21,6 +21,7 @@ import (
 	billy "github.com/go-git/go-billy/v5"
 	"github.com/go-git/go-git/v5/storage"
 	"github.com/google/oss-rebuild/pkg/rebuild/rebuild"
+	"github.com/google/oss-rebuild/pkg/registry/debian"
 	debreg "github.com/google/oss-rebuild/pkg/registry/debian"
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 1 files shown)


pkg/rebuild/debian/infer.go
```diff
@@ -79,20 +82,13 @@ func (Rebuilder) InferStrategy(ctx context.Context, t rebuild.Target, mux rebuil
...
-				dependencies = append(dependencies, deps...)
+				p.Requirements = append(p.Requirements, deps...)
 			}
 		}
 	}
-	if (orig.URL == "" || debianTar.URL == "") && (native.URL == "") {
-		return nil, errors.New("Failed to find source files in the .dsc file")
+	if (p.Orig.URL == "" || p.Debian.URL == "") && (p.Native.URL == "") {
+		log.Printf("Partially constructed strategy: %v", p)
+		return nil, fmt.Errorf("failed to find source files in the .dsc file: %s", p.DSC.URL)
```
msuozzo, line 91: errors.Errorf

pkg/rebuild/debian/infer.go
```diff
@@ -79,20 +82,13 @@ func (Rebuilder) InferStrategy(ctx context.Context, t rebuild.Target, mux rebuil
...
 				}
-				dependencies = append(dependencies, deps...)
+				p.Requirements = append(p.Requirements, deps...)
 			}
 		}
 	}
-	if (orig.URL == "" || debianTar.URL == "") && (native.URL == "") {
-		return nil, errors.New("Failed to find source files in the .dsc file")
+	if (p.Orig.URL == "" || p.Debian.URL == "") && (p.Native.URL == "") {
+		log.Printf("Partially constructed strategy: %v", p)
```
msuozzo, line 90: are you wedded to logging this? i think i'd lean away from logging here but I am open to it

msuozzo (approved):

msuozzo:
Can you tweak the PR name to mention something that is being refactored? you could also list out the changes in the commit/PR message so there's at least some context provided in the history.

pkg/rebuild/debian/infer.go
msuozzo: To elaborate, the cost to reconstruct this data seems quite low while the range of cases you're trying to account for is unclear. Could you instead log unknown suffixes?

PR #129, opened 2024-10-18
Upload dockerfile and build-info to remote.

wbxyz:
Fixes #122.

Issue #122, opened 2024-10-11
Build info and Dockerfile not present in metadata GCS bucket when rebuild fails


msuozzo:
The netlog and syscall logs will be present but not the basic metadata of the build.

PR #187, opened 2024-12-02
fix: Always write the BuildInfo, even on failures

wbxyz:
This is probably related to #122 and goes beyond what #129 accomplished.

In #129 we made sure dockerfile and build info are uploaded even when rebuild comparison or attestation fails. In this PR we upload whatever buildinfo is available even if GCB returned with an error.

### DIFF
--- a/pkg/rebuild/rebuild/rebuildremote.go
+++ b/pkg/rebuild/rebuild/rebuildremote.go
@@ -499,9 +499,7 @@ func RebuildRemote(ctx context.Context, input Input, id string, opts RemoteOptio
 	if err != nil {
 		return errors.Wrap(err, "creating build")
 	}
-	if err := doCloudBuild(ctx, opts.GCBClient, build, opts, &bi); err != nil {
-		return errors.Wrap(err, "performing build")
-	}
+	buildErr := errors.Wrap(doCloudBuild(ctx, opts.GCBClient, build, opts, &bi), "performing build")
 	{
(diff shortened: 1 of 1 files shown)


pkg/rebuild/rebuild/rebuildremote.go
```diff
@@ -497,23 +498,21 @@ func RebuildRemote(ctx context.Context, input Input, id string, opts RemoteOptio
...
 		defer lw.Close()
 		rw, err := opts.RemoteMetadataStore.Writer(ctx, Asset{Target: t, Type: BuildInfoAsset})
 		if err != nil {
-			return errors.Wrap(err, "creating remote writer for build info")
+			return stderrs.Join(buildErr, errors.Wrap(err, "creating remote writer for build info"))
 		}
 		defer rw.Close()
 		if err := json.NewEncoder(io.MultiWriter(lw, rw)).Encode(bi); err != nil {
-			return errors.Wrap(err, "marshalling and writing build info")
+			return stderrs.Join(buildErr, errors.Wrap(err, "marshalling and writing build info"))
```
msuozzo, line 514: as discussed offline, errors.Join doesn't seem to add enough value to depart from the existing linear-error pattern we're using in the rebuild package. it might be good to break this out into an issue where we can look for opportunities to add apis that benefit from this structure. one issue brought up was that the top-level error no longer has the stack trace attached since it originates from std errors instead of pkg/errors.

msuozzo (approved):

PR #189, opened 2024-12-02
fix: Move rebuildremote uploads to the debug store

wbxyz:
The uploads were not easy to access before, because they were behind the oblivious uuid. This change makes the rebuildremote debug handling more similar to the smoketest handling.

I don't love the increasing number of asset stores this PR adds, but it seemed necessary in order to store the BuildInfo in a location we can programatically find using the RunID.

I left a few comments suggesting we can combine the debug storage and the local store. My thinking is that we can combine them in a layered asset store that has an internal memfs cache. (draft #191) I'm not wedded to that idea, but it's an option to avoid the bloat of asset stores.

### DIFF
--- a/internal/api/apiservice/rebuild.go
+++ b/internal/api/apiservice/rebuild.go
@@ -119,6 +119,7 @@ func populateArtifact(ctx context.Context, t *rebuild.Target, mux rebuild.Regist
 	return nil
 }
 
+// TODO: LocalMetadataStore and DebugStoreBuilder can be combined into a layered AssetStore.
 type RebuildPackageDeps struct {
 	HTTPClient                 httpx.BasicClient
 	FirestoreClient            *firestore.Client
@@ -131,7 +132,8 @@ type RebuildPackageDeps struct {
 	BuildDefRepo               rebuild.Location
 	AttestationStore           rebuild.AssetStore
 	LocalMetadataStore         rebuild.AssetStore
-	RemoteMetadataStoreBuilder func(ctx context.Context, id string) (rebuild.LocatableAssetStore, error)
+	DebugStoreBuilder          func(ctx context.Context) (rebuild.AssetStore, error)
+	RemoteMetadataStoreBuilder func(ctx context.Context, uuid string) (rebuild.LocatableAssetStore, error)
 	OverwriteAttestations      bool
 	InferStub                  api.StubT[schema.InferenceRequest, schema.StrategyOneOf]
 }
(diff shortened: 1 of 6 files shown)


internal/api/apiservice/rebuild.go
```diff
@@ -200,6 +202,10 @@ func getStrategy(ctx context.Context, deps *RebuildPackageDeps, t rebuild.Target
 }
 
 func buildAndAttest(ctx context.Context, deps *RebuildPackageDeps, mux rebuild.RegistryMux, a verifier.Attestor, t rebuild.Target, strategy rebuild.Strategy, entry *repoEntry, useProxy bool, useSyscallMonitor bool) (err error) {
+	debugStore, err := deps.DebugStoreBuilder(ctx)
+	if err != nil {
+		return errors.Wrap(err, "creating debug store")
+	}
 	id := uuid.New().String()
```
msuozzo, line 209: can be a separate PR but I think this ID should be recorded in Firestore to avoid needing to access GCS to know how to read the logs. Alternatively, as mentioned in the TUI PR, I think it might be reasonable to copy the logs, no matter how large, to the debug store after execution.

internal/api/apiservice/rebuild.go
```diff
@@ -120,18 +120,20 @@ func populateArtifact(ctx context.Context, t *rebuild.Target, mux rebuild.Regist
...
+	FirestoreClient     *firestore.Client
+	Signer              *dsse.EnvelopeSigner
+	GCBClient           gcb.Client
+	BuildProject        string
+	BuildServiceAccount string
+	UtilPrebuildBucket  string
+	BuildLogsBucket     string
+	BuildDefRepo        rebuild.Location
+	AttestationStore    rebuild.AssetStore
+	// TODO: The local store and the debug store can be combined into a layered (cached) asset store.
```
msuozzo, line 133: nit: want to shorten this TODO and add it at the end of the next line so this PR doesn't impact formatting so much?

msuozzo (approved):

internal/api/apiservice/rebuild.go
wbxyz: Already mailed in #194. Just FYI this uuid is not the uuid needed for the logs. The logs are written by the GCB service using the GCB generated BuildID, whereas this uuid is our oblivious uuid. Related though, because #194 adds both ids to the firestore record.

internal/api/apiservice/rebuild.go
wbxyz: Moved to the top, couldn't really shorten it much but it doesn't affect formatting anymore.

PR #190, opened 2024-12-03
feat: Tui reads from gcb logs if available

wbxyz:


tools/ctl/ide/ui.go
```diff
@@ -207,6 +218,65 @@ func (e *explorer) showDetails(ctx context.Context, example rundex.Rebuild) {
 	showModal(e.app, e.container, details, modalOpts{Margin: 10})
 }
 
+func downloadLogs(ctx context.Context, localAssets rebuild.AssetStore, example rundex.Rebuild) error {
```
msuozzo, line 221: ugh. this makes me want to copy the logs to the new debug bucket. I get that it's potentially a waste of resources but this access pattern just doesn't bring me joy (especially given the additional CLI flag). I'm fine with it for now but it might be nice to copy to the debug storage after the GCB run to eliminate this complexity. Maybe add a TODO?

tools/ctl/ide/ui.go
```diff
@@ -207,6 +218,65 @@ func (e *explorer) showDetails(ctx context.Context, example rundex.Rebuild) {
...
+		}
+	}
+	if info.BuildID == "" {
+		return errors.New("BuildID is empty, cannot read gcb logs")
+	}
+	client, err := gcs.NewClient(ctx)
+	if err != nil {
+		return errors.Wrap(err, "creating gcs client")
+	}
+	obj := client.Bucket(logsBucket).Object(fmt.Sprintf("log-%s.txt", info.BuildID))
```
msuozzo, line 260: honestly, the log name helper might fit better in the GCB internal library. this isn't obvious and probably deserves some central interface to ensure we get it right (where we could also document/link to the other step-wise log files).

tools/ctl/ide/ui.go
```diff
@@ -207,6 +218,65 @@ func (e *explorer) showDetails(ctx context.Context, example rundex.Rebuild) {
...
+	r, err := obj.NewReader(ctx)
+	if err != nil {
+		return errors.Wrap(err, "reading gcb logs")
+	}
+	defer r.Close()
+	w, err := localAssets.Writer(ctx, logsAsset)
+	if err != nil {
+		return errors.Wrap(err, "making localAsset writer")
+	}
+	defer w.Close()
```
msuozzo, line 270: this is safe to call twice, correct?

msuozzo (approved):

tools/ctl/ide/ui.go
wbxyz: Added a TODO to maybe copy the logs to debug bucket.

tools/ctl/ide/ui.go
wbxyz: Moved to internal lib, added comment about per-step logs.

tools/ctl/ide/ui.go
wbxyz: Not exactly. For os.File writers, the second call with return an error. However in this case that should be the deferred call in which case an error does not bother us.

https://pkg.go.dev/os#File.Close

> Close will return an error if it has already been called.

We might want to improve our deferred close calls at some point in the future, but for now this is fine.

PR #191, opened 2024-12-03
Used CachedAssetStore for metadata.

wbxyz:


pkg/rebuild/rebuild/storage.go
```diff
@@ -246,3 +247,68 @@ var _ LocatableAssetStore = &FilesystemAssetStore{}
 func NewFilesystemAssetStore(fs billy.Filesystem) *FilesystemAssetStore {
 	return &FilesystemAssetStore{fs: fs}
 }
+
+type CachedAssetStore struct {
+	fs     billy.Filesystem
+	local  *FilesystemAssetStore
```
msuozzo, line 253: what if we made this generic to the underlying asset store? one store would be the cache, one would be the last-level. you could even reuse the basic cache abstraction we already have (although it might not simplify things)

PR #193, opened 2024-12-03
feat: Include attestation runs in TUI by default

wbxyz:


tools/ctl/ide/ui.go
```diff
@@ -416,7 +424,7 @@ func (e *explorer) LoadTree() error {
 	}
 	byBench := make(map[string][]string)
 	for _, run := range runs {
-		if run.Type == benchmark.AttestMode {
+		if !e.settings.includeAttest && run.Type == benchmark.AttestMode {
```
msuozzo, line 427: do we even need this constraint? would an alternative be some indication in the UI?

msuozzo (commented):
I could see just removing the constraint for now and then potentially re-introducing some visible indicator (possibly along with size / benchmark) to the UI

tools/ctl/ide/ui.go
wbxyz: I think eventually we will want some setting/filters on the explorer with a way to adjust them in the UI (I'm thinking like filter based on error message, or grouping based on log regex) but I can leave the settings construct out of this PR and just add an indication. The indication is useful regardless.

tools/ctl/ide/ui.go
```diff
@@ -368,7 +369,15 @@ func (e *explorer) makeVerdictGroupNode(vg *rundex.VerdictGroup, percent float32
 }
 
 func (e *explorer) makeRunNode(runid string) *tview.TreeNode {
-	node := tview.NewTreeNode(runid).SetColor(tcell.ColorGreen).SetSelectable(true)
+	var title string
+	if run, ok := e.runs[runid]; ok && run.Type == benchmark.AttestMode {
+		title = fmt.Sprintf("(a)%s", runid)
```
msuozzo, line 374: ugh I don't love the single-letters, especially for new users, but I guess it's fine for now...

would mainly want to keep them the same length and I can't think of two descriptors that are consistent with our current terminology and maintain that property.

tools/ctl/ide/ui.go
```diff
@@ -410,16 +419,16 @@ func (e *explorer) makeRunGroupNode(benchName string, runs []string) *tview.Tree
 func (e *explorer) LoadTree() error {
 	e.root.ClearChildren()
 	log.Printf("Fetching runs...")
+	var err error
```
msuozzo, line 422: is this still necessary?

msuozzo (approved):

tools/ctl/ide/ui.go
wbxyz: Discussed offline, we decided to move the descriptors to the end of the runid and use longer, more descriptive terms.

tools/ctl/ide/ui.go
wbxyz: Nope, removed.

PR #194, opened 2024-12-03
Include BuildID and oblivious ID in firestore

wbxyz:


msuozzo:
Can you sequence this change before the cache PR? I think that may need a bit more work.

msuozzo (approved):

PR #195, opened 2024-12-03
refactor: Rename SmoketestAttempt -> RebuildAttempt

wbxyz:
We use this format only to store data in firestore, but we use it for
both smoketest and attestation runs.

### DIFF
--- a/internal/api/apiservice/rebuild.go
+++ b/internal/api/apiservice/rebuild.go
@@ -319,7 +319,7 @@ func RebuildPackage(ctx context.Context, req schema.RebuildPackageRequest, deps 
 			dockerfile = string(b)
 		}
 	}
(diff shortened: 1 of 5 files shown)


cmd/api/main.go
```diff
@@ -136,9 +137,24 @@ func RebuildPackageInit(ctx context.Context) (*apiservice.RebuildPackageDeps, er
 	if err != nil {
 		return nil, errors.Wrap(err, "creating attestation uploader")
 	}
-	d.LocalMetadataStore = rebuild.NewFilesystemAssetStore(memfs.New())
-	d.RemoteMetadataStoreBuilder = func(ctx context.Context, id string) (rebuild.LocatableAssetStore, error) {
-		return rebuild.NewGCSStore(context.WithValue(ctx, rebuild.RunID, id), "gs://"+*metadataBucket)
+	if *debugStorage == "" {
```
msuozzo, line 140: LGTM but would you mind removing the diff and submitting this change first? Feels like a simpler change overall.

msuozzo (approved):

PR #197, opened 2024-12-04
Add workflow strategy

msuozzo:
This new strategy is based off of the melange syntax (which itself inherits from GitHub Actions and others) and allows us to expose individual commands as opposed to our current monolithic strategies.

### DIFF
--- a/pkg/rebuild/rebuild/workflowstrategy.go
+++ b/pkg/rebuild/rebuild/workflowstrategy.go
@@ -1,0 +1,169 @@
+// Copyright 2024 The OSS Rebuild Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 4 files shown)


pkg/rebuild/rebuild/muddlestrategy_test.go
```diff
@@ -0,0 +1,413 @@
...
+			if cmd == nil {
+				t.Fatal("git-checkout command not found")
+			}
+
+			data := struct {
+				With     map[string]string
+				Target   Target
+				BuildEnv BuildEnv
+				Location Location
+			}{
```
wbxyz, line 264: Should this use the same typedef as the system under test? I'm leaning yes because we would want this test to break if the builtin is depending on a field that's not present in the real data struct.

wbxyz: Actually, why not use generateForStep in this test?

wbxyz (approved):

pkg/rebuild/rebuild/muddlestrategy_test.go
msuozzo: good point. this was added a few refactors before that existed.

PR #201, opened 2024-12-06
Revamp CreateRun to use storage model

msuozzo:


pkg/rebuild/schema/schema.go
```diff
@@ -250,3 +246,12 @@ type RebuildAttempt struct {
 	ObliviousID     string          `firestore:"oblivious_id,omitempty"`
 	Created         int64           `firestore:"created,omitempty"`
 }
+
+// Run stores metadata on an execution grouping.
+type Run struct {
+	ID      string `firestore:"id,omitempty"`
+	Name    string `firestore:"benchmark_name,omitempty"`
```
wbxyz, line 253: Name is a confusing field on Run, it sounds like it would be a unique name for this specific run, and might be confused with Run.ID. WDYT about keeping the existing field names (BenchmarkName) or maybe shortening them slightly (BenchName)?

Same thing with "Hash"

wbxyz (approved):

pkg/rebuild/schema/schema.go
msuozzo: yep agreed. I'd probably stick the benchmark prefix.

PR #203, opened 2024-12-06
Add timewarp support for PureWheelBuild

msuozzo:


pkg/rebuild/pypi/strategy.go
```diff
@@ -34,13 +36,23 @@ func (b *PureWheelBuild) GenerateFor(t rebuild.Target, be rebuild.BuildEnv) (reb
...
+		*PureWheelBuild
+		BuildEnv *rebuild.BuildEnv
+	}{
+		PureWheelBuild: b,
+		BuildEnv:       &be,
+	}
 	deps, err := rebuild.PopulateTemplate(`
 /usr/bin/python3 -m venv /deps
+{{if not .RegistryTime.IsZero -}}
+export PIP_INDEX_URL={{.BuildEnv.TimewarpURL "pypi" .RegistryTime}}
```
wbxyz, line 49: Any reason to prefer this over setting the config on the CLI?

https://stackoverflow.com/questions/60814982/how-to-setup-pip-to-download-from-mirror-repository-by-default

IIUC `pip config set` is closer to what we do for NPM.

wbxyz (approved):

pkg/rebuild/pypi/strategy.go
msuozzo: Yeah fair question. Mainly, the juggled versions of npm/pip create more sensitivity to ordering than env vars AND we no longer need to worry about rc files getting accidentally packaged. I'd also like to move the NPM strategy to use the env var but haven't gotten the chance (the recovery logic I just added uses it already).

PR #211, opened 2024-12-09
perform domain suffix matching rather than plain string suffix one

mrahs:
This is to avoid matching unintended domains. E.g. notgoogle.com must not match google.com, but is.google.com must.

### DIFF
--- a/pkg/proxy/policy/policy_test.go
+++ b/pkg/proxy/policy/policy_test.go
@@ -89,7 +89,7 @@ func TestApplyOnURLMatchRule(t *testing.T) {
 			wantResp: http.StatusForbidden,
 		},
 		{
-			name: "host policy rule matches url host suffix",
+			name: "host policy rule matches full url host by suffix",
 			policy: Policy{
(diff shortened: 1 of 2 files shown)


mrahs:
@msuozzo can you approve the workflow and merge?

pkg/proxy/policy/policy_test.go
```diff
@@ -103,6 +103,21 @@ func TestApplyOnURLMatchRule(t *testing.T) {
...
+				AnyOf: []Rule{
+					URLMatchRule{
+						Host:      "host.com",
+						HostMatch: SuffixMatch,
+						Path:      "/path",
+						PathMatch: PrefixMatch,
+					},
+				},
+			},
+			url:      "https://sub.host.com/path/with/prefix",
```
msuozzo, line 118: I think my comment may have gotten eaten first time, but would you mind adding another test case for `nothost.com`?

mrahs: Done

PR #212, opened 2024-12-10
fix: Don't expect binary-only release versions in output

wbxyz:


pkg/rebuild/debian/strategy.go
```diff
@@ -63,11 +70,25 @@ apt install -y {{join " " .Requirements}}
 	if err != nil {
 		return rebuild.Instructions{}, err
 	}
+	// If the target is a binary-only release (version ends with something like +b1) we need to add an additonal rename.
+	var expectedArtifact string
```
msuozzo, line 74: nit: `expected`?

msuozzo (approved):

PR #213, opened 2024-12-11
Add exponential backoff in benchmark executors

wbxyz:


tools/benchmark/run.go
```diff
@@ -59,6 +59,43 @@ func makeHTTPRequest(ctx context.Context, u *url.URL, msg schema.Message) (*http
 	return req, nil
 }
 
+type Limiter struct {
```
msuozzo, line 62: this looks kinda like https://pkg.go.dev/golang.org/x/time/rate

want to try to use that instead? the previous impl was mainly chosen for simplicity but if we're not getting results from it, using something more substantial may be the best option.

That said, what's your theory as to why 429s are occurring? is it overloading at the start of execution? is it steady state excesses? I suspect it's mostly the latter in which case you might try a token bucket construct. if it's the former, you'd probably want to pair that with a slower initial token bucket fill.

msuozzo: this is what i'm thinking: [push-usosnltlpvlv](https://github.com/google/oss-rebuild/compare/main...msuozzo:push-usosnltlpvlv)

wbxyz: I think the impl proposed in [push-usosnltlpvlv](https://github.com/google/oss-rebuild/compare/main...msuozzo:push-usosnltlpvlv) did not update the period for any concurrent requests when Backoff() was called, which would result in the limiter over or under shooting the "target" rate by an amount exponential with the concurrency of the run.

I updated this implementation to be simpler and I took some of the constants you proposed in that branch (backoff by 33%, success speedup by 10%). In the future we could potentially refactor the benchmark executor as something that looks like a local implementation of a task queue but I think this change is incrementally better for now.

tools/benchmark/run/run.go
```diff
@@ -137,7 +137,24 @@ func (w *attestWorker) ProcessOne(ctx context.Context, p benchmark.Package, out
 			UseSyscallMonitor: w.useSyscallMonitor,
 			UseNetworkProxy:   w.useNetworkProxy,
 		}
-		verdict, err := w.execService.RebuildPackage(ctx, req)
+		var verdict *schema.Verdict
+		var err error
+		for range 3 {
```
msuozzo, line 144: can you add a TODO here to use the internal/taskqueue interface for a local queue that could abstract these params out of this code?

PR #214, opened 2024-12-11
fix: Parse attestation verdicts in run-bench

wbxyz:


tools/benchmark/run.go
```diff
@@ -145,18 +190,12 @@ func (w *attestWorker) ProcessOne(ctx context.Context, p Package, out chan schem
...
-			// support that here.
-			verdict = schema.Verdict{
-				Target: rebuild.Target{
-					Ecosystem: rebuild.Ecosystem(p.Ecosystem),
-					Package:   p.Name,
-					Version:   v,
-				},
-				Message: "",
+			var v schema.Verdict
+			if err := json.NewDecoder(resp.Body).Decode(&v); err != nil {
```
msuozzo, line 194: can we use stubs here?

wbxyz: Yeah, but maybe in a followup PR?

wbxyz: Added TODO comments

msuozzo: eh i mean the purpose of this change is to start using the return value from the attestation verdict so it seems like the right time to move to stubs.

msuozzo: ok i just went ahead and did the conversion in #219. in the course of doing so, found and fixed a bug in #220

PR #215, opened 2024-12-11
fix: Check for nil build results

wbxyz:
Only report errors about nil build results if the build was otherwise
error-free. This is because the build error is probably more descriptive
than build results, but we still want to emit an error when the results
are missing to avoid attesting without build image info.

### DIFF
--- a/pkg/rebuild/rebuild/rebuildremote.go
+++ b/pkg/rebuild/rebuild/rebuildremote.go
@@ -441,10 +441,15 @@ func doCloudBuild(ctx context.Context, client gcb.Client, build *cloudbuild.Buil
 	bi.BuildID = build.Id
 	bi.Steps = build.Steps
 	bi.BuildImages = make(map[string]string)
-	for i, s := range bi.Steps {
-		bi.BuildImages[s.Name] = build.Results.BuildStepImages[i]
+	buildErr := gcb.ToError(build)
+	// Don't try to read BuildStepImages if the build failed.
+	// It's possible we're missing some valid BuildStepImages this way, but not super important.
+	if buildErr == nil {
+		for i, s := range bi.Steps {
+			bi.BuildImages[s.Name] = build.Results.BuildStepImages[i]
+		}
 	}
-	return gcb.ToError(build)
+	return buildErr
 }
 
 func makeDockerfile(input Input, opts RemoteOptions) (string, error) {


pkg/rebuild/rebuild/rebuildremote.go
```diff
@@ -441,10 +441,15 @@ func doCloudBuild(ctx context.Context, client gcb.Client, build *cloudbuild.Buil
 	bi.BuildID = build.Id
 	bi.Steps = build.Steps
 	bi.BuildImages = make(map[string]string)
+	buildErr := gcb.ToError(build)
+	// Only return errors about build.Results if the build was otherwise error-free.
+	if buildErr == nil && (build.Results == nil || build.Results.BuildStepImages == nil || len(build.Results.BuildStepImages) != len(bi.Steps)) {
```
msuozzo, line 446: I think this is a bug in ToError. Why is it not emitting an error in this case?

wbxyz: It wasn't getting to gcb.ToError because it was panic-ing on a nil reference deref

msuozzo: Ah wait I didn't see that ToError was called after the loop in the current code.

Yeah I think just lazily populating the field with the data available is fine. I don't even think additional checks are necessary since len(BuildStepImages) == len(Steps) is a documented behavior of the API in the case of a successful build. As such, the attestation code could assume it's populated.

wbxyz: How's that?

pkg/rebuild/rebuild/rebuildremote.go
```diff
@@ -441,10 +441,13 @@ func doCloudBuild(ctx context.Context, client gcb.Client, build *cloudbuild.Buil
 	bi.BuildID = build.Id
 	bi.Steps = build.Steps
 	bi.BuildImages = make(map[string]string)
-	for i, s := range bi.Steps {
-		bi.BuildImages[s.Name] = build.Results.BuildStepImages[i]
+	buildErr := gcb.ToError(build)
+	if buildErr == nil {
+		for i, s := range bi.Steps {
```
msuozzo, line 446: do you want to just add a comment here explaining which conditions cause BuildStepImages to be empty?

msuozzo (approved):

PR #218, opened 2024-12-11
Add pattern flag to get-results for regex matching

wbxyz:
Most of the changes for this PR were accidentally included in #236 (e7c45d9b02b2bd142e8c01ef90eb634111db4fc6)


tools/ctl/rundex/rundex.go
```diff
@@ -245,7 +245,7 @@ func filterRebuilds(all <-chan Rebuild, req *FetchRebuildRequest) map[string]Reb
 	}
 	if req.Opts.Filter != "" {
```
msuozzo, line 246: changing --filter to --matches-prefix at the cli could maintain the current behavior while we could add --matches for Contains (or, potentially, a full regex).

wbxyz: Changed --filter to --prefix and the new behavior is --pattern.

msuozzo (approved):

PR #219, opened 2024-12-12
Convert benchmark run to use pipe

msuozzo:
This fixes a bug where Process returned (and closed the "out" channel) before
it was finished populating the channel. The pipe library enables us to buffer
the verdicts and count them to ensure we provide the anticipated number to the
output channel.


msuozzo:
The previous impl actually is sound so long as ProcessOne completes its output by the time it returns. Abandoning

PR #220, opened 2024-12-12
Convert benchmark run to use stubs

msuozzo:


tools/benchmark/run.go
```diff
@@ -236,12 +169,11 @@ func defaultLimiters() map[string]<-chan time.Time {
 }
 
 func isCloudRun(u *url.URL) bool {
-	return strings.HasSuffix(u.Host, ".run.app")
+	return u != nil && strings.HasSuffix(u.Host, ".run.app")
 }
 
 type RunBenchOpts struct {
-	Mode BenchmarkMode
-	// RunID is the ID for this run. Leave blank for one to be generated.
```
wbxyz, line 244: Want to document the blank runid behavior somewhere outside the opts struct? Maybe on the function signature?

wbxyz (approved):

tools/benchmark/run.go
msuozzo: eh it turns out it's an error so this was actively wrong :/

the error should be self-explanatory, too, so I think it's tolerable

PR #222, opened 2024-12-16
fix: Add artifact name to firestore path

wbxyz:


internal/api/apiservice/rebuild.go
```diff
@@ -337,7 +337,7 @@ func RebuildPackage(ctx context.Context, req schema.RebuildPackageRequest, deps
 			log.Println("Failed to load build info:", err)
 		}
 	}
-	_, err = deps.FirestoreClient.Collection("ecosystem").Doc(string(v.Target.Ecosystem)).Collection("packages").Doc(sanitize(v.Target.Package)).Collection("versions").Doc(v.Target.Version).Collection("attempts").Doc(req.ID).Set(ctx, schema.RebuildAttempt{
+	_, err = deps.FirestoreClient.Collection("ecosystem").Doc(string(v.Target.Ecosystem)).Collection("packages").Doc(sanitize(v.Target.Package)).Collection("versions").Doc(v.Target.Version).Collection("artifacts").Doc(v.Target.Artifact).Collection("attempts").Doc(req.ID).Set(ctx, schema.RebuildAttempt{
```
msuozzo, line 340: this would then differ from the smoke test path, correct? do we want to keep them in sync?

msuozzo (approved):
We may need to do some work to migrate old entries but this seems necessary for correctness with multiple artifacts per version.

PR #227, opened 2024-12-17
Add artifact support to the inference service

wbxyz:


internal/api/apiservice/rebuild.go
```diff
@@ -154,6 +154,7 @@ func getStrategy(ctx context.Context, deps *RebuildPackageDeps, t rebuild.Target
 		Ecosystem: t.Ecosystem,
 		Package:   t.Package,
 		Version:   t.Version,
+		Artifact:  t.Artifact,
```
msuozzo, line 157: is it possible to validate this in a test? maybe with the debian ecosystem support?

msuozzo (approved):

internal/api/apiservice/rebuild.go
wbxyz: Not yet. The current Debian strategy is artifact independent, but in a followup with the buildinfo strategy I can add some test. Added a todo in the meantime.

PR #229, opened 2024-12-19
Migrate the terraform config to use Artifact Registry

di:
Avoids https://github.com/hashicorp/terraform-provider-google/issues/17605.

### DIFF
--- a/deployments/main.tf
+++ b/deployments/main.tf
@@ -191,39 +191,46 @@ resource "google_storage_notification" "attestation-notification" {
 
 ## Container resources
 
-resource "google_container_registry" "registry" {
-  project  = var.project
-  location = "US"
+resource "google_artifact_registry_repository" "registry" {
+  provider = google-beta
+  location = "us-central1"
+  repository_id = "oss-rebuild-repo"
+  format  = "DOCKER"
 }
 
-data "google_container_registry_image" "gateway" {
-  name = "gateway"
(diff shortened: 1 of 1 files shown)


msuozzo (approved):

deployments/main.tf
```diff
@@ -191,39 +191,46 @@ resource "google_storage_notification" "attestation-notification" {
 
 ## Container resources
 
-resource "google_container_registry" "registry" {
-  project  = var.project
-  location = "US"
+resource "google_artifact_registry_repository" "registry" {
+  provider = google-beta
+  location = "us-central1"
+  repository_id = "oss-rebuild-repo"
```
di, line 197: This should probably not be hard-coded.

di: How about...
```suggestion
  repository_id = "oss-rebuild-service-images"
```
?

msuozzo: ah what i was saying that including "oss-rebuild" as a prefix would make the URI: "pkg.dev/oss-rebuild/oss-rebuild...". So something like "service-images" or "services" would be preferable.

di: Ah, makes sense.
```suggestion
  repository_id = "service-images"
```

di:
This should be good to merge!

msuozzo:
Thanks, Dustin!

PR #230, opened 2024-12-19
Change format flag to default empty

wbxyz:
Also, update the flag docstrings to reflect that the values are
command-specific.

### DIFF
--- a/tools/ctl/ctl.go
+++ b/tools/ctl/ctl.go
@@ -130,7 +130,7 @@ var tui = &cobra.Command{
 }
 
 var getResults = &cobra.Command{
-	Use:   "get-results -project <ID> -run <ID> [-bench <benchmark.json>] [-filter <verdict>] [-sample N]",
+	Use:   "get-results -project <ID> -run <ID> [-bench <benchmark.json>] [-filter <verdict>] [-sample N] [-format=summary|bench]",
 	Short: "Analyze rebuild results",
 	Args:  cobra.NoArgs,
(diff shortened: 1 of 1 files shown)


tools/ctl/ctl.go
```diff
@@ -157,6 +157,8 @@ var getResults = &cobra.Command{
 			return
 		}
 		switch *format {
+		case "":
```
msuozzo, line 160: you can use a comma instead: `case "", "summary"`

tools/ctl/ctl.go
```diff
@@ -297,14 +299,8 @@ var runBenchmark = &cobra.Command{
...
 		// TODO: Maybe add more format options, or include more data in the csv?
-		case "csv":
-			w := csv.NewWriter(cmd.OutOrStdout())
-			defer w.Flush()
-			for _, v := range verdicts {
-				if err := w.Write([]string{fmt.Sprintf("%v", v.Target), v.Message}); err != nil {
-					log.Fatal(errors.Wrap(err, "writing CSV"))
-				}
-			}
+		case "":
```
msuozzo, line 302: and here

msuozzo (approved):

tools/ctl/ctl.go
wbxyz: Done

tools/ctl/ctl.go
wbxyz: Done

PR #232, opened 2024-12-19
Change debian inference to use Debrebuild

wbxyz:
This needs more work around deciding when to use the Debrebuild strategy type over the existing strategy.

### DIFF
--- a/pkg/rebuild/debian/infer.go
+++ b/pkg/rebuild/debian/infer.go
@@ -5,6 +5,7 @@ package debian
 
 import (
 	"context"
+	"fmt"
 	"regexp"
 	"strings"
 
@@ -12,7 +13,6 @@ import (
 	"github.com/go-git/go-git/v5/storage"
 	"github.com/google/oss-rebuild/pkg/rebuild/rebuild"
 	"github.com/google/oss-rebuild/pkg/registry/debian"
-	debianreg "github.com/google/oss-rebuild/pkg/registry/debian"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/rebuild/debian/infer.go
```diff
@@ -32,8 +32,9 @@ func (Rebuilder) CloneRepo(_ context.Context, _ rebuild.Target, _ string, _ bill
 var origRegex = regexp.MustCompile(`\.orig\.tar\.(gz|xz|bz2)$`)
 var debianRegex = regexp.MustCompile(`\.(debian\.tar|diff)\.(gz|xz|bz2)$`)
 var nativeRegex = regexp.MustCompile(`\.tar\.(gz|xz|bz2)$`)
+var versionExtractRegex = regexp.MustCompile(`^(?P<name>[^_]+)_(?P<nonbinary_version>[^_+]+)(?P<binary_version>\+.*)?_(?P<arch>[^_]+)\.deb$`)
```
msuozzo, line 35: just versionRegex

pkg/rebuild/debian/strategy.go
```diff
@@ -110,4 +150,16 @@ var toolkit = []*flow.Tool{
...
 		}},
 	},
+	{
+		Name: "debian/build/debrebuild",
+		Steps: []flow.Step{{
+			Runs: textwrap.Dedent(`
+				debrebuild --buildresult=./out --builder=mmdebstrap {{ .With.buildinfo }}
+				{{- $expected := regexReplace .With.targetPath "\\+b[0-9]+(_[^_]+\\.deb)$" "$1"}}
+				{{- if ne $expected .With.targetPath }}
+				mv /src/{{$expected}} /src/{{.With.targetPath}}
```
msuozzo, line 160: nice

pkg/rebuild/debian/strategy_test.go
```diff
@@ -141,6 +141,72 @@ mv /src/pkg_1.0-1_amd64.deb /src/pkg_1.0-1+b1_amd64.deb`,
...
+			env: rebuild.BuildEnv{},
+			want: rebuild.Instructions{
+				Source:     `wget https://buildinfos.debian.net/buildinfo-pool/a/acl/acl_2.3.2-2_amd64.buildinfo`,
+				Build:      `debrebuild --buildresult=./out --builder=mmdebstrap acl_2.3.2-2_amd64.buildinfo`,
+				OutputPath: "acl_2.3.1-3_amd64.deb",
+				SystemDeps: []string{"wget", "devscripts", "apt-utils", "mmdebstrap"},
+			},
+		},
+		{
+			name: "binaryRelease",
```
msuozzo, line 175: nit: name can have spaces

msuozzo (approved):

pkg/rebuild/debian/strategy.go
wbxyz: Stolen from the debuild above, wdyt of moving the conditional rename step to a dedicated Tool?

msuozzo: yeah sounds good

pkg/rebuild/debian/strategy_test.go
wbxyz: The standard practice in this and other test files seems to be PascalCase, so I'll update this to match.

PR #233, opened 2024-12-19
Add tui flag to configure build definition dir

wbxyz:
This makes it easy to debug rebuilds locally, then commit upstream the
manual strategy changes that were generated.

### DIFF
--- a/tools/ctl/ide/ui.go
+++ b/tools/ctl/ide/ui.go
@@ -111,9 +111,10 @@ type explorer struct {
 	firestore     rundex.Reader
 	firestoreOpts rundex.FetchRebuildOpts
 	runs          map[string]rundex.Run
+	buildDefs     rebuild.LocatableAssetStore
 }
 
-func newExplorer(ctx context.Context, app *tview.Application, firestore rundex.Reader, firestoreOpts rundex.FetchRebuildOpts, rb *Rebuilder) *explorer {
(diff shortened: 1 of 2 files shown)


tools/ctl/ide/ui.go
```diff
@@ -305,9 +308,19 @@ func (e *explorer) showLogs(ctx context.Context, example rundex.Rebuild) {
 }
 
 func (e *explorer) editAndRun(ctx context.Context, example rundex.Rebuild) error {
-	localAssets, err := localfiles.AssetStore(example.RunID)
-	if err != nil {
-		return errors.Wrap(err, "failed to create local asset store")
+	var err error
+	var localAssets *rebuild.FilesystemAssetStore
+	if e.buildDefDir != "" {
+		fs, err := osfs.New("/").Chroot(e.buildDefDir)
```
msuozzo, line 314: not a fan of osfs this deep in an callstack. if we can, injecting / constructing this further up would be preferable. makes testing easier and parameterizes fs access as much as possible.

msuozzo (approved):

tools/ctl/ide/ui.go
wbxyz: Instead of plumbing the buildDefDir all the way through, I changed to plumbing an buildDef LocatableAssetStore. This is one level higher than parameterizing fs, and I think it's closer to the level of abstraction the TUI actually cares about.

PR #237, opened 2024-12-20
fix: Add support for diffing attestation artifacts

wbxyz:


tools/ctl/ide/ui.go
```diff
@@ -159,26 +172,67 @@ func diffArtifacts(ctx context.Context, example rundex.Rebuild) {
...
 			return
 		}
-		log.Printf("downloaded rebuild and upstream:\n\t%s\n\t%s", rba, usa)
+		metadata, _ := rebuild.NewGCSStore(context.WithValue(ctx, rebuild.RunID, bi.ID), "gs://rebuild-metadata")
+		// TODO: Clean up these artifacts.
+		rebuildAsset := rebuild.RebuildAsset.For(t)
+		upstreamAsset := rebuild.DebugUpstreamAsset.For(t)
+		rba = localAssets.URL(rebuildAsset).Path
+		usa = localAssets.URL(upstreamAsset).Path
+		if _, err := os.Stat(rba); errors.Is(err, os.ErrNotExist) {
```
msuozzo, line 214: this breaks the abstraction layer with localAssets, right? would it be possible to try to read from the localAssets and, on failure, do a copy?

wbxyz: Done.

tools/ctl/ide/ui.go
```diff
@@ -159,26 +172,67 @@ func diffArtifacts(ctx context.Context, example rundex.Rebuild) {
...
+		r, err := debugAssets.Reader(ctx, rebuild.Asset{Target: t, Type: rebuild.BuildInfoAsset})
+		if err != nil {
+			log.Println(errors.Wrap(err, "reading build info"))
 			return
 		}
-	}
-	if _, err := os.Stat(usa); errors.Is(err, os.ErrNotExist) {
-		if err := rebuild.AssetCopy(ctx, localAssets, debugAssets, upstreamAsset); err != nil {
-			log.Println(errors.Wrap(err, "failed to copy upstream asset"))
+		bi := rebuild.BuildInfo{}
```
msuozzo, line 203: var bi rebuild.BuildInfo

tools/ctl/ide/ui.go
```diff
@@ -159,26 +172,67 @@ func diffArtifacts(ctx context.Context, example rundex.Rebuild) {
...
+		}
+		if _, err := localAssets.Reader(ctx, upstreamAsset); errors.Is(err, os.ErrNotExist) {
+			if err := rebuild.AssetCopy(ctx, localAssets, debugAssets, upstreamAsset); err != nil {
+				log.Println(errors.Wrap(err, "failed to copy upstream asset"))
+				return
+			}
+			log.Printf("downloaded upstream: %s", usa)
+		}
+	} else {
+		r, err := debugAssets.Reader(ctx, rebuild.Asset{Target: t, Type: rebuild.BuildInfoAsset})
```
msuozzo, line 198: rebuild.BuildInfoAsset.For(t)

tools/ctl/ide/ui.go
```diff
@@ -159,26 +172,67 @@ func diffArtifacts(ctx context.Context, example rundex.Rebuild) {
...
-	// TODO: Check if these are already downloaded.
-	rebuildAsset := rebuild.DebugRebuildAsset.For(t)
-	upstreamAsset := rebuild.DebugUpstreamAsset.For(t)
-	rba := localAssets.URL(rebuildAsset).Path
-	usa := localAssets.URL(upstreamAsset).Path
-	if _, err := os.Stat(rba); errors.Is(err, os.ErrNotExist) {
-		if err := rebuild.AssetCopy(ctx, localAssets, debugAssets, rebuildAsset); err != nil {
-			log.Println(errors.Wrap(err, "failed to copy rebuild asset"))
+	var rba, usa string
+	wasSmoketest := example.ObliviousID == ""
```
msuozzo, line 176: can we add this as a method on rundex.Rebuild? I feel like there's a better way to do this and keeping the logic near the model feels like a cleaner place to play with it.

tools/ctl/ide/ui.go
```diff
@@ -159,26 +172,67 @@ func diffArtifacts(ctx context.Context, example rundex.Rebuild) {
...
-	if _, err := os.Stat(usa); errors.Is(err, os.ErrNotExist) {
-		if err := rebuild.AssetCopy(ctx, localAssets, debugAssets, upstreamAsset); err != nil {
-			log.Println(errors.Wrap(err, "failed to copy upstream asset"))
+		bi := rebuild.BuildInfo{}
+		if err := json.NewDecoder(r).Decode(&bi); err != nil {
+			log.Println(errors.Wrap(err, "decoding build info"))
 			return
 		}
-		log.Printf("downloaded rebuild and upstream:\n\t%s\n\t%s", rba, usa)
+		metadata, _ := rebuild.NewGCSStore(context.WithValue(ctx, rebuild.RunID, bi.ID), "gs://rebuild-metadata")
```
msuozzo, line 208: is there a plan to initialize this more cleanly with error-handling etc?

msuozzo, line 208: this isn't the correct URI as the bucket is "google-rebuild-metadata"

tools/ctl/ide/ui.go
```diff
@@ -159,26 +172,67 @@ func diffArtifacts(ctx context.Context, example rundex.Rebuild) {
...
+		usa = localAssets.URL(upstreamAsset).Path
+		if _, err := localAssets.Reader(ctx, rebuildAsset); errors.Is(err, os.ErrNotExist) {
+			if err := rebuild.AssetCopy(ctx, localAssets, metadata, rebuildAsset); err != nil {
+				log.Println(errors.Wrap(err, "failed to copy rebuild asset"))
+				return
+			}
+			log.Printf("downloaded rebuild: %s", rba)
+		}
+		if _, err := localAssets.Reader(ctx, upstreamAsset); errors.Is(err, os.ErrNotExist) {
+			w, _ := localAssets.Writer(ctx, upstreamAsset)
```
msuozzo, line 222: this error and the one below get ignored. might want to add checks

tools/ctl/ide/ui.go
wbxyz: Done.

tools/ctl/ide/ui.go
wbxyz: Hmm I thought I cleaned up all the error handling before mailing these but I guess I missed some. Should be fixed now

wbxyz: I thought I added this as a flag similar to the logs bucket, but that must have gotten lost in a rebase somewhere. Added the same as logs bucket using a context variable.

Long term, I think we might copy everything into the debug bucket during the build to make debugging easier and require fewer buckets. It would also make smoketest and GCB runs more similar. Out of scope for now though.

wbxyz: Oh I also changed the function to return an error and the choice to print error messages is handled by the caller now.

pkg/rebuild/rebuild/context.go
```diff
@@ -22,6 +22,7 @@ const (
 	AssetDirID
 	DebugStoreID
 	LogsBucketID
+	MetadataBucketID
```
msuozzo, line 25: you can put this off for a child change but LogsBucketID and MetadataBucketID are only used in ctl so don't really belong in the rebuild package. you can add a context file there to define them.

msuozzo (approved):

pkg/rebuild/rebuild/context.go
wbxyz: Moved into the ide package, because that's the only thing that actually reads from these.

PR #241, opened 2025-01-07
Add request logging to api.Handler function

wbxyz:
This gives us guaranteed request logging for any services that are using
the api.Handler interface.

### DIFF
--- a/internal/api/rpc.go
+++ b/internal/api/rpc.go
@@ -98,6 +98,7 @@ func Handler[I schema.Message, O any, D Dependencies](initDeps InitT[D], handler
 			http.Error(rw, http.StatusText(http.StatusBadRequest), http.StatusBadRequest)
 			return
 		}
+		log.Printf("received request: %+v", req)
 		if err := req.Validate(); err != nil {
 			log.Println(errors.Wrap(err, "validating request"))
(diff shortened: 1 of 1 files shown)


internal/api/rpc.go
```diff
@@ -98,6 +98,9 @@ func Handler[I schema.Message, O any, D Dependencies](initDeps InitT[D], handler
 			http.Error(rw, http.StatusText(http.StatusBadRequest), http.StatusBadRequest)
 			return
 		}
+		// Best-effort logging of the request, not worth further handling errors.
+		j, _ := json.Marshal(req)
```
msuozzo, line 102: hmm we might be able to get away with just the native struct output

wbxyz: Yeah "%+v" lgtm

internal/api/rpc.go
```diff
@@ -98,6 +98,7 @@ func Handler[I schema.Message, O any, D Dependencies](initDeps InitT[D], handler
 			http.Error(rw, http.StatusText(http.StatusBadRequest), http.StatusBadRequest)
 			return
 		}
+		log.Printf("received request: %+v\n", req)
```
msuozzo, line 101: super nit: for log.Printf, newlines are added so you can omit the \n

msuozzo (approved):

PR #242, opened 2025-01-08
Bump go-git version to v5.13.1

wbxyz:


wbxyz:
Actually this is already proposed in #239, I'll squash that one and revert this one.

PR #248, opened 2025-01-10
Add task queue support to run-bench command

wbxyz:


tools/benchmark/run.go
```diff
@@ -206,3 +210,99 @@ func RunBench(ctx context.Context, client *http.Client, apiURL *url.URL, set Pac
...
+					AuthorizationHeader: &taskspb.HttpRequest_OidcToken{
+						OidcToken: &taskspb.OidcToken{
+							ServiceAccountEmail: q.serviceAccountEmail,
+						},
+					},
+				},
+			},
+		},
+	}
+	createdTask, err := q.client.CreateTask(ctx, req)
```
msuozzo, line 257: super-nit: "task"?

tools/benchmark/run.go
```diff
@@ -206,3 +210,99 @@ func RunBench(ctx context.Context, client *http.Client, apiURL *url.URL, set Pac
 	go ex.Process(ctx, verdictChan, set.Packages)
 	return verdictChan, nil
 }
+
+type TaskQueue interface {
+	Append(ctx context.Context, url, body string) (*taskspb.Task, error)
```
msuozzo, line 215: nit: Append doesn't feel like the right verb. what about Add? Alternatively, Submit?

tools/benchmark/run_test.go
```diff
@@ -0,0 +1,89 @@
...
+					"https://example.com/smoketest",
+					"ecosystem=npm&id=runid&package=package_name&versions=1.0.0&versions=1.1.0",
+				},
+			},
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			queue := &mockQueue{}
+			url, _ := url.Parse("https://example.com")
```
msuozzo, line 80: unrelated but this built-in's requirement to manually check the error is a fairly annoying omission of an API. I feel like it might be nice to add a urlx.MustParse

wbxyz: Would it be better to use use a generic Must() in our tests? I don't have a strong feeling one way or the other

tools/benchmark/run.go
```diff
@@ -206,3 +210,99 @@ func RunBench(ctx context.Context, client *http.Client, apiURL *url.URL, set Pac
...
+type queue struct {
+	client              *cloudtasks.Client
+	queuePath           string
+	serviceAccountEmail string
+}
+
+func NewQueue(ctx context.Context, queuePath, serviceAccountEmail string) (TaskQueue, error) {
+	client, err := cloudtasks.NewClient(ctx)
+	if err != nil {
+		return nil, errors.Wrap(err, "TaksQueue NewClient")
```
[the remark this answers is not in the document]
wbxyz: Oops yeah.

Also, it was pretty annoying to find the right spelling of Cloud Tasks, task-queue, TaskQueue, Queue, and other variations. Suggested change is an improvement I think.

tools/benchmark/run.go
wbxyz: Add works for me.

tools/benchmark/run.go
```diff
@@ -206,3 +210,99 @@ func RunBench(ctx context.Context, client *http.Client, apiURL *url.URL, set Pac
 	go ex.Process(ctx, verdictChan, set.Packages)
 	return verdictChan, nil
 }
+
+type TaskQueue interface {
```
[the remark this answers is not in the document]
msuozzo: ah i'd originally just meant a separate file within tools/benchmark for now. that said, if you're fine moving it (and the tests), I think that could be fine.

tools/benchmark/run_test.go
msuozzo: there are other times where we want static URLs:

```
tools/ctl/ide/rebuilder.go
121:            in.URL, err = url.Parse("http://localhost:8080")

internal/timewarp/timewarp.go
42:     npmRegistry, _  = url.Parse("https://registry.npmjs.org/")
43:     pypiRegistry, _ = url.Parse("https://pypi.org/simple")
```

tools/benchmark/run.go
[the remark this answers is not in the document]
wbxyz: Tests are still benchmark specific I think. They make a mockQueue type but that's why they're not really a TaskQueue test.

tools/benchmark/run.go
```diff
@@ -10,9 +10,13 @@ import (
 	"sync"
 	"time"
 
+	cloudtasks "cloud.google.com/go/cloudtasks/apiv2"
```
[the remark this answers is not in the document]
msuozzo: sounds good. I'm fine with that

tools/benchmark/run_test.go
wbxyz: Opened https://github.com/google/oss-rebuild/pull/249

PR #249, opened 2025-01-10
Add urlx.MustParse helper

wbxyz:


internal/urlx/urlx.go
```diff
@@ -0,0 +1,12 @@
+package urlx
+
+import "net/url"
+
+// MustParse will call url.Parse and panic if there is an error, returning on success.
+func MustParse(rawURL string) *url.URL {
+	if u, err := url.Parse(rawURL); err != nil {
+		panic(err)
+	} else {
```
msuozzo, line 9: nit: you can drop the else

pkg/registry/cratesio/cratesio.go
```diff
@@ -91,7 +92,10 @@ func (r HTTPRegistry) Crate(ctx context.Context, pkg string) (*Crate, error) {
 		return nil, err
 	}
 	for i := range c.Versions {
-		downloadPath, _ := url.Parse(c.Versions[i].DownloadPath)
+		downloadPath, err := url.Parse(c.Versions[i].DownloadPath)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to parse version download path")
```
msuozzo, line 97: nit: "parsing version download path"

pkg/registry/cratesio/cratesio.go
```diff
@@ -115,7 +119,10 @@ func (r HTTPRegistry) Version(ctx context.Context, pkg, version string) (*CrateV
 	if err := json.NewDecoder(resp.Body).Decode(&v); err != nil {
 		return nil, err
 	}
-	downloadPath, _ := url.Parse(v.DownloadPath)
+	downloadPath, err := url.Parse(v.DownloadPath)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to parse version download path")
```
msuozzo, line 124: same here

tools/ctl/ctl.go
```diff
@@ -101,10 +102,7 @@ var tui = &cobra.Command{
 			tctx = context.WithValue(tctx, rebuild.DebugStoreID, "file://"+localfiles.AssetsPath())
 		} else {
 			if *debugStorage != "" {
-				u, err := url.Parse(*debugStorage)
-				if err != nil {
-					log.Fatal(errors.Wrap(err, "parsing --debug-storage as url"))
-				}
+				u := urlx.MustParse(*debugStorage)
```
msuozzo, line 105: lets revert these. panic's output isn't desirable in the CLI as it includes a full stacktrace. log.Fatal only outputs the error content.

internal/urlx/urlx.go
wbxyz: To drop the else, the url.Parse would need to be hoisted outside the if block. I weakly prefer this option, given they're the same LoC because it keeps the two exit options nested to the same depth.

If the `else` was much more complicated though, I think the guard-clause design pattern would be more compelling and getting rid of the `else` would be worth it.

msuozzo: ah right yeah. this is fine

msuozzo (approved):

Issue #250, opened 2025-01-10
oss-rebuild list command succeeds but get command fails with KMS client error


theopolis:
In a default environment, `oss-rebuild get` fails with a KMS error.

First try to list artifacts:
```sh
➜  oss-rebuild git:(main) go run cmd/oss-rebuild/main.go list pypi "ipython"                                                             
pypi/ipython/8.20.0/ipython-8.20.0-py3-none-any.whl/rebuild.intoto.jsonl
```

Then try to get:
```sh
➜  go run cmd/oss-rebuild/main.go get pypi "ipython" "8.20.0" ipython-8.20.0-py3-none-any.whl/rebuild.intoto.jsonl
2025/01/10 16:22:35 creating KMS client: credentials: could not find default credentials. See https://cloud.google.com/docs/authentication/external/set-up-adc for more information
exit status 1
```

This is with the google cloud SDK installed:
```sh
➜  oss-rebuild git:(main) hash gcloud && echo $?             
0
```

Question: is any pre-setup needed before a rebuild attestation can be inspected?

wbxyz:
Until that key visibility issue gets resolved, one option is to disable the verification on our CLI using `--verify=false`

Also FYI the artifact name in the CLI is not `.intoto` but the python generated artifact (wheel). However for convenience much of the time you can leave it our completely:

```bash
root@13517cfc4f87:/oss-rebuild#  go run cmd/oss-rebuild/main.go get pypi "ipython" "8.20.0" --verify=false --output=dockerfile
pypi artifact is being inferred as ipython-8.20.0-py3-none-any.whl
#syntax=docker/dockerfile:1.4
FROM gcr.io/cloud-builders/gsutil AS timewarp_provider
RUN gsutil cp -P gs://google-rebuild-bootstrap-tools/timewarp .
FROM alpine:3.19
COPY --from=timewarp_provider ./timewarp .
RUN <<'EOF'
 set -eux
 ./timewarp -port 8080 &
 while ! nc -z localhost 8080;do sleep 1;done
 apk add git python3
 mkdir /src && cd /src
 git clone 'https://github.com/ipython/ipython' .
 git checkout --force '19f24dd8e9dd698e62d9ef898c41dfd4545c71ab'
 /usr/bin/python3 -m venv /deps
 /deps/bin/pip install build
 /deps/bin/pip install wheel==0.42.0
 /deps/bin/pip install setuptools==67.7.2
 /deps/bin/pip install setuptools>=51.0.0

[...]

PR #251, opened 2025-01-10
Disable authentication for kms verifier

wbxyz:
People using the oss-rebuild CLI should not need an authenticated KMS
client instance. The option.WithoutAuthentication client should be
enough.

### DIFF
--- a/cmd/oss-rebuild/main.go
+++ b/cmd/oss-rebuild/main.go
@@ -133,7 +133,7 @@ func (b *Bundle) Byproduct(name string) ([]byte, error) {
 }
 
 func makeKMSVerifier(ctx context.Context, cryptoKeyVersion string) (dsse.Verifier, error) {
-	kc, err := kms.NewKeyManagementClient(ctx)
+	kc, err := kms.NewKeyManagementClient(ctx, option.WithoutAuthentication())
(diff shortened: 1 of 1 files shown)


msuozzo (approved):

msuozzo:
Wanna annotate this as fixing the issue?

cmd/oss-rebuild/main.go
```diff
@@ -133,7 +133,7 @@ func (b *Bundle) Byproduct(name string) ([]byte, error) {
 }
 
 func makeKMSVerifier(ctx context.Context, cryptoKeyVersion string) (dsse.Verifier, error) {
-	kc, err := kms.NewKeyManagementClient(ctx)
+	kc, err := kms.NewKeyManagementClient(ctx, option.WithoutAuthentication())
```
msuozzo, line 136: Have you locally tested this?

wbxyz:
> Wanna annotate this as fixing the issue?

It's not quite enough. See https://github.com/google/oss-rebuild/issues/250#issuecomment-2584456041

msuozzo:
Yeah so the API does require some form of identity which we need to resolve. The current plan is to embed the pubkey in the CLI.

PR #253, opened 2025-01-13
docs: Add note about GCP ADC requirement to README

theopolis:
As an alternative to https://github.com/google/oss-rebuild/pull/251

This adds a note about the GCP ADC requirements for fetching the KMS public signing key. It does not seem that  unauthenticated access is supported for public signing keys.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,9 @@ bin/
 # jj
 .jj/
 
+# VSCode project settings
+.vscode/
+
 # Go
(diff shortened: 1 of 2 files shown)


msuozzo (approved):

msuozzo (approved):

theopolis:
Thanks for the review and suggestion @msuozzo, just applied that now. I do not have merge/write access so do you mind merging when you have a moment later?

PR #256, opened 2025-01-13
Fix ignored PATH in version override

msuozzo:


wbxyz (approved):

pkg/rebuild/npm/strategy.go
```diff
@@ -94,8 +94,8 @@ wget -O - https://unofficial-builds.nodejs.org/download/release/v{{.NodeVersion}
 	}
 	build, err := rebuild.PopulateTemplate(`
 {{if ne .VersionOverride "" -}}
-{{- /* NOTE: Use builtin npm for 'npm version' as it wasn't introduced until NPM v6. */ -}}
-PATH=/usr/bin:/bin:/usr/local/bin /usr/bin/npm version --prefix {{.Location.Dir}} --no-git-tag-version {{.VersionOverride}}
```
theopolis, line 98: Since the `PATH` declaration is repeated a few times, thoughts on having it as a `const` so that no one forgets to update tests if this does change in the future?

msuozzo: See #260

PR #257, opened 2025-01-14
chore: Refactor maven implementation to use Registry and Rebuilder interfaces

theopolis:
Hi team, this is a work in progress so the description here and implementation/changeset will change as it gets closer to complete. For now I wanted to make sure I am on the right path.

It seems that the `maven` implementation in the code does not yet use the `Rebuilder` or `Registry` interfaces. To simplify the code this PR aims to port the maven code to implement these. This is starting with `Registry` as that is a dependency and far simpler.

This PR aims to keep the coding style, imports, and other important but non-functional elements the same.

### DIFF
--- a/pkg/rebuild/maven/infer.go
+++ b/pkg/rebuild/maven/infer.go
@@ -34,31 +34,117 @@ import (
 	"github.com/go-git/go-git/v5/storage"
 	"github.com/google/oss-rebuild/internal/uri"
 	"github.com/google/oss-rebuild/pkg/rebuild/rebuild"
-	mavenreg "github.com/google/oss-rebuild/pkg/registry/maven"
+	"github.com/google/oss-rebuild/pkg/registry/maven"
 	"github.com/pkg/errors"
 )
 
-type RepoConfig struct {
-	*git.Repository
-	URI    string
-	Dir    string
-	RefMap map[string]string
+func (Rebuilder) InferRepo(ctx context.Context, t rebuild.Target, mux rebuild.RegistryMux) (string, error) {
+	pom, err := NewPomXML(ctx, t, mux)
+	if err != nil {
+		return "", err
+	}
+	return uri.CanonicalizeRepoURI(pom.URL)
 }
 
-type BuildConfig struct {
-	Repo string
-	Dir  string
-	Ref  string
-	// TODO: Switch to an interface that can represent different build types.
-	Build MavenBuild
+func (Rebuilder) CloneRepo(ctx context.Context, t rebuild.Target, repoURI string, fs billy.Filesystem, s storage.Storer) (r rebuild.RepoConfig, err error) {
+	r.URI = repoURI
+	r.Repository, err = rebuild.LoadRepo(ctx, t.Package, s, fs, git.CloneOptions{URL: r.URI, RecurseSubmodules: git.DefaultSubmoduleRecursionDepth})
+	switch err {
+	case nil:
+	case transport.ErrAuthenticationRequired:
+		err = errors.Errorf("repo invalid or private")
+		return
+	default:
+		err = errors.Wrapf(err, "clone failed [repo=%s]", r.URI)
+	}
+	return
 }
 
-type MavenBuild struct {
-	// JDKVersion is the version of the JDK to use for the build.
-	JDKVersion string
+func (Rebuilder) InferStrategy(ctx context.Context, t rebuild.Target, mux rebuild.RegistryMux, repoConfig *rebuild.RepoConfig, hint rebuild.Strategy) (rebuild.Strategy, error) {
+	// Do pom.xml search.
+	head, _ := repoConfig.Repository.Head()
+	commitObject, _ := repoConfig.Repository.CommitObject(head.Hash())
+	_, pkgPath, err := findPomXML(commitObject, t.Package)
+	if err != nil {
+		log.Printf("pom.xml path heuristic failed [pkg=%s,repo=%s]: %s\n", t.Package, repoConfig.URI, err.Error())
+	}
+
+	// Do version heuristic search.
+	refMap, err := pomXMLSearch(t.Package, pkgPath, repoConfig.Repository)
+	if err != nil {
+		log.Printf("pom.xml version heuristic failed [pkg=%s,repo=%s]: %s\n", t.Package, repoConfig.URI, err.Error())
+	}
+
+	name, version := t.Package, t.Version
+	cfg := &MavenBuild{}
+	dir := path.Dir(pkgPath)
+	pomXMLGuess := refMap[version]
+	tagGuess, err := rebuild.FindTagMatch(name, version, repoConfig.Repository)
+	if err != nil {
+		return cfg, errors.Wrapf(err, "[INTERNAL] tag heuristic error")
+	}
+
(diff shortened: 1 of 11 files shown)


pkg/registry/maven/maven.go
```diff
@@ -16,18 +16,24 @@
...
+	"path"
 	"path/filepath"
 	"strings"
 	"time"
 
+	"github.com/google/oss-rebuild/internal/httpx"
 	"github.com/pkg/errors"
 )
 
+var registryURL, _ = url.Parse("https://search.maven.org")
```
msuozzo, line 35: we now have urlx.MustParse merged so that saves some cruft

pkg/registry/maven/maven.go
```diff
@@ -128,84 +110,124 @@ type MavenVersion struct {
...
+func (r HTTPRegistry) ReleaseFile(ctx context.Context, pkg, version string, typ string) (io.ReadCloser, error) {
 	g, a, found := strings.Cut(pkg, ":")
 	if !found {
-		err = errors.New("package identifier not of form 'group:artifact'")
-		return
+		return nil, errors.New("package identifier not of form 'group:artifact'")
 	}
 	path := filepath.Join(strings.ReplaceAll(g, ".", "/"), a, version, fmt.Sprintf("%s-%s%s", a, version, typ))
-	resp, err := http.Get("https://search.maven.org/remotecontent?filepath=" + path)
+	pathUrl, _ := url.Parse("remotecontent")
```
msuozzo, line 206: sand here

msuozzo:
Looks great so far!

theopolis:
Next step is to add some unit tests and then figure out how to run the smoke tests.

theopolis:
It looks like the implementation of the actual rebuild for `maven` was never implemented. Instead of trying to do that in this refactor PR, I saved that for a follow up.

Similarly there is a lot of commit parsing code, which might need a re-review and definitely some tests. I tried to be "minimally intrusive" in terms of changing the logic there. This way there is a clear flow from the existing code to the refactor to implement the rebuilder interface.

msuozzo:
> It looks like the implementation of the actual rebuild for `maven` was never implemented. Instead of trying to do that in this refactor PR, I saved that for a follow up.

yep totally fine

> Similarly there is a lot of commit parsing code, which might need a re-review and definitely some tests. I tried to be "minimally intrusive" in terms of changing the logic there. This way there is a clear flow from the existing code to the refactor to implement the rebuilder interface.

Yeah I actually have some unsubmitted code intended on making git history testing easier. I'll try to dust that off and see if it is still fit for purpose.

pkg/rebuild/maven/infer.go
```diff
@@ -34,121 +34,71 @@ import (
...
+func (Rebuilder) InferRepo(ctx context.Context, t rebuild.Target, mux rebuild.RegistryMux) (string, error) {
+	pom, err := NewPomXML(ctx, t, mux)
 	if err != nil {
-		return
+		return "", err
 	}
-	p, err := f.Contents()
-	if err != nil {
-		return
+	// In some projects the groupId only exists on the parent.
```
msuozzo, line 46: hmm this is a surprising place for these edits. would it be possible to move down to InferStrategy? In fact, I'm surprised these have any side effects outside this function. Is there caching somewhere that stores these changes?

pkg/rebuild/maven/infer.go
```diff
@@ -185,26 +135,78 @@ func doInference(ctx context.Context, t rebuild.Target, rcfg *RepoConfig) (Build
...
+	p, err := f.Contents()
+	if err != nil {
+		return
+	}
+	err = xml.Unmarshal([]byte(p), &pomXML)
+	return
+}
+
+func getJarJDK(ctx context.Context, name, version string, mux rebuild.RegistryMux) (string, error) {
+	//r, err := mavenreg.HTTPRegistry{}.ReleaseFile(context.Background(), name, version, mavenreg.TypeJar)
```
msuozzo, line 169: remove

pkg/rebuild/maven/strategy.go
```diff
@@ -0,0 +1,42 @@
...
+	rebuild.Location
+
+	// JDKVersion is the version of the JDK to use for the build.
+	JDKVersion string `json:"jdk_version" yaml:"jdk_version"`
+}
+
+var _ rebuild.Strategy = &MavenBuild{}
+
+func (b *MavenBuild) GenerateFor(t rebuild.Target, be rebuild.BuildEnv) (rebuild.Instructions, error) {
+	src, err := rebuild.BasicSourceSetup(b.Location, &be)
```
msuozzo, line 29: ugh so I jumped the gun and deleted this. you should be able to implement this something like:

```
func (b *MavenBuild) ToWorkflow() *rebuild.WorkflowStrategy {
	return &rebuild.WorkflowStrategy{
		Location: b.Location,
		Source: []flow.Step{{
			Uses: "git-checkout",
		}},
		Deps: []flow.Step{{
			Runs: "true",
		}},
		Build: []flow.Step{{
			Uses: "true",
		}},
	}
}

func (b *MavenBuild) GenerateFor(t rebuild.Target, be rebuild.BuildEnv) (rebuild.Instructions, error) {
}

```

pkg/registry/maven/maven.go
```diff
@@ -128,84 +65,111 @@ type MavenVersion struct {
...
-	r, err = ReleaseFile(pkg, version, TypePOM)
+// ReleaseFile returns a release file for a Maven package version.
+func (r HTTPRegistry) ReleaseFile(ctx context.Context, pkg, version string, typ string) (io.ReadCloser, error) {
+	g, a, found := strings.Cut(pkg, ":")
+	if !found {
+		return nil, errors.New("package identifier not of form 'group:artifact'")
+	}
+	pathUrl, _ := url.Parse("remotecontent")
+	pathUrl = registryURL.ResolveReference(pathUrl)
+	searchPath := filepath.Join(strings.ReplaceAll(g, ".", "/"), a, version, fmt.Sprintf("%s-%s%s", a, version, typ))
```
msuozzo, line 162: super-nit: I think this should be `path.Join` since it expects forward slashes

msuozzo (approved):
Looks great! I'd be fine merging as-is given we have some work to do on infer.go testing.

pkg/rebuild/maven/pom_test.go
```diff
@@ -0,0 +1,104 @@
...
+		t.Run(tc.testName, func(t *testing.T) {
+			ctx := context.Background()
+			target := rebuild.Target{}
+			mux := rebuild.RegistryMux{
+				Maven: TestRegistry{
+					releaseFile: tc.input,
+				},
+			}
+
+			if pom, err := NewPomXML(ctx, target, mux); err != nil && !tc.isError {
```
msuozzo, line 91: something like this would be a bit cleaner and ~equivalent

```
pom, err := NewPomXML(ctx, target, mux)
if err != nil && !tc.isError {
    t.Fatalf("NewPomXML() error = %v", err)
}
if tc.isError {
    return
}
if diff := cmp.Diff(tc.expected, pom); diff != "" {
    t.Errorf("NewPomXML() mismatch (-want +got):\n%s", diff)
}
```

PR #262, opened 2025-01-16
Refactor pypi strategy to use flow lib

msuozzo:


pkg/rebuild/pypi/strategy.go
```diff
@@ -30,42 +31,110 @@ type PureWheelBuild struct {
...
+				{{- end -}}`)[1:],
+			Needs: []string{},
+		}},
+	},
+	{
+		Name: "pypi/install-deps",
+		Steps: []flow.Step{{
+			Runs: textwrap.Dedent(`
+				{{.With.locator}}pip install build
+				{{range $req := .With.requirements | fromJSON -}}
```
wbxyz, line 99: How does an error in fromJSON get handled here?

pkg/rebuild/flow/json.go
wbxyz, line 1: Can these be private? I think the golang funcs are only used as part of resolveTemplate. Other packages reference them but only using the template strings.

pkg/rebuild/flow/flow.go
```diff
@@ -23,7 +23,12 @@ type Step struct {
 }
 
 func resolveTemplate(buf *bytes.Buffer, tmpl string, data any) error {
-	t, err := template.New("").Option("missingkey=zero").Parse(tmpl)
+	t, err := template.New("").Option("missingkey=zero").Funcs(template.FuncMap{
+		"join":     strings.Join,
+		"split":    strings.Split,
+		"fromJSON": FromJSON,
+		"toJSON":   ToJSON,
```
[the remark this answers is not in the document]
msuozzo: done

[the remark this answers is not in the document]
msuozzo: nah so template automatically returns the error when a func has a 2-tuple return val and the second is an error and non-nil. these don't (and shouldn't) panic on error.

pkg/rebuild/pypi/strategy.go
msuozzo: same answer as above

pkg/rebuild/flow/flow.go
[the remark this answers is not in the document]
wbxyz: Got it, I didn't know that behavior when writing the comment. Knowing it now I agree the non-panic version is better.

pkg/rebuild/flow/json.go
msuozzo: gonna leave this as-is. as mentioned offline, we may want to move these to a jsonx if we find sufficient internal use-cases

wbxyz (approved):

PR #266, opened 2025-01-16
Include field name in missing required field error

wbxyz:


pkg/rebuild/schema/form/form.go
```diff
@@ -92,7 +93,7 @@ func Unmarshal(v url.Values, out any) error {
 		urlval := v.Get(opt.name)
 		if urlval == "" {
 			if opt.required {
-				return ErrMissingRequired
+				return errors.Wrapf(ErrMissingRequired, "field %s not present", field.Name)
```
msuozzo, line 96: this new message duplicates the error text. we can drop it and just use field.Name.

msuozzo (approved):

PR #269, opened 2025-01-17
Refactor manualstrategy to use flow lib

msuozzo:


pkg/rebuild/rebuild/manualstrategy.go
```diff
@@ -25,18 +27,18 @@ type ManualStrategy struct {
 
 var _ Strategy = &ManualStrategy{}
 
-// GenerateFor generates the instructions for a ManualStrategy.
-func (s *ManualStrategy) GenerateFor(t Target, be BuildEnv) (Instructions, error) {
-	src, err := BasicSourceSetup(s.Location, &be)
```
wbxyz, line 30: I think this was the last usage of BasicSourceSetup? So can maybe be deleted in this PR or a followup.

msuozzo: oh good point! will change in a follow-up

msuozzo: #275

PR #270, opened 2025-01-17
Refactor debian to use flow lib

msuozzo:


pkg/rebuild/cratesio/infer.go
```diff
@@ -253,15 +253,15 @@ func (Rebuilder) InferStrategy(ctx context.Context, t rebuild.Target, mux rebuil
 			return nil, errors.New("rust version heuristic failed")
 		}
 	}
-	return &CratesIOCargoPackage{
```
wbxyz, line 256: Different change?

Also why prefer workflow strategies over the ecosystem specific?

wbxyz (commented):
Let's get #268 merged to make it clear this PR isn't also adding tests.

pkg/rebuild/cratesio/infer.go
msuozzo: yep done

pkg/rebuild/debian/strategy.go
```diff
@@ -41,63 +42,84 @@ type DebianPackage struct {
...
+		}},
+	},
+	{
+		Name: "debian/build/package",
+		Steps: []flow.Step{{
+			Runs: textwrap.Dedent(`
+				set -eux
+				cd */
+				debuild -b -uc -us
+				expected=$(echo '{{.With.targetPath}}' | sed -E 's/\+b[0-9]+(_[^_]+\.deb)$/\1/g')
```
wbxyz, line 120: Why move this parsing inside the generated instructions?

Leaving it outside the instructions helps simplify the template and allows this regex to be tested independently.

Adding it to the instructions means there is an awkward `sed` and `mv` that are no-ops for most debian packages.

wbxyz (approved):

pkg/rebuild/debian/strategy.go
msuozzo: actually this made me realize that we can implement this in the golang template (admittedly in its awkward syntax) so I've removed this requirement and the test modifications are now ~removed.

PR #271, opened 2025-01-21
Define Semver.Compare

msuozzo:


theopolis (approved):
Looks good, the unit tests would catch if the logic changed here.

PR #272, opened 2025-01-21
Add new setuptools generator

msuozzo:


pkg/rebuild/pypi/infer.go
```diff
@@ -211,6 +212,11 @@ func inferRequirements(name, version string, zr *zip.Reader) ([]string, error) {
 	if err != nil {
 		return nil, errors.Wrapf(err, "[INTERNAL] Failed to get upstream generator")
 	}
+	// Determine setuptools version.
+	if slices.ContainsFunc(reqs, func(s string) bool { return strings.HasPrefix(s, "setuptools==") }) {
```
wbxyz, line 216: I don't think all the version requirements in our strategies are "==" currently, should this be a regex match instead of prefix?

wbxyz (approved):

pkg/rebuild/pypi/infer.go
msuozzo: at this point in the function, this is only populated by getGenerator which, in the setuptools case just added, uses ==.

PR #281, opened 2025-01-22
Add crates inference test

msuozzo:
Increases package coverage from 17% -> 48%

### DIFF
--- a/pkg/rebuild/cratesio/infer_test.go
+++ b/pkg/rebuild/cratesio/infer_test.go
@@ -1,0 +1,258 @@
+package cratesio
+
+import (
+	"archive/tar"
+	"bytes"
+	"context"
+	"io"
+	"net/http"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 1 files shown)


pkg/rebuild/cratesio/infer.go
```diff
@@ -94,7 +92,7 @@ func (Rebuilder) CloneRepo(ctx context.Context, t rebuild.Target, repoURI string
 			log.Printf("Cargo.toml version heuristic failed [pkg=%s,repo=%s]: %s\n", t.Package, r.URI, err.Error())
 		}
 	}
-	return
```
wbxyz, line 97: nit: Now that we don't have any bare returns, should we change the args to be anonymous?

wbxyz (approved):

pkg/rebuild/cratesio/infer.go
msuozzo: idr why that's in the PR so i'll deal with this change in a separate one

PR #289, opened 2025-01-28
Record rebuild start time

wbxyz:
When executing streaming-style rebuilds, we will use a fixed runID. In these cases, we will likely want to group rebuilds based on when they were started, rather than when they were completed.

### DIFF
--- a/pkg/rebuild/schema/schema.go
+++ b/pkg/rebuild/schema/schema.go
@@ -256,7 +256,8 @@ type RebuildAttempt struct {
 	RunID           string          `firestore:"run_id,omitempty"`
 	BuildID         string          `firestore:"build_id,omitempty"`
 	ObliviousID     string          `firestore:"oblivious_id,omitempty"`
-	Created         int64           `firestore:"created,omitempty"`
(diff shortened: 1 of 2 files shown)


msuozzo:
"Created" referred to the date at which the entry in Firestore was created. You could instead add a Started (and potentially Finished) field. I don't feel strongly about keeping around a Created but I think changing the semantics in-place has some negative side effects.

wbxyz:
Sounds good to me. I added a new field called "Started" and reverted "Created" to existing behavior.

pkg/rebuild/schema/schema.go
```diff
@@ -256,6 +256,7 @@ type RebuildAttempt struct {
 	RunID           string          `firestore:"run_id,omitempty"`
 	BuildID         string          `firestore:"build_id,omitempty"`
 	ObliviousID     string          `firestore:"oblivious_id,omitempty"`
+	Started         int64           `firestore:"started,omitempty"`
 	Created         int64           `firestore:"created,omitempty"`
```
msuozzo, line 260: nit: want to add super brief comments differentiating these two cases?

msuozzo (approved):

PR #290, opened 2025-01-28
Add schema.ExecutionMode to generalize BenchmarkMode

wbxyz:
The type of execution is not directly tied to benchmarks. This change helps decouple the benchmark construct from code locations that switch between execution modes.

### DIFF
--- a/pkg/rebuild/schema/schema.go
+++ b/pkg/rebuild/schema/schema.go
@@ -268,3 +268,11 @@ type Run struct {
 	Type          string `firestore:"run_type,omitempty"`
 	Created       int64  `firestore:"created,omitempty"`
 }
+
+// Execution mode describes the manner in which a rebuild happens.
+type ExecutionMode string
+
+const (
+	SmoketestMode ExecutionMode = "smoketest" // No attestations, faster.
+	AttestMode    ExecutionMode = "attest"    // Creates attestations, slower.
+)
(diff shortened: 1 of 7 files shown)


pkg/rebuild/schema/schema.go
```diff
@@ -267,3 +267,11 @@ type Run struct {
 	Type          string `firestore:"run_type,omitempty"`
 	Created       int64  `firestore:"created,omitempty"`
 }
+
+// Execution mode describes the manner in which a rebuild happens.
+type ExecutionMode string
```
msuozzo, line 272: is this required? i believe we spoke offline and i don't quite remember if there was a requirement that this get moved here. i don't object in priciple but wanted to understand where the dep gets used.

looking at this chain, it looks the other usage is in tools/pypi_rss/main.go which I think would make this better-suited to moving to tools dir or, i suppose, leaving it here and then refactoring `benchmark` when the time comes.

wbxyz: Originally I needed it for some functionality in the `rebuild` package which would have introduced a dependency cycle on `benchmark`.

I ended up not needing to use the code in `rebuild` but the experience nudged me to move this enum out anyway. I'd lean towards keeping it in schema or models.go because we might want to use it in places outside tools in the future. No strong preference where specifically we put it in pkg/ though.

msuozzo: sg, i'm fine moving it here

msuozzo (approved):

PR #291, opened 2025-01-28
Change internal taskqueue to accept a schema.Message

wbxyz:
This makes the taskqueue Add() function similar to stubs. As used currently, we're *always* enqueuing a schema.Message type so this helps remove boilerplate.


msuozzo:
hmm this inverts the layering i was envisioning for this lib (for now). schema.Message _could_ be moved to a separate pkg that didn't also define and rely on rebuild-related things. as it stands, though, this would make our taskqueue reliant on all rebuild impls. this feels a bit smelly but not particularly bad.

i'd say for now this is fine but can you open an issue to move schema.Message and form out of pkg/rebuild?

msuozzo (approved):

PR #292, opened 2025-01-28
Create rebuild feed library

wbxyz:


pkg/feed/tracker.go
```diff
@@ -0,0 +1,27 @@
+package feed
+
+import "github.com/google/oss-rebuild/pkg/rebuild/rebuild"
+
+type Tracker interface {
```
msuozzo, line 5: i think we need some docs here. the current impl seems to use package-based tracking but i can imagine that we'd want something else in the future.

pkg/feed/grouping.go
```diff
@@ -0,0 +1,59 @@
...
+	"slices"
+	"sort"
+
+	"github.com/google/oss-rebuild/pkg/rebuild/rebuild"
+	"github.com/google/oss-rebuild/pkg/rebuild/schema"
+)
+
+func GroupForSmoketest(targets []rebuild.Target, runID string) []schema.SmoketestRequest {
+	var res []schema.SmoketestRequest
+	grouped := map[rebuild.Ecosystem]map[string][]rebuild.Target{}
```
msuozzo, line 13: I think we should make() this. i don't think there's a real difference but this is supposedly used mostly for statically-sized maps whereas ours is expect to grow.

pkg/feed/grouping.go
```diff
@@ -0,0 +1,59 @@
...
+	for _, t := range targets {
+		if _, ok := grouped[t.Ecosystem]; !ok {
+			grouped[t.Ecosystem] = map[string][]rebuild.Target{}
```
msuozzo, line 16: same here

pkg/feed/grouping.go
```diff
@@ -0,0 +1,59 @@
...
+			for _, t := range targets {
+				req.Versions = append(req.Versions, t.Version)
+			}
+			slices.Sort(req.Versions)
+			res = append(res, req)
+		}
+	}
+	// Sort the final result for stability
+	sort.Slice(res, func(i, j int) bool {
+		if res[i].Ecosystem != res[j].Ecosystem {
```
msuozzo, line 37: cmp.Or

pkg/feed/tracker.go
```diff
@@ -0,0 +1,27 @@
...
+	IsTracked(rebuild.Target) (bool, error)
+}
+
+type StaticListTracker struct {
```
msuozzo, line 9: Can we avoid this concrete impl altogether?

I feel like we could replace this with some glue for allowing users to create a tracker from a closure. really we just want a func(rebuild.Target) (bool, error) and that's pretty easy to define inline without needing to abstract it away. leaking signatures like "map[rebuild.Ecosystem]map[string]bool" into library code isn't my favorite...

feed.TrackerFrom(func(t rebuild.Target) (bool, error) {
  ...
})

(or TrackerFromFunc or TrackerFunc but the latter sounds like a type a la http.HandlerFunc)

wbxyz: In a not-too distant PR, we will have a GCP-backed tracker, that caches the tracked list either on disk or in memory. In that case I think we'll definitely want to have a concrete impl that requires more state than just a function.

I'd lean towards keeping the static list tracker here, at least for now. That would keeping the tests of the behavior here, and gives a reference to compare the more complex GCP backed impl later. After the GCP tracker is implemented, potentially we can drop this entirely, or figure out future refactors we want to do, thoughts?

pkg/feed/grouping.go
wbxyz: Sounds good. I also changed GroupForAttest to pre-allocate len(targets) because we know exactly how long it'll end up being.

msuozzo (approved):

PR #293, opened 2025-01-28
PyPI RSS tracker

wbxyz:


tools/pypi_rss/main.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+func (l *listener) trim() error {
+	// TODO: trim handled
+	return nil
+}
+
+func (l *listener) handle(ctx context.Context, events []Event) error {
+	var count int
+	toRun := []rebuild.Target{}
```
msuozzo, line 71: var toRun []rebuild.Target

tools/pypi_rss/main.go
```diff
@@ -0,0 +1,224 @@
...
+	"github.com/google/oss-rebuild/pkg/rebuild/schema"
+	"github.com/google/oss-rebuild/tools/benchmark"
+	"github.com/pkg/errors"
+)
+
+var (
+	versionURLRegex = regexp.MustCompile(`https://pypi\.org/project/([^/]+)/([^/]+)/`)
+	apiURI          = flag.String("api", "", "OSS Rebuild API endpoint URI")
+	executionMode   = flag.String("execution-mode", "attest", "[attest|smoketest] the mode in which to execute rebuilds for incoming python updates")
+	taskQueuePath   = flag.String("task-queue", "", "the path identifier of the task queue to use")
```
msuozzo, line 29: i think the correct GCP term is "resource path" and I think considering this only has a single supported platform, you should specify the platform and product that you're looking for here.

tools/pypi_rss/main.go
```diff
@@ -0,0 +1,224 @@
...
+	taskQueueEmail  = flag.String("task-queue-email", "", "the email address of the serivce account Cloud Tasks should authorize as")
```
msuozzo, line 30: oof this has to be precise given how fiddly this is. i think this is the best i can do: "the service account email used as the identity for Cloud Tasks-initiated calls"

tools/pypi_rss/main.go
```diff
@@ -0,0 +1,224 @@
...
+	refBench        = flag.String("benchmark", "", "reference benchmark describing the packages we're interested in")
```
msuozzo, line 31: hmm not sure what "reference" means here. can we remove it? "benchmark defining the set of tracked packages" might be how i'd write it

tools/pypi_rss/main.go
```diff
@@ -0,0 +1,224 @@
+package main
```
msuozzo, line 1: could we add a brief comment talking about what this binary does and pointing to the pypi RSS feed docs?

tools/pypi_rss/main.go
```diff
@@ -0,0 +1,224 @@
...
+	latestEtag string
+	handled    map[string]map[string]bool
+	tracker    feed.Tracker
+	apiURL     *url.URL
+	mode       schema.ExecutionMode
+	queue      taskqueue.Queue
+}
+
+func (l *listener) trim() error {
+	// TODO: trim handled
```
msuozzo, line 65: what does this mean?

tools/pypi_rss/main.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+	var events []Event
+	for _, item := range data.Channel.Items {
+		eventTime, err := time.Parse(time.RFC1123, item.PubDate)
+		if err != nil {
+			return nil, errors.Wrap(err, "parsing date")
+		}
+		if matches := versionURLRegex.FindStringSubmatch(item.Link); len(matches) == 3 {
+			events = append(events, Event{
+				Pkg:     matches[1],
```
msuozzo, line 162: named groups instead?

tools/pypi_rss/main.go
```diff
@@ -0,0 +1,224 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "parsing XML")
+	}
+	var events []Event
+	for _, item := range data.Channel.Items {
+		eventTime, err := time.Parse(time.RFC1123, item.PubDate)
+		if err != nil {
+			return nil, errors.Wrap(err, "parsing date")
+		}
+		if matches := versionURLRegex.FindStringSubmatch(item.Link); len(matches) == 3 {
```
msuozzo, line 160: hmm can this be matches != nil? i also might pull the else up into a guard clause instead of if..else

tools/pypi_rss/main.go
```diff
@@ -0,0 +1,224 @@
...
+		}
+	}
+	slices.SortFunc(events, func(a, b Event) int {
+		return a.Time.Compare(b.Time)
+	})
+	return events, nil
+}
+
+func (l *listener) poll(ctx context.Context) error {
+	events, err := l.fetch(http.DefaultClient)
```
msuozzo, line 177: if this http client is a member on the listener, that should make testing easier

tools/pypi_rss/main.go
```diff
@@ -0,0 +1,224 @@
...
+func main() {
+	flag.Parse()
+	ctx := context.Background()
+	queue, err := taskqueue.NewQueue(ctx, *taskQueuePath, *taskQueueEmail)
+	if err != nil {
+		log.Fatal(errors.Wrap(err, "creating task queue"))
+	}
+	var tracker feed.Tracker
+	{
+		tracked := map[rebuild.Ecosystem]map[string]bool{}
```
msuozzo, line 194: nit: make

tools/pypi_rss/main.go
```diff
@@ -0,0 +1,224 @@
...
+			}
+		}
+		tracker = feed.NewStaticListTracker(tracked)
+	}
+	mode := schema.ExecutionMode(*executionMode)
+	if mode != schema.AttestMode && mode != schema.SmoketestMode {
+		log.Fatalf("--execution-mode must be '%s' or '%s' but got '%s'", schema.SmoketestMode, schema.AttestMode, *executionMode)
+	}
+	l := &listener{
+		url:     "https://pypi.org/rss/updates.xml",
```
msuozzo, line 210: hoist to constants

tools/pypi_rss/main.go
```diff
@@ -0,0 +1,224 @@
...
+	"github.com/google/oss-rebuild/internal/urlx"
+	"github.com/google/oss-rebuild/pkg/feed"
+	"github.com/google/oss-rebuild/pkg/rebuild/rebuild"
+	"github.com/google/oss-rebuild/pkg/rebuild/schema"
+	"github.com/google/oss-rebuild/tools/benchmark"
+	"github.com/pkg/errors"
+)
+
+var (
+	versionURLRegex = regexp.MustCompile(`https://pypi\.org/project/([^/]+)/([^/]+)/`)
```
msuozzo, line 26: move to a separate block to keep this one for just flags

tools/pypi_rss/main.go
```diff
@@ -0,0 +1,224 @@
...
+		}
+		tracker = feed.NewStaticListTracker(tracked)
+	}
+	mode := schema.ExecutionMode(*executionMode)
+	if mode != schema.AttestMode && mode != schema.SmoketestMode {
+		log.Fatalf("--execution-mode must be '%s' or '%s' but got '%s'", schema.SmoketestMode, schema.AttestMode, *executionMode)
+	}
+	l := &listener{
+		url:     "https://pypi.org/rss/updates.xml",
+		handled: map[string]map[string]bool{},
```
msuozzo, line 211: nit: make

tools/pypi_rss/main.go
```diff
@@ -0,0 +1,224 @@
...
+		apiURL:  urlx.MustParse(*apiURI),
+		mode:    mode,
+		queue:   queue,
+	}
+	for {
+		if err := l.poll(ctx); err != nil {
+			log.Printf("Failed to check latest feed: %v", err)
+			return
+		}
+		time.Sleep(5 * time.Minute)
```
msuozzo, line 222: make this  a constant and/or flag

msuozzo (approved):

tools/pypi_rss/main.go
wbxyz: Changed to member on the listener, agreed that it's better to have a member in case we want to inject an http client, but I don't think it changes the tests in this particular case because we can't use the httpxtest library here because we want to also test the behavior of the request and response tags.

msuozzo (approved):

PR #295, opened 2025-01-28
Add interface for patching env vars in network proxy

andusy:
Currently the proxy is only able to patch environment variables into the docker container by providing a key and mapping it to the certificate truststore. Allow key-value pairs through a different flag to support custom environment variables.

### DIFF
--- a/pkg/proxy/docker/docker.go
+++ b/pkg/proxy/docker/docker.go
@@ -55,7 +55,7 @@ const (
 	proxyCertPath    = "/var/cache/proxy.crt"
 	proxyCertJKSPath = "/var/cache/proxy.crt.jks"
 	// Official interface for providing additional args to JVMs.
-	javaEnvVar = "JAVA_TOOL_OPTIONS"
+	javaTruststoreEnvVar = "JAVA_TOOL_OPTIONS"
 	// Env var to which docker requests will be sent by the docker CLI.
 	dockerEnvVar = "DOCKER_HOST"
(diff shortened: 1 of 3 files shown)


pkg/proxy/docker/docker.go
```diff
@@ -419,22 +419,24 @@ type patchSet struct {
...
 
 // ContainerTruststorePatcherOpts defines the optional parameters for creating a ContainerTruststorePatcher.
 type ContainerTruststorePatcherOpts struct {
-	EnvVars         []string
-	JavaEnvVar      bool
-	RecursiveProxy  bool
-	NetworkOverride string
+	EnvVars           []string
+	TrustStoreEnvVars []string
+	JavaEnvVar        bool
```
msuozzo, line 437: I think JavaEnvVar now needs "Truststore" in the name

pkg/proxy/docker/docker.go
```diff
@@ -419,22 +419,24 @@ type patchSet struct {
...
 }
 
 // ContainerTruststorePatcherOpts defines the optional parameters for creating a ContainerTruststorePatcher.
 type ContainerTruststorePatcherOpts struct {
-	EnvVars         []string
-	JavaEnvVar      bool
-	RecursiveProxy  bool
-	NetworkOverride string
+	EnvVars           []string
+	TrustStoreEnvVars []string
```
msuozzo, line 436: nit: I think here we're considering "truststore" as one word so the capitalization should reflect that

msuozzo (approved):

PR #298, opened 2025-02-03
Move Message and `form` into `internal/api` package

### LINKED ISSUE #297: Move schema.Message and `form` outside `rebuild/schema`
The Message interface and `form` marshalling library are not specific to rebuilding behavior or the rebuild service interfaces. We should move them outside the `rebuild/schema` package to avoid forcing users import the entire rebuild interface when only wanting to use the `form` marshaling functionality.

wbxyz:
Fixes #297


internal/form/form.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/pkg/errors"
+)
+
+var (
+	ErrInvalidType      = errors.New("invalid type")
+	ErrUnsupportedField = errors.New("unsupported field")
+	ErrMissingRequired  = errors.New("missing required field")
+)
+
+type Message interface {
```
msuozzo, line 18: hmm i don't love form.Message. Message isn't really defined in our system as "something that can be form-encoded" but is intended to be something like protobuf's Message i.e. a generic struct meant for use as param and return.

what if we move both of these under api? Message in the top-level pkg with form nested underneath?

msuozzo (approved):

PR #303, opened 2025-02-05
Add prebuild image resources in tf

msuozzo:


deployments/main.tf
```diff
@@ -302,6 +347,7 @@ resource "terraform_data" "image" {
   lifecycle {
     replace_triggered_by = [
       terraform_data.service_version.output,
+      terraform_data.prebuild_version.output,
```
wbxyz, line 350: Does this mean they all get replaced if only one of service_version or prebuild_version changes?

wbxyz (approved):

build/package/Dockerfile.timewarp
wbxyz, line 1: Just trying to understand the motivation for creating images of these utils, do we intend to run them as containers? Or is this just to keep all our releases uniform regardless of what we actually end up needing?

wbxyz: Oh I guess this is now one step of the binary release? Step one is we build a dedicated image for each binary using these docker files, then step two is the "binary" tf resource unpacks the binary from these images before uploading the binary to GCS. Is that a correct understanding?

deployments/main.tf
msuozzo: no this just means the check gets triggered. existing uploads will exist and, thus, be skipped.

build/package/Dockerfile.timewarp
msuozzo: yeah uniform build, versioning, and releasing.

deployments/main.tf
wbxyz: Ah, that makes sense, although I'd think terraform's choice of "replace_triggered_by" is a confusing name for that behavior.

msuozzo: well it's a name that comes with the territory of local-exec. it doesn't really fit in the tf resource model so we need to "replace" the current state any time we want to validate we're in the correct state. we could componentize this and split apart the two image groups but the relative cost of validating that the images do still exist is quite low for the simplicity it allows.

PR #306, opened 2025-02-05
 Record service version in provenance 

msuozzo:


cmd/api/main.go
```diff
@@ -53,11 +54,17 @@ var (
 	logsBucket            = flag.String("logs-bucket", "", "GCS bucket for rebuild logs")
 	debugStorage          = flag.String("debug-storage", "", "if provided, the location in which rebuild debug info should be stored")
 	prebuildBucket        = flag.String("prebuild-bucket", "", "GCS bucket from which prebuilt build tools are stored")
+	serviceRepo           = flag.String("service-repo", "", "repo from which the service was built")
```
wbxyz, line 57: Would it be possible to embed this information during build time?

I think one way to do it would be for terraform to write the variable into a file before calling the build, then using the embed libarary to populate a constant.

It looks like another way to do it would be to use the "-X" flag on the linker to specify the value: https://pkg.go.dev/cmd/link

Maybe something like

```
SERVICE_VERSION=${service_version} ... go build ... -ldflags="-X cmd/${each.value.image}/main.ServiceVersion=$SERVICE_VERSION" ...
```

msuozzo: oh nice! I'd thought about embedding it but had in mind something a lot more complicated than that.

that said, there are two different ways to implement this and both would have tradeoffs:

1. the docker context would receive the .git/ folder (i.e. the full git repo state) and extract the commit itself. Notably, this isn't the default for [docker builds](https://docs.docker.com/build/concepts/context/#keep-git-directory) as it can bloat the context and break layer caching.
2. inject the commit hash as an arg to the dockerfile. at this point, though, we're kinda just moving the "just trust us" one layer up. I think there are benefits to this approach but we do still have to live with some degree of risk related to inauthentic information.

I think this requires some more refactoring so I'll defer to a child PR to investigate.

wbxyz: I'd lean towards option 2 I think, it seems a little more straightforward and exposes less complexity to the docker context.

Deferring to child PR SGTM.

wbxyz (approved):

PR #322, opened 2025-02-06
Support non-canonical source repo URIs

msuozzo:


cmd/api/main.go
```diff
@@ -138,9 +138,26 @@ func RebuildPackageInit(ctx context.Context) (*apiservice.RebuildPackageDeps, er
...
+	if BuildRepo == "" {
+		return nil, errors.New("empty service repo")
+	}
+	if repoURI, err := url.Parse(BuildRepo); err != nil {
+		return nil, errors.Wrap(err, "parsing service repo URI")
+	} else {
+		switch repoURI.Scheme {
+		case "file":
+			// Redact file:// schemes
+			serviceRepo = ""
```
wbxyz, line 151: Can we make this something informative, like "debug-build" or "local-checkout"?

Leaving the location's Repo field empty is hard to differentiate from the zero value.

wbxyz (approved):

cmd/api/main.go
msuozzo: i'm open to other options here but having a non-URI in an otherwise consistently URI field is not my preference. what's the benefit in differentiating Location.Repo from the zero value?

separately, I'm considering changing the attestation publication path somehow in these cases since I'm concerned we'd publish using the prod key with a non-mint build.

wbxyz: Keeping it a valid URI is a good point, one option is that we redact the "real" file URI with a canonical file URI like "file://local" or something.

Leaving serviceRepo empty while proceeding with a non-empty BuildVersion and ultimately logging the d.ServiceRepo in an attestation seems like a bug (what good is the build version if the repo is empty?)

That said, I think the main (only?) place this ends up is attesations, in which case changing the attestation publication path sounds like it would mitigate that strangeness.

wbxyz (approved):

cmd/api/main.go
wbxyz: Keeping it a valid URI is a good point, one option is that we redact the "real" file URI with a canonical file URI like "file://local" or something.

Leaving serviceRepo empty while proceeding with a non-empty BuildVersion and ultimately logging the d.ServiceRepo in an attestation seems like a bug (what good is the build version if the repo is empty?)

That said, I think the main (only?) place this ends up is attesations, in which case changing the attestation publication path sounds like it would mitigate that strangeness.

msuozzo: yeah per offline discussion, the flag guard should also work to prevent us from unintentionally leaking a local path to an attestation.

wbxyz (approved):

PR #323, opened 2025-02-06
Add JAR manifest parser

msuozzo:


pkg/archive/manifest.go
```diff
@@ -0,0 +1,239 @@
...
+	"io"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+// Implements MANIFEST.MF spec: https://docs.oracle.com/javase/8/docs/technotes/guides/jar/jar.html#JARManifest
+
+// Section represents a section in the manifest file
+type Section struct {
```
wbxyz, line 19: Not a request for this PR, but having a generic "ordered map" would be useful elsewhere in our codebase. The pypi RSS feed needs something like this for the set of processed events so it knows which events it should evict over time, and debian's DSC parser also does something similar.

pkg/archive/manifest.go
```diff
@@ -0,0 +1,239 @@
...
+type Manifest struct {
+	MainSection     *Section
+	EntrySections   []*Section
+	OriginalContent []byte // Keep original content for modification
+}
+
+// NewManifest creates a new empty manifest
+func NewManifest() *Manifest {
+	return &Manifest{
+		MainSection:   &Section{Attributes: make(map[string]string)},
```
wbxyz, line 72: NewSection()?

wbxyz (approved):

pkg/archive/manifest.go
msuozzo: which exact properties are you interesting in keep ordered? i feel like the two other cases differ in their requirements enough that it might be suboptimal to try to reuse that same abstraction. it's an idea tho

PR #333, opened 2025-02-10
Set posix mode on bootstrap-tools

wbxyz:
The gsutil cp command supports -P to copy the posix mode of the source
filesystem, but we're reading the file from a pipe so that's not
available. Instead we can set the posix mode after it's uploaded.

### DIFF
--- a/deployments/main.tf
+++ b/deployments/main.tf
@@ -381,7 +381,8 @@ resource "terraform_data" "binary" {
         docker save ${each.value.image}:${each.value.version} | \
           tar -xO --wildcards "*/layer.tar" | \
           tar -xO ${each.value.name} | \
-          gcloud storage cp - $path
+          gcloud storage cp - $path && \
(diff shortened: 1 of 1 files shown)


deployments/main.tf
```diff
@@ -382,6 +382,7 @@ resource "terraform_data" "binary" {
           tar -xO --wildcards "*/layer.tar" | \
           tar -xO ${each.value.name} | \
           gcloud storage cp - $path
+        gcloud storage objects update $path --custom-metadata=goog-reserved-posix-mode=550
```
msuozzo, line 385: && this with the previous command

msuozzo (approved):

deployments/main.tf
wbxyz: Done.

msuozzo: can you move the && to the previous line? that's the style used in the script above

PR #334, opened 2025-02-10
Add the steps format on infer

wbxyz:
This makes it easier to debug the core GCB invocation, without creating
unwanted artifacts like container images, attestations, or debug
objects.

### DIFF
--- a/tools/ctl/ctl.go
+++ b/tools/ctl/ctl.go
@@ -29,6 +29,7 @@ import (
 	"github.com/google/oss-rebuild/internal/api/inferenceservice"
 	"github.com/google/oss-rebuild/internal/oauth"
 	"github.com/google/oss-rebuild/internal/taskqueue"
+	"github.com/google/oss-rebuild/internal/textwrap"
 	"github.com/google/oss-rebuild/pkg/rebuild/rebuild"
 	"github.com/google/oss-rebuild/pkg/rebuild/schema"
 	"github.com/google/oss-rebuild/tools/benchmark"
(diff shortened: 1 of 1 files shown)


tools/ctl/ctl.go
```diff
@@ -547,6 +549,50 @@ var infer = &cobra.Command{
...
+			if err != nil {
+				log.Fatal(errors.Wrap(err, "generating dockerfile"))
+			}
+			buildScript := fmt.Sprintf(textwrap.Dedent(`
+				#!/usr/bin/env bash
+				set -eux
+				cat <<'EOS' | docker buildx build --tag=img -
+				%s
+				EOS
+				docker run --name=container img
```
msuozzo, line 580: this won't allow access to the output artifact, correct?

tools/ctl/ctl.go
```diff
@@ -547,6 +549,50 @@ var infer = &cobra.Command{
...
+				Artifact:  *artifact,
+			}
+			s, err := resp.Strategy()
+			if err != nil {
+				log.Fatal(errors.Wrap(err, "parsing strategy"))
+			}
+			if s == nil {
+				log.Fatal("no strategy")
+			}
+			in := rebuild.Input{
```
msuozzo, line 566: nit: one line

tools/ctl/ctl.go
```diff
@@ -547,6 +549,50 @@ var infer = &cobra.Command{
 				log.Fatal(errors.Wrap(err, "generating dockerfile"))
 			}
 			cmd.OutOrStdout().Write([]byte(dockerfile))
+		case "steps":
```
msuozzo, line 552: nit: can you make this less general? build-only-steps or, hopefully, something better than that. "steps" makes me think it's the actual GCB steps that would be used in the build but it's kind of a stripped down, simplified build. maybe debug-steps?

msuozzo (approved):

tools/ctl/ctl.go
wbxyz: Yeah I don't know of a great option, but debug-steps seems like an improvement to indicate the artifact isn't available and no verification is happening.

PR #345, opened 2025-02-13
Patch debrebuild to add sleep

wbxyz:
This mitigates what seems to be a race condition in mmdebstrap, where the chroot isn't completely setup by the time our customize-hook runs.

### DIFF
--- a/pkg/rebuild/debian/strategy.go
+++ b/pkg/rebuild/debian/strategy.go
@@ -4,6 +4,8 @@
 package debian
 
 import (
+	"encoding/base64"
+	"fmt"
 	"path"
 	"regexp"
 
@@ -81,6 +83,9 @@ func (b *Debrebuild) ToWorkflow() *rebuild.WorkflowStrategy {
 				"buildinfoMd5": b.BuildInfo.MD5,
 			},
 		}},
+		Deps: []flow.Step{{
+			Uses: "debian/deps/patch-debrebuild",
+		}},
 		Build: []flow.Step{{
 			Uses: "debian/build/debrebuild",
 			With: map[string]string{
@@ -138,6 +143,16 @@ var toolkit = []*flow.Tool{
 		}},
 	},
 	{
+		Name: "debian/deps/patch-debrebuild",
+		Steps: []flow.Step{{
+			Runs: fmt.Sprintf(`echo %s | base64 -d | patch /usr/bin/debrebuild`, base64.StdEncoding.EncodeToString([]byte(`@@ -725,2 +725,3 @@
+         ),
++        '--customize-hook=sleep 10',
+         '--customize-hook=chroot "$1" sh -c "'`))),
+			Needs: []string{"devscripts=2.25.2"},
+		}},
+	},
+	{
 		Name: "debian/build/handle-binary-version",
 		Steps: []flow.Step{{
 			Runs: textwrap.Dedent(`
@@ -169,7 +184,7 @@ var toolkit = []*flow.Tool{
 		Steps: []flow.Step{
 			{
 				Runs:  "debrebuild --buildresult=./out --builder=mmdebstrap {{ .With.buildinfo }}",
-				Needs: []string{"devscripts", "apt-utils", "mmdebstrap"},
+				Needs: []string{"devscripts=2.25.2", "apt-utils", "mmdebstrap"},
 			},
 			{
 				Uses: "debian/build/handle-binary-version",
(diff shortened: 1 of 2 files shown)


pkg/rebuild/debian/strategy.go
```diff
@@ -81,13 +81,24 @@ func (b *Debrebuild) ToWorkflow() *rebuild.WorkflowStrategy {
...
 			},
 		}},
-		Build: []flow.Step{{
-			Uses: "debian/build/debrebuild",
+		Deps: []flow.Step{{
+			Uses: "debian/deps/install",
 			With: map[string]string{
-				"targetPath": "{{.Target.Artifact}}",
-				"buildinfo":  path.Base(b.BuildInfo.URL),
+				"requirements": flow.MustToJSON([]string{"devscripts", "apt-utils", "mmdebstrap"}),
```
msuozzo, line 87: I think you could also add these as system deps (`Needs`) of the debrebuild tool which would mean you could drop this step.

pkg/rebuild/debian/strategy.go
```diff
@@ -81,13 +81,24 @@ func (b *Debrebuild) ToWorkflow() *rebuild.WorkflowStrategy {
...
 			},
 		}},
+		Build: []flow.Step{
+			{
+				Uses: "debian/patch/debrebuild",
```
msuozzo, line 92: i'd name this debian/deps/patch-debrebuild and move this ref to Deps

the second path element was meant to be a phase grouping and I think in this case the patching is really more setting up the dep than it is part of the build

pkg/rebuild/debian/strategy.go
```diff
@@ -137,6 +148,13 @@ var toolkit = []*flow.Tool{
 			Needs: []string{},
 		}},
 	},
+	{
+		Name: "debian/patch/debrebuild",
+		Steps: []flow.Step{{
+			Runs:  "echo 'QEAgLTcyMyw2ICs3MjMsNyBAQAogICAgICAgICAgICAgICAuIChTdHJpbmc6OlNoZWxsUXVvdGU6OnNoZWxsX3F1b3RlICRkc2NfZm5hbWUpCiAgICAgICAgICAgICAgIC4gIiBcIlwkMVwiIgogICAgICAgICApLAorICAgICAgICAnLS1jdXN0b21pemUtaG9vaz1zbGVlcCAxMCcsCiAgICAgICAgICctLWN1c3RvbWl6ZS1ob29rPWNocm9vdCAiJDEiIHNoIC1jICInCiAgICAgICAgICAgLiAoCiAgICAgICAgICAgICBqb2luICcgJiYgJywK' | base64 -d | patch /usr/bin/debrebuild",
```
msuozzo, line 154: I think the raw patch itself should be in code here and convert it to base64 inline. you could do that in the template itself or add a var somewhere to store the blob. also, please reduce the context lines from 3 before/after to maybe 1?

msuozzo, line 154: i'm vaguely concerning with how this interacts with the floating version of debrebuild. when that script changes, what should the behavior of this tool be? should there be more constraints on the debrebuild version? thoughts?

wbxyz: Yeah I was thinking about that. Specified a specific version. To avoid double downloading it I needed to specify the same version in the Needs section of debrebuild tool.

pkg/rebuild/debian/strategy.go
```diff
@@ -137,6 +147,13 @@ var toolkit = []*flow.Tool{
 			Needs: []string{},
 		}},
 	},
+	{
+		Name: "debian/deps/patch-debrebuild",
+		Steps: []flow.Step{{
+			Runs:  fmt.Sprintf("echo '%s' | base64 -d | patch /usr/bin/debrebuild", debrebuildSleepPatch),
```
msuozzo, line 153: could we inline this? actually seems like it would be pretty clear to do so.

msuozzo (approved):

wbxyz:
Hi josch! Thanks for following up.

After our conversation at reproducible builds I switched our system over to sbuild+unshare (#943) which seems to have better results as well.

Totally agree about "sleep X", we weren't happy with that solution either. I'm very happy we were able to remove it.

Unfortunately I can't remember the exact packages that we saw this behavior on. At the time we couldn't rule out our own execution environment as part of the issue, which is part of why we didn't file this as an upstream issue. If I see anything else come up I'll be sure to reach out and file a bug if appropriate.

PR #348, opened 2025-02-13
Omit redundant pypi build dir

msuozzo:


pkg/rebuild/pypi/strategy_test.go
```diff
@@ -72,6 +72,21 @@ export PIP_INDEX_URL=http://pypi:2006-01-02T03:04:05Z@orange
 				OutputPath: "dist/the_artifact",
 			},
 		},
+		{
+			"no dir",
```
wbxyz, line 76: nit: Convention in our codebase seems to be PascalCase

wbxyz (approved):

pkg/rebuild/pypi/strategy_test.go
wbxyz: At least across the strategy files, looks like it varies widely outside those examples. Some are snake case, some are capitalized words with spaces, some are lowercase with spaces.

Mainly, the other tests in this file use PascalCase.

PR #349, opened 2025-02-19
Fix manifest line wrapping

msuozzo:


pkg/archive/manifest.go
```diff
@@ -186,45 +186,38 @@ func WriteManifest(w io.Writer, m *Manifest) error {
...
+	base := len(name) + len(sep)
 	for len(remaining) > 72 {
 		// Find last space before 72 bytes
 		splitIdx := 71
-		for splitIdx > 0 && remaining[splitIdx] != ' ' {
+		for splitIdx > base && remaining[splitIdx] != ' ' {
 			splitIdx--
 		}
-		if splitIdx == 0 {
+		if splitIdx == base {
```
wbxyz, line 207: Is the space in the separator a valid split? If so, should this check be `splitIdx == len(name)`? Might need a +/-1 for correctness.

wbxyz (approved):

pkg/archive/manifest.go
msuozzo: i don't believe it is a valid split and this PR is really meant to address that (previously, it was breaking at that initial space). given the tests, I don't believe there are off-by-one issues.

PR #353, opened 2025-02-21
Fix timewarp pypi support

msuozzo:


internal/timewarp/timewarp_test.go
```diff
@@ -60,6 +60,11 @@ func TestHandler_ServeHTTP(t *testing.T) {
 						},
 					},
 				},
+				URLValidator: func(expected, actual string) {
```
loosebazooka, line 63: Could you not extract this out somwhere as a utility?

internal/timewarp/timewarp.go
```diff
@@ -295,7 +295,7 @@ func timeWarpPyPIProjectRequest(client httpx.BasicClient, obj map[string]any, at
 				if t.Before(at.Add(time.Second)) {
 					pastFiles = append(pastFiles, file)
 				}
-				if t.Before(firstSeen) {
+				if t.Before(firstSeen) || firstSeen.IsZero() {
```
loosebazooka, line 298: I don't seem to see where this is being tested?

msuozzo: it is tested in the version part of the URL that gets requested

internal/timewarp/timewarp_test.go
msuozzo: definitely could do so. it's been an easy copy-paste but it's being used across more of our tests. i can add this in a follow-on.

internal/timewarp/timewarp.go
msuozzo: to clarify: firstSeen will never update in the loop if this isn't here so the block below was using the default behavior of omitting latestVersion (which, for our purposes, has the same effect of fetching the actual latest version). added another test to exercise that default path, as well.

loosebazooka (approved):

PR #357, opened 2025-02-21
Adjust Manifest naming and visibility

msuozzo:


loosebazooka (commented):
lgtm, but needs prev PR, I think this could've been an independent PR

loosebazooka (approved):

PR #360, opened 2025-02-21
feat: delete files containing git properties

algomaster99:
Reference: https://github.com/git-commit-id/git-commit-id-maven-plugin/issues/825

The files `git.json` and `git. properties` are generated by https://github.com/git-commit-id/git-commit-id-maven-plugin. We delete these files in the jar file to stabilize. We assume default configuration that the files are either called `git.json` or `git.properties`.

### DIFF
--- a/pkg/archive/jar.go
+++ b/pkg/archive/jar.go
@@ -175,3 +175,15 @@ func splitPreservingQuotes(s string, sep rune) []string {
 	}
 	return result
 }
+
+var StableGitProperties = ZipArchiveStabilizer{
+	Name: "jar-git-properties",
+	Func: func(mr *MutableZipReader) {
+		for _, mf := range mr.File {
+			if strings.HasSuffix(mf.Name, "git.properties") || strings.HasSuffix(mf.Name, "git.json") {
+				mr.DeleteFile(mf.Name)
+			}
+		}
+
+	},
+}
--- a/pkg/archive/jar.go
+++ b/pkg/archive/jar.go
@@ -180,7 +180,17 @@ var StableGitProperties = ZipArchiveStabilizer{
 	Name: "jar-git-properties",
 	Func: func(mr *MutableZipReader) {
 		for _, mf := range mr.File {
+			// These files contain git properties set by git-commit-id-maven-plugin.
+			// They contain many unreproducible attributes as documented
+			// [here](https://github.com/git-commit-id/git-commit-id-maven-plugin/issues/825).
+
+			// Only JSON and properties file formats are available as documented
+			// [here](https://github.com/git-commit-id/git-commit-id-maven-plugin/blob/95d616fc7e16018deff3f17e2d03a4b217e55294/src/main/java/pl/project13/maven/git/GitCommitIdMojo.java#L454).
+			// By default, these file are created in ${project.build.outputDirectory} and are hence at the root of the jar.
+			// We assume that the file name is 'git' as this is the default value for the plugin.
+			// However, the plugin allows customizing the file name.
 			if strings.HasSuffix(mf.Name, "git.properties") || strings.HasSuffix(mf.Name, "git.json") {
+				// We remove these files to ensure reproducibility.
 				mr.DeleteFile(mf.Name)
 			}
 		}
--- a/pkg/archive/jar.go
+++ b/pkg/archive/jar.go
@@ -177,7 +177,6 @@ func splitPreservingQuotes(s string, sep rune) []string {
 	return result
 }
 
-var StableGitProperties = ZipArchiveStabilizer{
 var StableGitProperties = ZipEntryStabilizer{
 	Name: "jar-git-properties",
 	Func: func(zf *MutableZipFile) {
--- a/pkg/archive/jar.go
+++ b/pkg/archive/jar.go
@@ -178,23 +178,25 @@ func splitPreservingQuotes(s string, sep rune) []string {
 }
 
 var StableGitProperties = ZipArchiveStabilizer{
+var StableGitProperties = ZipEntryStabilizer{
 	Name: "jar-git-properties",
-	Func: func(mr *MutableZipReader) {
-		for _, mf := range mr.File {
-			// These files contain git properties set by git-commit-id-maven-plugin.
-			// They contain many unreproducible attributes as documented
-			// [here](https://github.com/git-commit-id/git-commit-id-maven-plugin/issues/825).
-
-			// Only JSON and properties file formats are available as documented
-			// [here](https://github.com/git-commit-id/git-commit-id-maven-plugin/blob/95d616fc7e16018deff3f17e2d03a4b217e55294/src/main/java/pl/project13/maven/git/GitCommitIdMojo.java#L454).
-			// By default, these file are created in ${project.build.outputDirectory} and are hence at the root of the jar.
-			// We assume that the file name is 'git' as this is the default value for the plugin.
-			// However, the plugin allows customizing the file name.
-			gitRegex := regexp.MustCompile(`\bgit\.(json|properties)$`)
-			if gitRegex.MatchString(mf.Name) {
-				mr.DeleteFile(mf.Name)
-			}
+	Func: func(zf *MutableZipFile) {
+		// These files contain git properties set by git-commit-id-maven-plugin.
+		// They contain many unreproducible attributes as documented
+		// [here](https://github.com/git-commit-id/git-commit-id-maven-plugin/issues/825).
+		// Only JSON and properties file formats are available as documented
+		// [here](https://github.com/git-commit-id/git-commit-id-maven-plugin/blob/95d616fc7e16018deff3f17e2d03a4b217e55294/src/main/java/pl/project13/maven/git/GitCommitIdMojo.java#L454).
+		// By default, these file are created in ${project.build.outputDirectory} and are hence at the root of the jar.
+		// We assume that the file name is 'git' as this is the default value for the plugin.
+		// We don't handle the case where the file name is changed by the user.
+		gitJsonRegex := regexp.MustCompile(`\bgit\.json$`)
+		if gitJsonRegex.MatchString(zf.Name) {
+			zf.SetContent([]byte("{}"))
 		}
 
+		gitPropertiesRegex := regexp.MustCompile(`\bgit\.properties$`)
+		if gitPropertiesRegex.MatchString(zf.Name) {
+			zf.SetContent([]byte{})
+		}
 	},
 }
--- a/pkg/archive/jar.go
+++ b/pkg/archive/jar.go
@@ -5,7 +5,7 @@ package archive
 
 import (
 	"bytes"
-	"regexp"
+	"path"
 	"sort"
 	"strings"
 )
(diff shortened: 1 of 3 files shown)


algomaster99:
@msuozzo ping for review

pkg/archive/jar.go
```diff
@@ -105,3 +105,25 @@ var StableJAROrderOfAttributeValues = ZipEntryStabilizer{
...
+			// These files contain git properties set by git-commit-id-maven-plugin.
+			// They contain many unreproducible attributes as documented
+			// [here](https://github.com/git-commit-id/git-commit-id-maven-plugin/issues/825).
+
+			// Only JSON and properties file formats are available as documented
+			// [here](https://github.com/git-commit-id/git-commit-id-maven-plugin/blob/95d616fc7e16018deff3f17e2d03a4b217e55294/src/main/java/pl/project13/maven/git/GitCommitIdMojo.java#L454).
+			// By default, these file are created in ${project.build.outputDirectory} and are hence at the root of the jar.
+			// We assume that the file name is 'git' as this is the default value for the plugin.
+			// However, the plugin allows customizing the file name.
+			if strings.HasSuffix(mf.Name, "git.properties") || strings.HasSuffix(mf.Name, "git.json") {
```
msuozzo, line 122: do we want to add a / separator here to avoid accidentally omitting a file like `plugit.json` accidentally?

pkg/archive/jar.go
```diff
@@ -105,3 +105,25 @@ var StableJAROrderOfAttributeValues = ZipEntryStabilizer{
...
+				// We remove these files to ensure reproducibility.
+				mr.DeleteFile(mf.Name)
```
msuozzo, line 124: alternatively, we could set the content to empty. that would ensure the file is checked for presence but not equality. that would also enable us to change this to a zip file stabilizer instead of a full archive stabilizer.

that said, I'm not sure an empty file is valid json which i could imagine causing issues.

what do you think?

pkg/archive/jar_test.go
```diff
@@ -312,3 +312,151 @@ func getSeparatedValues(attributeValue string) []string {
...
+		{
+			test: "delete_git_json",
+			input: []*ZipEntry{
+				{
+					&zip.FileHeader{Name: "META-INF/MANIFEST.MF"},
+					[]byte("Built-By: aman\r\n\r\n"),
+				},
+				{
+					&zip.FileHeader{Name: "git.json"},
+					[]byte("{\n" +
```
msuozzo, line 350: we have "textwrap.Dedent" elsewhere in the repo which should make this easier to define with formatting inline.

pkg/archive/jar_test.go
```diff
@@ -312,3 +312,151 @@ func getSeparatedValues(attributeValue string) []string {
...
+		{
+			test: "delete_git_properties",
+			input: []*ZipEntry{
+				{
+					&zip.FileHeader{Name: "META-INF/MANIFEST.MF"},
+					[]byte("Built-By: aman\r\n\r\n"),
+				},
+				{
+					&zip.FileHeader{Name: "git.properties"},
+					[]byte("git.build.user.email=mannu.poski10@gmail.com\r\ngit.build.user.name=Aman Sharma\r\n\r\n"),
```
msuozzo, line 331: i'm not sure i'd personally want my email address in code like this :)

pkg/archive/jar.go
```diff
@@ -105,3 +105,25 @@ var StableJAROrderOfAttributeValues = ZipEntryStabilizer{
...
+		for _, mf := range mr.File {
+			// These files contain git properties set by git-commit-id-maven-plugin.
+			// They contain many unreproducible attributes as documented
+			// [here](https://github.com/git-commit-id/git-commit-id-maven-plugin/issues/825).
+
+			// Only JSON and properties file formats are available as documented
+			// [here](https://github.com/git-commit-id/git-commit-id-maven-plugin/blob/95d616fc7e16018deff3f17e2d03a4b217e55294/src/main/java/pl/project13/maven/git/GitCommitIdMojo.java#L454).
+			// By default, these file are created in ${project.build.outputDirectory} and are hence at the root of the jar.
+			// We assume that the file name is 'git' as this is the default value for the plugin.
+			// However, the plugin allows customizing the file name.
```
msuozzo, line 121: good detail. have you seen anyone customize this in the wild?

pkg/archive/jar.go
algomaster99: I instead added a word boundary `\b` as the default location in the jar file is `git.json` and a separator would not match there, but a word boundary will match both.

pkg/archive/jar_test.go
algomaster99: Makes sense when I think about it :laughing:

pkg/archive/jar.go
algomaster99: Yes, I saw it [here](https://github.com/jrivard/xmlchai/blob/c84858ee4bae14c0bd16691688fa764f8b4f4ef6/pom.xml#L126). This project is part of Reproducible Central dataset so it was included in my analysis.

pkg/archive/jar.go
algomaster99: > that said, I'm not sure an empty file is valid json which i could imagine causing issues.

I looked it up and it seems [it is complicated](https://stackoverflow.com/a/22361229/11751642). I tried to give `diffoscope` and empty JSON and it is unable to use its JSON parser and hence falls back to `xxd`.

```
diffoscope a.json b.json # a.json is empty

--- a.json
+++ b.json
@@ -1 +1 @@
-00000000: 0a                                       .
+00000000: 7b22 6222 3a20 317d 0a                   {"b": 1}.
```

I can try setting the content to `{}`. That should be respected by any parser.

algomaster99:
It seems that there is a convention to group import declarations. However, `gofmt` did not take care of [`1946914` (#360)](https://github.com/google/oss-rebuild/pull/360/commits/1946914c172c071f0fa86e216eb4a3cf0039e7f7) as I had to do it manually.

algomaster99:
By the way, we will also have to declare the new stabilizers we write under [`AllZipStabilizers`](https://github.com/google/oss-rebuild/blob/11354d07c966e894687fe42bd8da5361a55662a4/pkg/archive/zip.go#L126-L134), right?

msuozzo:
> It seems that there is a convention to group import declarations. However, gofmt did not take care of [1946914 (#360)](https://github.com/google/oss-rebuild/pull/360/commits/1946914c172c071f0fa86e216eb4a3cf0039e7f7) as I had to do it manually.

weird. wouldn't worry too much about it as long as gofmt isn't actively complaining

> By the way, we will also have to declare the new stabilizers we write under [AllZipStabilizers](https://github.com/google/oss-rebuild/blob/11354d07c966e894687fe42bd8da5361a55662a4/pkg/archive/zip.go#L126-L134), right?

potentially? alternatively, we might add a separate JAR-specific top-level constant that we also expose in the set of all stabilizers in the CLI. Since it's not specific to the zip format but rather the JAR format, it might be worth keeping them separate.

pkg/archive/jar.go
msuozzo: yeah I think this comes down to the discussion in the other thread so let's close this one out.

pkg/archive/jar.go
algomaster99: We could try doing so. Our parser should look for plugins whose artifact ID is `git-commit-id-maven-plugin` in all `**/*/pom.xml`. This could be declared under `plugins` or under `profiles.plugins`. `profiles` activate a subset of plugins based on the flags provided to the command. Once we do that, we can check its configuration. There is also a [go library](https://github.com/vifraa/gopom) to parse POMs but does not seem very active. But a simple XML parser could also do the job.

algomaster99:
> wouldn't worry too much about it as long as gofmt isn't actively complaining

:+1: 

> potentially? alternatively, we might add a separate JAR-specific top-level constant

Cool. We can take this up in a separate PR.

pkg/archive/jar.go
algomaster99: I was trying to give a go at implementing this, but it seems if we parse POM, we will have to parse the properties as well. Some are declared and some are built-ins so it just seems a bit overkill at this point.
Another dimension to this is shading which can perhaps add multiple `git.json` files.

The current solution would delete any `git.json` and `git.properties` file and that seems okay to me.

msuozzo: Ah ok the part I was missing was that git.properties files form a sort of template for the eventual output.

In that case, I think I'm on board for emptying the files. I'm a bit leery about deleting them entirely but setting the json to {} and the properties to empty should be possible without doing any deeper properties parsing, correct?

I do think we should try to identify the paths/names we expect here as it's unfortunate the plugin allows such deep customization of the output locations.

Let me know if you have any concerns or you think the complexity of this policy outweighs the benefit vs deletion.

algomaster99: > think I'm on board for emptying the files. I'm a bit leery about deleting them entirely but setting the json to {}

I lost track actually. I already implemented that in https://github.com/google/oss-rebuild/pull/360/commits/a460c4e415ab6bc1c45c5636cd2efc202dd41775.

> I do think we should try to identify the paths/names we expect here as it's unfortunate the plugin allows such deep customization of the output locations.

We would have to look for plugin in the POM file
```
<plugin>
      <groupId>pl.project13.maven</groupId>
      <artifactId>git-commit-id-plugin</artifactId>
      <version>4.9.10</version>
</plugin>
``` 
with configuration
```
<configuration>
...
<generateGitPropertiesFilename>${project.build.outputDirectory}/classes/git.json</generateGitPropertiesFilename
...
</configuration>
```
If we look for `generateGitPropertiesFilename`, we should be able to get all paths to the git metadata files. However, the problem is with parsing the properties within `${}` as they could be either declared in the same POM file or some other POM file (parent POMs). Some properties are also built-in like `project.build.outputDirectory` whose value is `target`.
I can give this is a go. It is mostly POM parsing.

pkg/archive/jar.go
```diff
@@ -170,3 +171,26 @@ func splitPreservingQuotes(s string, sep rune) []string {
...
+		// [here](https://github.com/git-commit-id/git-commit-id-maven-plugin/blob/95d616fc7e16018deff3f17e2d03a4b217e55294/src/main/java/pl/project13/maven/git/GitCommitIdMojo.java#L454).
+		// By default, these file are created in ${project.build.outputDirectory} and are hence at the root of the jar.
+		// We assume that the file name is 'git' as this is the default value for the plugin.
+		// We don't handle the case where the file name is changed by the user.
+		gitJsonRegex := regexp.MustCompile(`\bgit\.json$`)
+		if gitJsonRegex.MatchString(zf.Name) {
+			zf.SetContent([]byte("{}"))
+		}
+
+		gitPropertiesRegex := regexp.MustCompile(`\bgit\.properties$`)
```
msuozzo, line 191: hmm this would match `foo-git.properties` so I think we'd want something more like `path.Base(zf.Name) == "git.properties"`.

same above with git.json.

pkg/archive/jar.go
msuozzo: Makes sense. I think we can survive with this implementation as-is. it's imperfect but the shortcomings are well-documented and we can work towards a more rigorous implementation in the future.

msuozzo (approved):
Sorry for the delay! I'm cool to merge this as-is provided you're happy with the state of things.

Thanks for the contribution!

pkg/archive/jar.go
algomaster99: Sorry for the delay from my end. I will find time to work on this soon and submit a separate PR with improvements.

pkg/archive/jar.go
algomaster99: Yes, I agree. Good catch :)
Fixed in 0044c79

msuozzo: Looks good!

algomaster99:
Resolved conflicts here too.

pkg/archive/jar.go
algomaster99: > If we look for generateGitPropertiesFilename, we should be able to get all paths to the git metadata files. However, the problem is with parsing the properties within ${} as they could be either declared in the same POM file or some other POM file (parent POMs). Some properties are also built-in like project.build.outputDirectory whose value is target.
I can give this is a go. It is mostly POM parsing.

Btw, I have implemented this and now we know the exact filepath for git properties file. I closed https://github.com/algomaster99/oss-rebuild/pull/1 as there were too many merge conflicts so might be hard to review.

PR #362, opened 2025-02-27
chore: Add to list of MANIFEST.MF skip keys

theopolis:
This does not remove any keys, only adds additional skip keys. There is a slight reordering to achieve alpha-order of the slice.

In addition to these keys there are also a suite of `Plugin-Build` and `Plugin-Built` keys that are also metadata and not always stable for comparison. And then there are many `X-` keys that also container variances. Unsure how you would like to handle these or if there is any authoritative reference.


wbxyz:
This LGTM, but I know Matthew's been working with you on Maven things so I'll let him give the final review.

theopolis:
This might benefit from a test that assures the list remains in alpha-order for simpler development.

I also think the delete check could implement an `EqualFold` in the logic.

msuozzo (approved):

PR #363, opened 2025-03-03
refactor: add top level jar stabilizer

algomaster99:
Reference: https://github.com/google/oss-rebuild/pull/360#issuecomment-2679027236

### DIFF
--- a/pkg/archive/archive.go
+++ b/pkg/archive/archive.go
@@ -13,7 +13,7 @@ import (
 	"github.com/pkg/errors"
 )
 
-var AllStabilizers = slices.Concat(AllZipStabilizers, AllTarStabilizers, AllGzipStabilizers)
+var AllStabilizers = slices.Concat(AllZipStabilizers, AllTarStabilizers, AllGzipStabilizers, AllJarStabilizers)
 
 // Stabilize selects and applies the default stabilization routine for the given archive format.
 func Stabilize(dst io.Writer, src io.Reader, f Format) error {
--- a/pkg/archive/jar.go
+++ b/pkg/archive/jar.go
@@ -9,6 +9,11 @@ import (
 	"strings"
 )
 
+var AllJarStabilizers []any = []any{
+	StableJARBuildMetadata,
+	StableJAROrderOfAttributeValues,
+}
+
 var StableJARBuildMetadata = ZipEntryStabilizer{
 	Name: "jar-build-metadata",
 	Func: func(zf *MutableZipFile) {


algomaster99:
@msuozzo

msuozzo (approved):

algomaster99:
@msuozzo can we merge this? or is there something more to be done here?

I had thoughts about writing a discoverer of stabilizers using `reflect` so that we don't have to register stabilizers every time we write one. However, that might entail some performance losses.

msuozzo:
> @msuozzo can we merge this? or is there something more to be done here?
> 
> I had thoughts about writing a discoverer of stabilizers using `reflect` so that we don't have to register stabilizers every time we write one. However, that might entail some performance losses.

Yeah for now we can avoid reflection.

pkg/archive/jar.go
msuozzo, line 1: Would you be able to rebase to head? I think we changed `any` to be a dedicated typedef so this may need to use that new type.

msuozzo (approved):

algomaster99:
Done @msuozzo !

algomaster99:
Also added `StableGitProperties` stabilizer :)

pkg/archive/jar.go
```diff
@@ -10,6 +10,12 @@ import (
 	"strings"
 )
 
+var AllJarStabilizers []Stabilizer = []Stabilizer{
```
msuozzo, line 13: nit: you should be able to omit the type on the var which will prevent declaration stutter.

pkg/archive/jar.go
```diff
@@ -10,6 +10,12 @@ import (
...
+	StableJARBuildMetadata,
+	StableJAROrderOfAttributeValues,
+	StableGitProperties,
```
msuozzo, line 16: ah this should probably be StableJARGitProperties to keep with the naming scheme here. can be done in a follow-on change tho.

pkg/archive/archive.go
```diff
@@ -13,7 +13,7 @@ import (
 	"github.com/pkg/errors"
 )
 
-var AllStabilizers = slices.Concat(AllZipStabilizers, AllTarStabilizers, AllGzipStabilizers)
+var AllStabilizers = slices.Concat(AllZipStabilizers, AllTarStabilizers, AllGzipStabilizers, AllJarStabilizers)
```
msuozzo, line 16: I think i'm feeling some hesitance adding the JAR stabilizers here because these would be the first stabilizers to be definitively ecosystem-specific and the existing uses aren't yet able to differentiate.

The simple solution would be to add ecosystem-mapping somewhere (probably at a higher level than the archive package) that could be used from our provenance generation logic. For the stabilizer CLI, we could infer the ecosystem based off of the file extension and, when ambiguous, require the user explicitly pass in an ecosystem flag.

Apologies for creating more work but I think we're asking for unexpected consequences without that protection.

Do you feel up to making those changes? otherwise, I can take a crack at them and unblock you here.

msuozzo: okay I went ahead and fixed this in #418. I think it should be safer to submit this once that's in.

algomaster99: That's a good point. I agree with you. I will take a look at that PR. Thanks!

msuozzo: It's now merged so this should be safe to rebase onto. You can also make use of the new AllJarStabilizers constant in that new ecosystem-specific logic.

algomaster99: Sounds great. I will take a look soon.

pkg/archive/jar.go
algomaster99: Fixed [`9b372b2` (#363)](https://github.com/google/oss-rebuild/pull/363/commits/9b372b27ca3c3131f91304265c982af12f710f2b)

pkg/archive/jar.go
algomaster99: Removed it. My Java habbits seem to seep in here :sweat_smile:

pkg/archive/jar_test.go
```diff
@@ -180,7 +180,7 @@ func TestStableJARBuildMetadata(t *testing.T) {
 	}
 }
 
-func TestStableOrderOfAttributeValues(t *testing.T) {
+func TestStableJAROrderOfAttributeValues(t *testing.T) {
```
algomaster99, line 183: @msuozzo note that I also changed the name of the test to follow naming conventions.

algomaster99:
This is ready to for your review again :)

pkg/archive/jar_test.go
msuozzo: yep sounds good!

pkg/rebuild/stability/stabilizer.go
```diff
@@ -29,7 +29,7 @@ func StabilizersForTarget(t rebuild.Target) ([]archive.Stabilizer, error) {
 	switch t.Ecosystem {
 	case rebuild.Maven:
 		if format == archive.ZipFormat {
-			stabilizers = append(stabilizers, archive.StableJARBuildMetadata, archive.StableJAROrderOfAttributeValues, archive.StableGitProperties)
+			stabilizers = append(stabilizers, archive.AllJarStabilizers)
```
msuozzo, line 32: nit: I think this should have the spread operator here:

`append(stabilizers, archive.AllJarStabilizers...)`

pkg/archive/jar.go
```diff
@@ -177,7 +183,7 @@ func splitPreservingQuotes(s string, sep rune) []string {
 	return result
 }
 
-var StableGitProperties = ZipEntryStabilizer{
+var StableJARGitProperties = ZipEntryStabilizer{
```
msuozzo, line 186: in general, I'd prefer to have this in a separate PR since it's separate from the description of this one. if you'd like to leave it in now, I think that's fine but just some level-setting for the future.

algomaster99: I thought about it too. I was perhaps just lazy to make another PR for this small change :sweat_smile: I will revert it and then send another PR for it :)

pkg/rebuild/stability/stabilizer.go
algomaster99: This is a nitpick as it just makes the code clearer right? Otherwise, `go` seems to implcity append elements of array instead of appending the array (this would anyway fail type checking).

algomaster99:
@msuozzo PR seems clean now :)

pkg/rebuild/stability/stabilizer.go
msuozzo: Jeez I didn't realize this but because Stabilizer is an alias of `any`, `[]Stabilizer` will accept the slice as an element 🤦 

```
type Foo any

func main() {
	s := []Foo{0}
	fmt.Printf("%v\n", s)
	fmt.Printf("%v\n", append(s, 1))
	fmt.Printf("%v\n", append(s, []Foo{1}))
	fmt.Printf("%v\n", append(s, []Foo{1}...))
}
```

prints...

```
[0]
[0 1]
[0 [1]]
[0 1]
```

so we really can't rely on types here, apparently :(

msuozzo (approved):

pkg/rebuild/stability/stabilizer.go
msuozzo: filed an issue to track.

PR #367, opened 2025-03-04
Add the assets format of get-results

wbxyz:


tools/ctl/ctl.go
```diff
@@ -218,6 +224,30 @@ var getResults = &cobra.Command{
...
+		case "assets":
+			regclient := http.DefaultClient
+			mux := rebuild.RegistryMux{
+				Debian:   debianreg.HTTPRegistry{Client: regclient},
+				CratesIO: cratesreg.HTTPRegistry{Client: regclient},
+				NPM:      npmreg.HTTPRegistry{Client: regclient},
+				PyPI:     pypireg.HTTPRegistry{Client: regclient},
+			}
+			butler := localfiles.NewButler(*metadataBucket, *logsBucket, *debugStorage, mux)
+			atype := rebuild.AssetType(*assetType)
```
msuozzo, line 236: is there any validation we can/should do here?

tools/ctl/ctl.go
```diff
@@ -218,6 +224,30 @@ var getResults = &cobra.Command{
...
+			for _, r := range rebuilds {
+				t := rebuild.Target{
+					Ecosystem: rebuild.Ecosystem(r.Ecosystem),
+					Package:   r.Package,
+					Version:   r.Version,
+					Artifact:  r.Artifact,
+				}
+				path, err := butler.Fetch(cmd.Context(), *runFlag, r.WasSmoketest(), atype.For(t))
```
msuozzo, line 244: nit: I'd pull out cmd.Context() to outside the loop

tools/ctl/ctl.go
```diff
@@ -218,6 +224,30 @@ var getResults = &cobra.Command{
...
+			mux := rebuild.RegistryMux{
+				Debian:   debianreg.HTTPRegistry{Client: regclient},
+				CratesIO: cratesreg.HTTPRegistry{Client: regclient},
+				NPM:      npmreg.HTTPRegistry{Client: regclient},
+				PyPI:     pypireg.HTTPRegistry{Client: regclient},
+			}
+			butler := localfiles.NewButler(*metadataBucket, *logsBucket, *debugStorage, mux)
+			atype := rebuild.AssetType(*assetType)
+			for _, r := range rebuilds {
+				t := rebuild.Target{
```
msuozzo, line 238: nit: r.Target()

msuozzo (approved):

tools/ctl/ctl.go
wbxyz: I don't think there's an easy way to check without maintaining a list of "all assets". The assetlocator only supports a handful of these asset times so far, it will give an error if you put an invalid asset.

tools/ctl/ctl.go
wbxyz: Yeah good call.

tools/ctl/ctl.go
wbxyz: Ah ha! Forgot about that.

PR #372, opened 2025-03-06
Fix builddef to use StrategyOneOf

msuozzo:


pkg/builddef/set.go
```diff
@@ -30,21 +32,21 @@ func NewFilesystemBuildDefinitionSet(fs billy.Filesystem) *FilesystemBuildDefini
 	return &FilesystemBuildDefinitionSet{fs: fs}
 }
 
-func (s *FilesystemBuildDefinitionSet) Get(ctx context.Context, t rebuild.Target) (rebuild.Strategy, error) {
+func (s *FilesystemBuildDefinitionSet) Get(ctx context.Context, t rebuild.Target) (schema.StrategyOneOf, error) {
```
wbxyz, line 35: nit: Want to make this a pointer to schema.StrategyOneOf?

It has the potential to be pretty large.

wbxyz (approved):

pkg/builddef/set.go
msuozzo: StrategyOneOf is just pointers so i think it should be fine

PR #375, opened 2025-03-06
Support user-provided stabilizer config

msuozzo:


internal/verifier/summary.go
```diff
@@ -36,7 +36,8 @@ func SummarizeArtifacts(ctx context.Context, metadata rebuild.LocatableAssetStor
 		return rb, up, errors.Wrap(err, "reading artifact")
 	}
 	defer checkClose(r)
-	err = archive.Stabilize(rb.StabilizedHash, io.TeeReader(r, rb.Hash), t.ArchiveType())
+	opts := archive.StabilizeOpts{Stabilizers: append(archive.AllStabilizers, stabilizers...)}
```
wbxyz, line 39: Does this modify archive.AllStabalizers?

wbxyz (approved):

internal/api/apiservice/rebuild.go
```diff
@@ -224,6 +228,19 @@ func buildAndAttest(ctx context.Context, deps *RebuildPackageDeps, mux rebuild.R
 	if err != nil {
 		return errors.Wrap(err, "creating rebuild store")
 	}
+	var stabilizers []archive.Stabilizer
+	if entry != nil {
+		stabilizers, err = archive.CreateCustomStabilizers(entry.StabilizerConfig, t.ArchiveType())
```
wbxyz, line 233: Should these be added to the attestation?

internal/verifier/summary.go
wbxyz: We talked about it briefly yesterday, and you mentioned you don't think it's a concern: even though AllStabilizers and the new slice possibly share a reference to the same underlying array, they have different size ints so they have a different "view" of that array.

Thinking about it more, my concern is
1. A user of StabilizeOpts could reasonably start deleting stabilizers from the list (if they don't make sense in the context) which could potentially modify AllStabilizer's underlying array I think?
2. If two goroutines on the same worker are handling separate requests, the could clobber each other's write into the underlying array.

I wish there were an immutable slice type, or an easier way to force the underlying array to be copied not referenced.

msuozzo: done

internal/api/apiservice/rebuild.go
msuozzo: done. now serialized into attestation.

wbxyz (approved):

internal/api/apiservice/rebuild.go
wbxyz: Cool, thanks!

Issue #228, opened 2024-12-19
Documented diff support


msuozzo:
Enable reporting of non-exact rebuilds by adding support for documenting expected, benign differences (like local files or docs formatting changes) in build definitions while maintaining our existing trust model. Our end state should look something like this:

- Build definitions can specify expected benign differences
- Stabilizer handles documented diffs
- ArtifactEquivalence includes diff specifications

### Tasks
- [ ] Add stabilizer config to ArtifactEquivalence attestation
- [ ] Implement diff support as a Stabilizer
  - Define diff config in opts
  - Add diff support to stabilizer logic
- [ ] Add diff specification to build definition
  - Support full file exclusion and per-file diff exclusion
  - Make a higher-level build def construct aggregating StrategyOneOf and Diff config
- [ ] Define review guidelines for permissible diffs

### Future
Looking forward, we're thinking of publishing "negative attestations" for packages we expect to be able to rebuild that fail to do so. This will provide externval visibility into potentially interesting rebuild variations.

wbxyz:
Related to the future work, we can sort out which packages we want to commit to supporting in #245 and we can discuss the negative attestations or RebuildFailure attestation type in #247

PR #378, opened 2025-03-11
Support wildcards in custom stabilizer paths

msuozzo:
UX improvement for #228


pkg/archive/custom.go
wbxyz, line 1: Want to add one glob pattern to each of these tests?

wbxyz (approved):

pkg/archive/custom.go
msuozzo: done

PR #379, opened 2025-03-11
Support multiple paths per stabilizer entry

msuozzo:
UX improvement for #228

### DIFF
--- a/pkg/archive/custom.go
+++ b/pkg/archive/custom.go
@@ -8,6 +8,8 @@ import (
 	"io"
 	"regexp"
 
+	"slices"
+
 	"github.com/google/oss-rebuild/internal/glob"
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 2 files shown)


pkg/archive/custom.go
```diff
@@ -95,14 +97,17 @@ func CreateCustomStabilizers(entries []CustomStabilizerEntry, format Format) ([]
 // - Pattern is a regex that accepts the golang RE2 syntax.
 // - Replace can define a substitution for the matched content.
 type ReplacePattern struct {
-	Path    string `yaml:"path"`
-	Pattern string `yaml:"pattern"`
-	Replace string `yaml:"replace"`
+	Paths   []string `yaml:"paths"`
```
wbxyz, line 100: nit: The docstring above could be updated to match field name, and maybe rephrased now that it's a slice?

Also the the one below.

pkg/archive/custom_test.go
wbxyz, line 1: nit: Add a test for each that has multiple Paths

wbxyz (approved):

pkg/archive/custom.go
msuozzo: done

pkg/archive/custom_test.go
msuozzo: done

PR #381, opened 2025-03-11
Include tetragon policy to track open_at

wbxyz:


pkg/rebuild/rebuild/rebuildremote.go
```diff
@@ -203,9 +222,11 @@ var standardBuildTpl = template.Must(
 				#!/usr/bin/env bash
 				set -eux
 				{{- if .UseSyscallMonitor}}
-				touch /workspace/tetragon.jsonl
-				echo '{{.SyscallPolicy}}' > /workspace/tetragon_policy.yaml
-				export TID=$(docker run --name=tetragon --detach --pid=host --cgroupns=host --privileged -v=/workspace/tetragon.jsonl:/workspace/tetragon.jsonl -v=/workspace/tetragon_policy.yaml:/workspace/tetragon_policy.yaml -v=/sys/kernel/btf/vmlinux:/var/lib/tetragon/btf quay.io/cilium/tetragon:v1.1.2 /usr/bin/tetragon --tracing-policy=/workspace/tetragon_policy.yaml --export-filename=/workspace/tetragon.jsonl)
+				mkdir /workspace/tetragon/
+				{{- range .SyscallPolicies}}
```
msuozzo, line 226: you can use the explicit range variables to get the index and value as `$var` values:

`{{ range $i, $policy := .SyscallPolicies}}`

pkg/rebuild/rebuild/rebuildremote.go
```diff
@@ -95,20 +94,40 @@ spec:
...
+	for _, policyYaml := range tetragonPoliciesYaml {
+		var data any
+		if err := yaml.Unmarshal([]byte(policyYaml), &data); err != nil {
+			log.Fatalf("Malformed tetragon policy: %v", err)
+		}
+		b, err := json.Marshal(data)
+		if err != nil {
+			log.Fatalf("Converting tetragon policy to json: %v", err)
+		}
+		tetragonPoliciesJSON = append(tetragonPoliciesJSON, string(b))
```
msuozzo, line 129: nit: let's also check that the serialized JSON doesn't have single quotes in it

msuozzo (approved):

pkg/rebuild/rebuild/rebuildremote.go
wbxyz: Cool, TIL

PR #382, opened 2025-03-11
Parse artifact in debian registry

wbxyz:


pkg/rebuild/debian/infer.go
```diff
@@ -93,19 +91,15 @@ func inferBuildInfo(t rebuild.Target) (rebuild.Strategy, error) {
...
-		}
-		arch := matches[versionRegex.SubexpIndex("arch")]
-		infoURL = debian.BuildInfoURL(name, version, arch)
-	} else {
-		return nil, fmt.Errorf("failed to parse artifact %s", t.Artifact)
+	a, err := debian.ParseDebianArtifact(t.Artifact)
+	if err != nil {
+		return nil, err
+	}
+	version := a.NonBinaryVersion
```
msuozzo, line 98: maybe a.DebianVersion or a.SourceVersion? i think we should be precise here (or, i guess, as precise as we _can be_ given the variety of debian packaging).

pkg/registry/debian/debian.go
```diff
@@ -20,6 +20,7 @@ var (
 	registryURL         = "https://deb.debian.org/debian/pool/"
 	buildinfoURL        = "https://buildinfos.debian.net/buildinfo-pool/"
 	binaryReleaseRegexp = regexp.MustCompile(`(\+b[\d\.]+)$`)
+	versionRegex        = regexp.MustCompile(`^(?P<name>[^_]+)_(?P<nonbinary_version>[^_+]+)(?P<binary_version>\+.*)?_(?P<arch>[^_]+)\.deb$`)
```
msuozzo, line 23: let's also update the "nonbinary_version" name here to be more precise/descriptive

pkg/rebuild/debian/infer.go
```diff
@@ -93,19 +91,15 @@ func inferBuildInfo(t rebuild.Target) (rebuild.Strategy, error) {
...
+	if strings.HasPrefix(a.BinaryVersion, "+deb") {
+		version += a.BinaryVersion
 	}
+	infoURL := debian.BuildInfoURL(name, version, a.Arch)
```
msuozzo, line 102: would be _really_ helpful to have a comment or two here explaining the relevant abstractions. notably, the difference between the package, artifact, and various debian-specific version IDs.

pkg/registry/debian/debian.go
```diff
@@ -147,6 +148,35 @@ func (r HTTPRegistry) DSC(ctx context.Context, component, name, version string)
...
+	if matches == nil {
+		return DebianArtifact{}, errors.Errorf("unexpected artifact name: %s", artifact)
+	}
+	a := DebianArtifact{
+		Name:             matches[versionRegex.SubexpIndex("name")],
+		NonBinaryVersion: matches[versionRegex.SubexpIndex("nonbinary_version")],
+		BinaryVersion:    matches[versionRegex.SubexpIndex("binary_version")],
+		Arch:             matches[versionRegex.SubexpIndex("arch")],
+	}
+	if strings.HasPrefix(a.BinaryVersion, "+deb") {
```
msuozzo, line 173: per offline discussion, this could potentially go in the regex

pkg/registry/debian/debian.go
```diff
@@ -147,6 +148,35 @@ func (r HTTPRegistry) DSC(ctx context.Context, component, name, version string)
 	return DSCURI, d, err
 }
 
+type DebianArtifact struct {
```
msuozzo, line 151: potential improvement here: descructure the full complexity of deb versions and then implement the logic to rematerialize them. i think this would be a great help in self-documenting our understanding of artifact naming and in calling out all the fun edge cases.

pkg/registry/debian/debian_test.go
```diff
@@ -301,6 +301,66 @@ RLpmHHG1JOVdOA==
...
+			expected: DebianArtifact{
+				Name:             "libacl1",
+				NonBinaryVersion: "2.3.2-2",
+				BinaryVersion:    "+b1",
+				Arch:             "amd64",
+			},
+			wantErr: false,
+		},
+		{
+			name:     "+deb version",
```
msuozzo, line 333: want to include a `~deb` case?

pkg/registry/debian/debian.go
```diff
@@ -17,11 +17,114 @@ import (
...
+	debRegex              = regexp.MustCompile(`^(?P<name>[^_]+)_(?P<version>[^_]+)_(?P<arch>[^_]+)\.deb$`)
+	nativeVersionRegex    = regexp.MustCompile(`^((?P<epoch>[0-9]+):)?(?P<upstream_version>[0-9][A-Za-z0-9\.\+\~]*)$`)
+	nonNativeVersionRegex = regexp.MustCompile(`^((?P<epoch>[0-9]+):)?(?P<upstream_version>[0-9][A-Za-z0-9\.\+\~\-]*)\-(?P<debian_revision>[A-Za-z0-9\+\.\~]+)$`) // upstream_version is only allowed to contain "-" if debian_revision is non-empty
+	binaryNMURegex        = regexp.MustCompile(`^(?P<base>.*)+b(?P<binaryNMU>[0-9]+)$`)
 )
 
+// Debian pakage versions are rather complex. See these docs for details
+// https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-version
+
+type FullVersion struct {
```
msuozzo, line 31: Maybe PackageVersion? Or ArtifactVersion?

pkg/registry/debian/debian.go
```diff
@@ -17,11 +17,114 @@ import (
...
 var (
-	registryURL         = "https://deb.debian.org/debian/pool/"
-	buildinfoURL        = "https://buildinfos.debian.net/buildinfo-pool/"
-	binaryReleaseRegexp = regexp.MustCompile(`(\+b[\d\.]+)$`)
+	registryURL           = "https://deb.debian.org/debian/pool/"
+	buildinfoURL          = "https://buildinfos.debian.net/buildinfo-pool/"
+	debRegex              = regexp.MustCompile(`^(?P<name>[^_]+)_(?P<version>[^_]+)_(?P<arch>[^_]+)\.deb$`)
+	nativeVersionRegex    = regexp.MustCompile(`^((?P<epoch>[0-9]+):)?(?P<upstream_version>[0-9][A-Za-z0-9\.\+\~]*)$`)
+	nonNativeVersionRegex = regexp.MustCompile(`^((?P<epoch>[0-9]+):)?(?P<upstream_version>[0-9][A-Za-z0-9\.\+\~\-]*)\-(?P<debian_revision>[A-Za-z0-9\+\.\~]+)$`) // upstream_version is only allowed to contain "-" if debian_revision is non-empty
+	binaryNMURegex        = regexp.MustCompile(`^(?P<base>.*)+b(?P<binaryNMU>[0-9]+)$`)
```
msuozzo, line 25: Might be useful to define this acronym and potentially link to docs since I think it's the first time it's referenced

pkg/rebuild/debian/infer.go
wbxyz: I nested the version details in a dedicated struct because it's complicated enough, and also because we use versions on their own.

I did my best matching names with the documentation but lmk if anything else stands out.

pkg/rebuild/debian/infer.go
wbxyz: The version is structured more, so it's a little more self-documenting. I added a comment explaining the choices for BuildInfoURL though.

pkg/registry/debian/debian.go
wbxyz: Yup, I made the regex more intelligent so this add-back isn't needed. Also broke up the regex into a few smaller and more specific regex patterns.

pkg/registry/debian/debian.go
wbxyz: Done. I haven't implemented many re-materialize things yet, but it's a lot more structured now. I also added a bunch of real examples with different edge cases to our test cases.

pkg/registry/debian/debian_test.go
wbxyz: Done.

pkg/registry/debian/debian.go
wbxyz: Done. The docs are the same link thats a few lines down here, and also in the DebianArtifact struct.

pkg/registry/debian/debian.go
wbxyz: PackageVersion is ambiguous because there's also the source package which is potentially different. ArtifactVersion sounds too precise because multiple artifacts are accurately referenced by a single version. Maybe BuildVersion to capture this is unique each time a build is executed? Idk FullVersion doesn't bother me much though.

wbxyz: In the docs (and .dsc) it's just called `Version` so maybe we should just go with that?

msuozzo: I just don't know what it's "full" of. I could get on board with "BuildVersion" and potentially "Version". but isn't the latter a method  on the registry in this package? idk maybe "Version" is fine..

pkg/registry/debian/debian.go
```diff
@@ -17,11 +17,125 @@ import (
...
+	}
+	return source
+}
+
+// Native returns true if a package is "native", meaning there is no distinction between Debian package releases and upstream releases.
+func (v *FullVersion) Native() bool {
+	return v.DebianRevision == ""
+}
+
+func (v *FullVersion) BinaryNMU() string {
```
msuozzo, line 97: wanna throw a comment here and explain the acronym inline?

pkg/registry/debian/debian.go
```diff
@@ -17,11 +17,125 @@ import (
...
+	}
+	return ""
+}
+
+// NonBinaryString converts this version into a string, minus any binary-only version identifiers.
+func (v *FullVersion) NonBinaryString() string {
+	return strings.TrimSuffix(v.String(), "+b"+v.BinaryNMU())
+}
+
+type DebianArtifact struct {
```
msuozzo, line 114: hmm I just realized this stutters with the package name. If you're going with "Version" now, "Artifact" or DebArtifact could be appropriate.

pkg/registry/debian/debian.go
```diff
@@ -17,11 +17,125 @@ import (
...
+	}
+	// If it's native and the upstream ends with \+b[0-9]+
+	if match := binaryNMURegex.FindStringSubmatch(v.DebianRevision); !v.Native() && match != nil {
+		return match[binaryNMURegex.SubexpIndex("binaryNMU")]
+	}
+	return ""
+}
+
+// NonBinaryString converts this version into a string, minus any binary-only version identifiers.
+func (v *FullVersion) NonBinaryString() string {
```
msuozzo, line 110: again I don't think this terminology works because it's describing the thing in the negative instead of describing it or naming it. what parts of the version does this include? is there some debian way of referring to this part of the name?

pkg/registry/debian/debian.go
```diff
@@ -17,11 +17,125 @@ import (
...
+		source = spl[len(spl)-1]
+	}
+	// Remove any binary-only version suffix.
+	if match := binaryNMURegex.FindStringSubmatch(source); match != nil {
+		source = match[binaryNMURegex.SubexpIndex("base")]
+	}
+	return source
+}
+
+// Native returns true if a package is "native", meaning there is no distinction between Debian package releases and upstream releases.
```
msuozzo, line 92: i'd rephrase this to "if a package uses the upstream versioning scheme directly"

alternatively, "lacks a separate debian version identifier"

pkg/registry/debian/debian.go
```diff
@@ -17,11 +17,125 @@ import (
...
+	}
+	upstream := v.Upstream
+	debianRevision := ""
+	if v.DebianRevision != "" {
+		debianRevision = "-" + v.DebianRevision
+	}
+	return epoch + upstream + debianRevision
+}
+
+// RealUpstream will return the true upstream, even if there's been a rollback denoted with +really.
```
msuozzo, line 75: should v.Upstream really contain these pieces to begin with?

pkg/registry/debian/debian.go
```diff
@@ -9,19 +9,135 @@ import (
...
+func (v *FullVersion) RealUpstream() string {
```
msuozzo, line 76: it seems like this description is missing the bit about the binary suffix. "real" also doesn't thrill me since the future re-releases are no less real.

pkg/registry/debian/debian.go
wbxyz: I don't want to document the acronym in multiple places, want me to move the comment from the regex down here?

pkg/registry/debian/debian.go
wbxyz: It felt wrong to have rebuild.Artifact and debian.Artifact, especially when we have the HTTPRegistry.Artifact() method which doesn't return a DebianArtifact.

What about ArtifactName?

pkg/registry/debian/debian.go
wbxyz: It's a combination of epoch, upstream_version, debian_revision, with any binary version stripped off the end of either the upstream_version, or the debian_revision.

We could maybe call it SourcePackageVersion but that's inferring origination information which seems like it's outside the domain of the registry library.

pkg/registry/debian/debian.go
wbxyz: I tried to stick to the official documentation of "native": https://www.debian.org/doc/debian-policy/ch-source.html#source-packages

pkg/registry/debian/debian.go
wbxyz: They're part of `upstream_version` described here https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-version

Regardless, we need to store them somewhere because otherwise we cant reconstruct the full version identifier.

pkg/registry/debian/debian.go
wbxyz: Added a comment about binary suffixes being removed.

I wanted to capture both rollbacks and not, and because debian uses "+really" as their terminology it seemed fitting.

pkg/registry/debian/debian.go
wbxyz: I was thinking it was "full" as in a "total" combination of the epoch, upstream_version, and debian_revision.

Thinking about it more, we might want to use this for DSC or source package versions in the future, in which case BuildVersion might not make sense.

Renamed to Version.

msuozzo: sg

pkg/registry/debian/debian.go
msuozzo: I think there remains an issue of clarity in using the same name for both things. do you have any ideas for how to differentiate them? we use "RawFoo" and "Foo" in some places but i'm not sure that applies here.

pkg/registry/debian/debian.go
msuozzo: I think my suggestions are not in conflict with that definition and, to my eye, are clearer. I'm concerned with the existing terminology because there very much is a distinction between the debian release and the upstream release and the upstream may ship binaries and source that are still rebuilt or modified for debian. semantically, it's the versioning scheme that's "native" here, not the release or package.

pkg/registry/debian/debian.go
msuozzo: I mean, can't you just expand the acronym to Binary Non-Maintainer Upload in the comment? you can send people to another description but this sort of foreign acronym makes code harder to understand (and is against golang style, to boot).

pkg/registry/debian/debian.go
msuozzo: Interesting. I _could_ see the registry knowing what the source package is but that's certainly a sort of inverted definition.

BinaryIndependentString?

pkg/registry/debian/debian.go
msuozzo: oh wait yeah Artifact isn't a good name, not least of all since it's not actually the artifact. I could get behind ArtifactName. I might lean towards using "ID" or "Identifier" since Name is used for a struct member.

wbxyz: I went with ArtifactIdentifier

pkg/registry/debian/debian.go
wbxyz: Expanded the acronym on the method, and added a docstring without defining *what* a binary-only non maintainer upload is. If people really care their can look at the docs linked or the regex above.

pkg/registry/debian/debian.go
wbxyz: BinaryIndependentString works for me, updated.

pkg/registry/debian/debian.go
wbxyz: After offline discussion, we landed on renaming this as "RollbackBase" which is only populated when the version is a rollback.

pkg/registry/debian/debian.go
wbxyz: Yeah I can see how this reads as more authoritative than our code should be. I guess the debian docs are understandably more strongly worded than our codebase is.

msuozzo (approved):

PR #383, opened 2025-03-11
Read artifacts from the snapshot service

wbxyz:


pkg/registry/debian/debian.go
```diff
@@ -177,9 +180,77 @@ func ParseDebianArtifact(artifact string) (DebianArtifact, error) {
...
+	// Example series of urls to follow:
+	// https://snapshot.debian.org/mr/binary/acl/
+	// https://snapshot.debian.org/mr/package/acl/2.3.2-2/binfiles/libacl1/2.3.2-2+b1?fileinfo=1
+	// https://snapshot.debian.org/file/53f2b0612c8ed8a60970f9a206ae65eb84681f6e
+	a, err := ParseDebianArtifact(artifact)
+	if err != nil {
+		return "", err
+	}
+	var response fileInfo
+	fileinfoURL := fmt.Sprintf("%s/mr/package/%s/%s/binfiles/%s/%s%s?fileinfo=1", snapshotURL, name, a.NonBinaryVersion, a.Name, a.NonBinaryVersion, a.BinaryVersion)
```
msuozzo, line 206: path.Join() and/or construct via url.URL

pkg/registry/debian/debian.go
```diff
@@ -177,9 +180,77 @@ func ParseDebianArtifact(artifact string) (DebianArtifact, error) {
...
+			if found[i].Name == artifact {
+				verified = true
+				break
+			}
+		}
+		if !verified {
+			return "", errors.Errorf("artifact name %s not found in fileinfo:%+v", artifact, response.FileInfo)
+		}
+	}
+	return fmt.Sprintf("%s/file/%s", snapshotURL, hash), nil
```
msuozzo, line 244: path.Join

msuozzo (approved):

pkg/registry/debian/debian.go
```diff
@@ -266,9 +269,88 @@ func (r HTTPRegistry) DSC(ctx context.Context, component, name, version string)
...
+	a, err := ParseDebianArtifact(artifact)
+	if err != nil {
+		return "", err
+	}
+	var response fileInfo
+	fileinfoURL := urlx.Copy(snapshotURL)
+	{
+		fileinfoURL.Path += path.Join("mr/package", name, a.Version.BinaryIndependentString(), "binfiles", a.Name, a.Version.String())
+		query := fileinfoURL.Query()
+		query.Add(
```
msuozzo, line 299: one line

pkg/registry/debian/debian.go
```diff
@@ -266,9 +269,88 @@ func (r HTTPRegistry) DSC(ctx context.Context, component, name, version string)
...
+	for _, f := range response.Result {
+		if f.Architecture == a.Arch {
+			hash = f.Hash
+			break
+		}
+	}
+	if hash == "" {
+		return "", errors.New("no matching architecture found")
+	}
+	// Verify we found the correct artifact
```
msuozzo, line 325: do we expect this to fail? i'm curious whether there are cases where the arch matches but the artifact doesn't. if there are such cases, do we want to just search through all the FileInfo entries in the first place and not do this lookup twice?

pkg/registry/debian/debian.go
```diff
@@ -266,9 +269,88 @@ func (r HTTPRegistry) DSC(ctx context.Context, component, name, version string)
...
+	}
+	// Result is a list of file hashes with architecture
+	Result []struct {
+		Architecture string
+		Hash         string
+	}
+}
+
+func (r HTTPRegistry) ArtifactURL(ctx context.Context, name, artifact string) (string, error) {
+	// Example series of urls to follow:
```
msuozzo, line 286: the wording of this line is a bit confusing.

pkg/registry/debian/debian.go
```diff
@@ -266,9 +269,88 @@ func (r HTTPRegistry) DSC(ctx context.Context, component, name, version string)
 	return DSCURI, d, err
 }
 
+// fileInfo is the response from the binfiles endpoint on the snapshot service.
+type fileInfo struct {
+	// FileInfo is a map from file hash to extra info
```
msuozzo, line 274: what is extra info?

pkg/registry/debian/debian.go
```diff
@@ -266,9 +269,88 @@ func (r HTTPRegistry) DSC(ctx context.Context, component, name, version string)
...
+	FileInfo map[string][]struct {
+		Name string
+	}
+	// Result is a list of file hashes with architecture
```
msuozzo, line 278: this comment describes the structure instead of the semantics. if we're not adding information, let's remove the comment

pkg/registry/debian/debian.go
```diff
@@ -266,9 +269,88 @@ func (r HTTPRegistry) DSC(ctx context.Context, component, name, version string)
...
+	// https://snapshot.debian.org/mr/package/acl/2.3.2-2/binfiles/libacl1/2.3.2-2+b1?fileinfo=1
+	// https://snapshot.debian.org/file/53f2b0612c8ed8a60970f9a206ae65eb84681f6e
+	a, err := ParseDebianArtifact(artifact)
+	if err != nil {
+		return "", err
+	}
+	var response fileInfo
+	fileinfoURL := urlx.Copy(snapshotURL)
+	{
+		fileinfoURL.Path += path.Join("mr/package", name, a.Version.BinaryIndependentString(), "binfiles", a.Name, a.Version.String())
```
msuozzo, line 297: should this really be `+=`? I'd either overwrite the Path or include it as the first arg of path.Join.

pkg/registry/debian/debian.go
wbxyz: I don't think I have a specific example where this isn't true, but doing this check seems like the correct thing to do for completeness.

Given how many sharp edges there are in Debian versioning, I'd prefer to keep the verification check just so any failure here is explicitly logged during this step rather than something less obvious later on during execution.

This is not a particularly expensive loop, most of the time `response.FileInfo[hash]` is a slice size of one. The only time this fileinfos is larger than one is if a specific hash has appeared under multiple different filenames, or maybe if the package appearing in testing first and then stable later.

msuozzo: yeah i wasn't concerned with the expense of the thing, only the complexity inherent in doing a matching procedure twice in both parts of the response. i might even prefer removing the arch matching if it means only doing one match procedure.

pkg/registry/debian/debian.go
wbxyz: First arg of path.Join is a good call. I think the += was left over from string-based formatting

pkg/registry/debian/debian.go
wbxyz: If we really must drop one, I think the arch matching is the one to keep. It's not apparent to me if a filename can be re-used for multiple hashes. Maybe across different distributions the same file name has different contents? IDK, I didn't feel comfortable making that the primary key.

Conversely, the url contains the source package name, the source version, the artifact stem name (like libacl1) and the binary version. Once we find an entry for a given arch, that's the entire set of information we have from the Target.

pkg/registry/debian/debian.go
```diff
@@ -266,9 +269,90 @@ func (r HTTPRegistry) DSC(ctx context.Context, component, name, version string)
...
+			break
+		}
+	}
+	if hash == "" {
+		return "", errors.New("no matching architecture found")
+	}
+	// Verify we found the correct artifact
+	{
+		verified := false
+		found, ok := response.FileInfo[hash]
```
msuozzo, line 330: nit: maybe "files" instead of found?

pkg/registry/debian/debian.go
```diff
@@ -266,9 +269,90 @@ func (r HTTPRegistry) DSC(ctx context.Context, component, name, version string)
...
+		if !ok {
+			return "", errors.Errorf("no fileinfo at %s, for the hash %s", fileinfoURL, hash)
+		}
+		for i := range found {
```
msuozzo, line 334: use second value instead of index

pkg/registry/debian/debian.go
msuozzo: I'm fine with leaving this as-is, I just want us to be able to understand and document what situations we're trying to identify. Tweaking the comment on this block might be good to add some more detail. Otherwise, let's ship it.

msuozzo (approved):

pkg/registry/debian/debian.go
wbxyz: I went with "infos" because of the type, agreed "found" isn't great.

pkg/registry/debian/debian.go
wbxyz: Matthew's comment got deleted, here's a copy:

> I'm fine with leaving this as-is, I just want us to be able to understand and document what situations we're trying to identify. Tweaking the comment on this block might be good to add some more detail. Otherwise, let's ship it.

Sounds good. Added a comment, and I also checked the ArchiveName (which is the repository from which the snapshot service pulled this artifact) to make it more clear how this check is different.

PR #385, opened 2025-03-13
Change debian registry to use *url.URL instead of strings

wbxyz:


msuozzo (approved):

pkg/registry/debian/debian.go
```diff
@@ -64,12 +66,17 @@ func poolDir(name string) string {
 }
 
 func PoolURL(component, name, artifact string) string {
-	return registryURL + fmt.Sprintf("%s/%s/%s/%s", component, poolDir(name), name, artifact)
+	// Copy the registryURL
+	u := urlx.MustParse(registryURL.String())
```
msuozzo, line 70: wait there should be a better way to copy, no?

msuozzo: ugh. I thought wrong.

msuozzo: well, on the other hand, you know what the URL is so we actually know that the shallow copy using a pointer will be safe:

u := *registryURL

wbxyz: I can't find a copy method on the standard library.

While writing this PR I looked around and other people seemed to be doing this: https://github.com/golang/go/issues/38351#issuecomment-2033232106
```
u := *otherURL
return &u
```

But that's two lines and doesn't obviously do a copy, especially because one of the members of url.URL is itself a pointer.

msuozzo: But since we know the URL being used and we know that there are no query params or user info, we can say that that doesn't matter, right?

wbxyz: Yeah we know that the literal doesn't have mutable things currently but

a) We would be making a hidden dependency on that property in a location separate from the definition
b) We don't know for sure if any users of this literal will set the parameters, which would percolate back to shared base, and then other users.

That, in addition to the less obvious copy semantics, makes me lean towards the existing string-based copy.

Is there a reason you think we should prefer the pointer dereference method of copying?

msuozzo: > users of this literal will set the parameters

wdym? we're returning a string here so the obj isn't leaking.

> Is there a reason you think we should prefer the pointer dereference method of copying?

Callsite clarity would be one since that's a common idiom for shallow-copying.

I guess one option would be to add urlx.Copy() which implements the re-parse approach. would remove the need for the comment.

wbxyz: Added urlx.Copy() in #391

msuozzo: dope. thanks

PR #389, opened 2025-03-14
Add testcase names

msuozzo:


internal/api/apiservice/rebuild_test.go
```diff
@@ -186,6 +191,7 @@ func TestRebuildPackage(t *testing.T) {
 			})),
 		},
 		{
+			name:   "deb binary-only release success",
```
wbxyz, line 194: Want to mention that these debian examples are using the DSC version of build strategy (called DebianPackage but could be renamed). This is opposed to the debrebuild strategy type that we've been using more recently but we don't have an example here yet.

wbxyz (approved):

internal/api/apiservice/rebuild_test.go
msuozzo: wanna do that in a follow-on? would probably be useful to also add a test for the new version

wbxyz: Yeah sure that sgtm.

PR #392, opened 2025-03-14
Add dedicated BuildDefinition type

msuozzo:
Motivated by #228

### DIFF
--- a/internal/api/apiservice/rebuild_test.go
+++ b/internal/api/apiservice/rebuild_test.go
@@ -11,18 +11,25 @@ import (
 	"encoding/json"
 	"io"
 	"net/http"
+	"os"
+	"path"
 	"strings"
 	"testing"
 	"time"
 
 	"github.com/go-git/go-billy/v5/memfs"
+	"github.com/go-git/go-billy/v5/osfs"
 	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/go-git/go-git/v5/plumbing/cache"
+	"github.com/go-git/go-git/v5/storage/filesystem"
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/google/oss-rebuild/internal/gcb/gcbtest"
+	"github.com/google/oss-rebuild/internal/gitx/gitxtest"
 	"github.com/google/oss-rebuild/internal/httpx/httpxtest"
 	"github.com/google/oss-rebuild/pkg/archive"
 	"github.com/google/oss-rebuild/pkg/archive/archivetest"
+	"github.com/google/oss-rebuild/pkg/builddef"
 	"github.com/google/oss-rebuild/pkg/rebuild/cratesio"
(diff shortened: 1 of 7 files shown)


internal/verifier/attestation.go
```diff
@@ -26,8 +26,7 @@ const (
 )
 
 // CreateAttestations creates the SLSA attestations associated with a rebuild.
-func CreateAttestations(ctx context.Context, input rebuild.Input, finalStrategy rebuild.Strategy, id string, rb, up ArtifactSummary, metadata rebuild.AssetStore, serviceLoc, prebuildLoc, buildDefLoc rebuild.Location) (equivalence, build *in_toto.ProvenanceStatementSLSA1, err error) {
-	t, manualStrategy := input.Target, input.Strategy
+func CreateAttestations(ctx context.Context, t rebuild.Target, defn *schema.BuildDefinition, strategy rebuild.Strategy, id string, rb, up ArtifactSummary, metadata rebuild.AssetStore, serviceLoc, prebuildLoc, buildDefLoc rebuild.Location) (equivalence, build *in_toto.ProvenanceStatementSLSA1, err error) {
```
wbxyz, line 29: Should `defn` include something indicating that it's not necessarily the definition that was actually used to build? Maybe inputDefn? Using the naming from the attestation, I guess it would be buildFix?

internal/api/apiservice/rebuild_test.go
```diff
@@ -372,8 +421,17 @@ RLpmHHG1JOVdOA==
...
+				relpath := path[1:]
+				buildDef := string(must(yaml.Marshal(tc.buildDef)))
+				commits := []gitxtest.Commit{
+					{Files: gitxtest.FileContent{relpath: buildDef}},
+				}
+				must(gitxtest.CreateRepo(commits, &repoOpts))
+			}
 
-			verdict, err := rebuildPackage(ctx, schema.RebuildPackageRequest{Ecosystem: tc.target.Ecosystem, Package: tc.target.Package, Version: tc.target.Version, Artifact: tc.target.Artifact}, &d)
+			verdict, err := rebuildPackage(ctx, schema.RebuildPackageRequest{Ecosystem: tc.target.Ecosystem, Package: tc.target.Package, Version: tc.target.Version, Artifact: tc.target.Artifact, StrategyFromRepo: tc.buildDef != nil}, &d)
```
wbxyz, line 434: nit: Should the RebuildPackageRequest.StrategyFromRepo field be renamed to match definition?

wbxyz (approved):

internal/verifier/attestation.go
msuozzo: the "definition" is the content of the repo so there is no definition when the strategy is entirely inferred. that is separate from the definition's Strategy property which seems the thing you're more describing.

The Strategy in the param list more represents the thing "actually used to build" which, when it's provided, will reflect the one present in the definition.

So I think the new names are consistent and reflective of their function.

internal/api/apiservice/rebuild_test.go
msuozzo: Yep good call

PR #397, opened 2025-03-20
Add shell-script as an output format in the infer tool

gkousik:
This is a prototype change - happy to take in feedback and address it!

### DIFF
--- a/tools/ctl/ctl.go
+++ b/tools/ctl/ctl.go
@@ -4,6 +4,7 @@
 package main
 
 import (
+	"bytes"
 	"context"
 	"crypto/sha256"
 	"encoding/csv"
@@ -21,6 +22,7 @@ import (
 	"slices"
 	"sort"
 	"strings"
+	"text/template"
 	"time"
 
 	"github.com/cheggaaa/pb"
@@ -51,6 +53,28 @@ var rootCmd = &cobra.Command{
 	Short: "A debugging tool for OSS-Rebuild",
 }
 
+var debuildShellScript = template.Must(
+	template.New(
+		"rebuild shell script",
+	).Funcs(template.FuncMap{
+		"join": func(sep string, s []string) string { return strings.Join(s, sep) },
+	}).Parse(
+		textwrap.Dedent(`
+# Install dependencies.
+set -eux
+apt update
+apt install -y {{join " " .SystemDeps}}
+
+mkdir /src && cd /src
+{{.Source}}
+{{.Deps}}
+
+# Run the build.
+{{.Build}}
+mkdir /out && cp /src/{{.OutputPath}} /out/
+`)[1:], // remove leading newline
+	))
+
 func buildFetchRebuildRequest(bench, run, prefix, pattern string, clean bool) (*rundex.FetchRebuildRequest, error) {
 	var runs []string
 	if run != "" {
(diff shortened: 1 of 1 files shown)


wbxyz:
Thanks for the change!

I think there will almost certainly be others interested in a non-container script version.

One thought: because the "production" infra won't be using MakeShellScript, wdyt about moving that function, and the template into ctl.go?

pkg/rebuild/rebuild/rebuildremote.go
```diff
@@ -177,6 +178,42 @@ var debuildContainerTpl = template.Must(
...
+			// NOTE: For syntax docs, see https://docs.docker.com/build/dockerfile/release-notes/
+			// TODO: Find a base image that has build-essentials installed, that would improve startup time significantly, and it would pin the build tools we're using.
+			textwrap.Dedent(`
+# Install dependencies.
+set -eux
+{{- if .UseTimewarp}}
+curl {{if .UtilPrebuildAuth}}-H @/run/secrets/auth_header {{end -}}
+https://{{.UtilPrebuildBucket}}.storage.googleapis.com/{{if .UtilPrebuildDir}}{{.UtilPrebuildDir}}/{{end}}timewarp > timewarp
+chmod +x timewarp
+{{- end}}
```
wbxyz, line 197: If this gets moved to ctl.go, I think we can drop the timewarp setup stuff to hopefully making this script simpler.

gkousik: Done, removed it.

gkousik:
> Thanks for the change!
> 
> I think there will almost certainly be others interested in a non-container script version.
> 
> One thought: because the "production" infra won't be using MakeShellScript, wdyt about moving that function, and the template into ctl.go?

Sure - I don't have much context on where to put the code. Took your suggestion and put it in ctl.go!

msuozzo:
Thanks, Kousik!

tools/ctl/ctl.go
```diff
@@ -51,6 +53,30 @@ var rootCmd = &cobra.Command{
...
+apt install -y {{join " " .SystemDeps}}
+
+mkdir /src && cd /src
+{{.Source}}
+{{.Deps}}
+
+# Run the build.
+{{.Build}}
+ls
+ls /src/
```
wbxyz, line 75: These can probably be cleaned up, they were helpful for debugging but not necessary for a "clean" script.

wbxyz:
> > Thanks for the change!
> > I think there will almost certainly be others interested in a non-container script version.
> > One thought: because the "production" infra won't be using MakeShellScript, wdyt about moving that function, and the template into ctl.go?
> 
> Sure - I don't have much context on where to put the code. Took your suggestion and put it in ctl.go!

No worries! Thanks for the changes :)

tools/ctl/ctl.go
```diff
@@ -51,6 +53,30 @@ var rootCmd = &cobra.Command{
 	Short: "A debugging tool for OSS-Rebuild",
 }
 
+var debuildShellScriptTpl = template.Must(
```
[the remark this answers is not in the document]
gkousik: Done.

tools/ctl/ctl.go
gkousik: Agreed, removed!

PR #398, opened 2025-03-27
Add a debian epoch of 1: on failures

wbxyz:
When snapshot doesn't have a matching version, try again using an epoch
of 1:

### DIFF
--- a/internal/api/apiservice/rebuild_test.go
+++ b/internal/api/apiservice/rebuild_test.go
@@ -331,21 +331,21 @@ RLpmHHG1JOVdOA==
 					URL: "https://snapshot.debian.org/mr/package/xz-utils/5.2.4/binfiles/xz-utils/5.2.4?fileinfo=1",
 					Response: &http.Response{
 						StatusCode: 200,
-						Body:       io.NopCloser(bytes.NewReader([]byte(`{"fileinfo":{"deadbeef":[{"archive_name":"debian","name":"xz-utils_5.2.4_amd64.deb"}]},"result":[{"architecture":"amd64","hash":"deadbeef"}]}`))),
+						Body:       httpxtest.Body(`{"fileinfo":{"deadbeef":[{"archive_name":"debian","name":"xz-utils_5.2.4_amd64.deb"}]},"result":[{"architecture":"amd64","hash":"deadbeef"}]}`),
 					},
 				},
 				{
 					URL: "https://snapshot.debian.org/file/deadbeef",
 					Response: &http.Response{
 						StatusCode: 200,
-						Body:       io.NopCloser(bytes.NewReader([]byte("deb_contents"))),
+						Body:       httpxtest.Body("deb_contents"),
 					},
 				},
 				{
 					URL: "https://buildinfos.debian.net/buildinfo-pool/x/xz-utils/xz-utils_5.2.4-1_all-amd64-source.buildinfo",
 					Response: &http.Response{
 						StatusCode: 200,
-						Body:       io.NopCloser(bytes.NewReader([]byte("buildinfo content"))),
+						Body:       httpxtest.Body("buildinfo content"),
 					},
 				},
 			},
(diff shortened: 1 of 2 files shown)


pkg/registry/debian/debian.go
```diff
@@ -266,9 +270,100 @@ func (r HTTPRegistry) DSC(ctx context.Context, component, name, version string)
...
+			query := fileinfoURL.Query()
+			query.Add(
+				"fileinfo",
+				"1",
+			)
+			fileinfoURL.RawQuery = query.Encode()
+		}
+		var rd io.ReadCloser
+		rd, err = r.get(ctx, fileinfoURL.String())
+		if err != nil && strings.Contains(err.Error(), "404 Not Found") {
```
msuozzo, line 311: this is...smelly. once we need to start interrogating the error value that's returned, the current get() impl really ceases to create cleaner code. is there a better way we can construct the error to avoid this? and/or could we use the http.StatusText() conversion function to get the string from http.StatusNotFound instead of this fixed string?

pkg/registry/debian/debian.go
```diff
@@ -292,25 +293,37 @@ func (r HTTPRegistry) ArtifactURL(ctx context.Context, name, artifact string) (s
...
-			"fileinfo",
-			"1",
-		)
-		fileinfoURL.RawQuery = query.Encode()
-	}
-	{
-		r, err := r.get(ctx, fileinfoURL.String())
-		if err != nil {
+	var fileinfoURL *url.URL
+	for _, epoch := range []string{"", "1"} {
```
msuozzo, line 297: would you be able to add a comment on what an epoch is and why we try two of them?

pkg/registry/debian/debian.go
wbxyz: Agreed. I decided to inline the get() impl here, and leave it elsewhere. In doing so, I discovered that the CachedClient is incorrectly converting non-ok messages into errors which goes against the standard library's implementation. Sent #406 to fix that.

internal/api/apiservice/rebuild_test.go
```diff
@@ -262,7 +269,81 @@ RLpmHHG1JOVdOA==
...
+						Status:     "404 Not Found",
+						StatusCode: 404,
+						Body:       io.NopCloser(bytes.NewReader([]byte(`404 Not Found`))),
+					},
+				},
+				{
+					URL: "https://snapshot.debian.org/mr/package/xz-utils/1:5.2.4/binfiles/xz-utils/1:5.2.4?fileinfo=1",
+					Response: &http.Response{
+						StatusCode: 200,
+						Body:       io.NopCloser(bytes.NewReader([]byte(`{"fileinfo":{"deadbeef":[{"name":"xz-utils_5.2.4_amd64.deb"}]},"result":[{"architecture":"amd64","hash":"deadbeef"}]}`))),
```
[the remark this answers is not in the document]
wbxyz: Added in #411

pkg/registry/debian/debian.go
```diff
@@ -300,22 +301,47 @@ func (r HTTPRegistry) ArtifactURL(ctx context.Context, name, artifact string) (s
...
-		fileinfoURL.RawQuery = query.Encode()
-	}
-	{
-		r, err := r.get(ctx, fileinfoURL.String())
+	var fileinfoURL *url.URL
+	// Epochs are used when the source version scheme has changed.
+	// Frequently the epoch is dropped from the version identifier used by the package manager
+	// (when only one or two versions are available for a given distribution, the source version schemes don't need to be disambiguated).
+	// If the artifact doesn't exist under an empty epoch, we try again with epoch "1" which frequently works.
+	// If this proves to be insufficient, we can get it from the .buildinfo file (which is stored at a URL missing the epoch but contains a version identifier that includes the epoch).
```
msuozzo, line 309: nit: maybe add a NOTE to the beginning of this so it stands out a bit?

pkg/registry/debian/debian.go
```diff
@@ -300,22 +301,47 @@ func (r HTTPRegistry) ArtifactURL(ctx context.Context, name, artifact string) (s
...
-			return "", err
+			return "", errors.Wrap(err, "building fileinfo request")
 		}
-		defer r.Close()
-		if err = json.NewDecoder(r).Decode(&response); err != nil {
+		var resp *http.Response
+		resp, err = r.Client.Do(req)
+		if err != nil {
+			return "", errors.Wrap(err, "fetching fileinfo")
+		} else if resp.StatusCode == 404 {
```
msuozzo, line 329: nit: use http.Status* constants here and below

pkg/registry/debian/debian.go
```diff
@@ -300,22 +301,47 @@ func (r HTTPRegistry) ArtifactURL(ctx context.Context, name, artifact string) (s
...
+			// That fileinfo doesn't exist, try the next epoch.
+			continue
+		} else if resp.StatusCode != 200 {
+			return "", errors.Wrap(errors.New(resp.Status), "fetching fileinfo")
+		}
+		defer resp.Body.Close()
+		if err = json.NewDecoder(resp.Body).Decode(&response); err != nil {
```
msuozzo, line 336: if you use the declaration operator here, you shouldn't need the extra statement on 339

pkg/registry/debian/debian.go
```diff
@@ -300,22 +301,47 @@ func (r HTTPRegistry) ArtifactURL(ctx context.Context, name, artifact string) (s
...
 			return "", err
 		}
+		err = nil
+		// If we succeeded, break out of the loop.
+		break
+	}
+	if err != nil {
```
msuozzo, line 343: hmm i don't think we check the case where we got 2 404s. i think this predicate should be tweaked (also because I don't think err will be non-nil)

pkg/registry/debian/debian.go
```diff
@@ -300,22 +301,47 @@ func (r HTTPRegistry) ArtifactURL(ctx context.Context, name, artifact string) (s
...
+		}
+		var req *http.Request
+		req, err = http.NewRequestWithContext(ctx, http.MethodGet, fileinfoURL.String(), nil)
 		if err != nil {
-			return "", err
+			return "", errors.Wrap(err, "building fileinfo request")
 		}
-		defer r.Close()
-		if err = json.NewDecoder(r).Decode(&response); err != nil {
+		var resp *http.Response
```
msuozzo, line 325: this can be a declaration

msuozzo (approved):

pkg/registry/debian/debian.go
wbxyz: Changed the behavior such that the loop only sets fileinfoURL on success, so a nil fileinfoURL means the loop failed.

PR #404, opened 2025-03-27
Add timeout cancellation on GCB calls

wbxyz:


internal/api/apiservice/rebuild.go
```diff
@@ -353,6 +353,7 @@ func rebuildPackage(ctx context.Context, req schema.RebuildPackageRequest, deps
 func RebuildPackage(ctx context.Context, req schema.RebuildPackageRequest, deps *RebuildPackageDeps) (*schema.Verdict, error) {
 	started := time.Now()
 	ctx = context.WithValue(ctx, rebuild.RunID, req.ID)
+	ctx = context.WithValue(ctx, rebuild.GCBDeadlineID, time.Now().Add(req.BuildTimeout))
```
msuozzo, line 356: I think we should leave the specific remote executor out of the constant name so this would be RemoteDeadlineID or BuildDeadlineID?

internal/gcb/gcb.go
```diff
@@ -59,15 +60,41 @@ func (c *clientImpl) WaitForOperation(ctx context.Context, op *cloudbuild.Operat
...
-func DoBuild(ctx context.Context, client Client, project string, build *cloudbuild.Build) (*cloudbuild.Build, error) {
+func DoBuild(ctx context.Context, client Client, project string, build *cloudbuild.Build, deadline time.Time) (*cloudbuild.Build, error) {
 	op, err := client.CreateBuild(ctx, project, build)
 	if err != nil {
 		return nil, err
 	}
-	op, err = client.WaitForOperation(ctx, op)
-	if err != nil {
-		return nil, errors.Wrap(err, "fetching operation")
+	ctx, cancel := context.WithDeadline(ctx, deadline)
```
msuozzo, line 75: can you configure the deadline on the ctx (i.e. one layer up in rebuildremote) and keep the monitoring code here?

pkg/rebuild/rebuild/rebuildremote.go
```diff
@@ -477,7 +477,11 @@ func makeBuild(t Target, dockerfile string, opts RemoteOptions) (*cloudbuild.Bui
 }
 
 func doCloudBuild(ctx context.Context, client gcb.Client, build *cloudbuild.Build, opts RemoteOptions, bi *BuildInfo) error {
-	build, err := gcb.DoBuild(ctx, client, opts.Project, build)
+	deadline := time.Now().Add(55 * time.Minute)
```
msuozzo, line 480: i don't think there should be a default, at least not here. If there's no configured dealine, let's avoid setting one. instead, we could configure this at the top-level handler (apiservice) where we can hard-code the specifics we know about the service. I suppose we _could_ do so in the gcb code but that would encode specifics of the type of build cluster we assume we're running in so it'd probably be overkill atm.

internal/gcb/gcb.go
wbxyz, line 75: Add a TODO mentioning the GCB Build native Timeout setting, which we might want to set as well.

internal/gcb/gcb.go
```diff
@@ -59,15 +60,41 @@ func (c *clientImpl) WaitForOperation(ctx context.Context, op *cloudbuild.Operat
...
+	case err := <-done:
+		log.Printf("Finished op: %v", err)
+	case <-ctx.Done():
+		if err := ctx.Err(); err == context.DeadlineExceeded {
+			if err := client.CancelOperation(op); err != nil {
+				log.Printf("Best effort GCB cancellation failed: %v", err)
+				// After cancelling, wait for the operation one more time to update the op.Error and the op.Status
+			} else if op, err = client.WaitForOperation(ctx, op); err != nil {
+				log.Printf("Failed update operation after cancellation: %v", err)
+			}
```
wbxyz, line 96: Instead of waiting again, pull out from `done` channel to the above goroutine to finish

pkg/rebuild/rebuild/rebuildremote.go
wbxyz: Moving the deadline to the top level apiservice SGTM.

internal/gcb/gcb.go
wbxyz: Yup sounds good.

internal/api/apiservice/rebuild.go
wbxyz: We talked about it a bit offline, when we eventually abstract away the specific implementation of build infra, we will use client objects that self-contain the timeout duration.

Until then, the writer for this context var will need to be aware of the backend implementation to choose an appropriate timeout (GCB has a 60 minute maximum). For that reason, I think it's probably best to let the var reflect the implementation.

I did make sure to keep the API variable agnostic (RebuildPackageRequest.BuildTimeout) which should give us the freedom to swap the implementation in the future.

internal/gcb/gcb.go
```diff
@@ -59,15 +60,39 @@ func (c *clientImpl) WaitForOperation(ctx context.Context, op *cloudbuild.Operat
...
+	case err := <-done:
+		log.Printf("Finished op: %v", err)
+	case <-ctx.Done():
+		if err := ctx.Err(); err == context.DeadlineExceeded {
+			if err := client.CancelOperation(op); err != nil {
+				log.Printf("Best effort GCB cancellation failed: %v", err)
+			}
+			// After cancelling, allow WaitForOperation to return
+			err = <-done
+			log.Printf("Error after cancelling GCB: %v", err)
```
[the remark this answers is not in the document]
wbxyz: Agreed, re-worked the handling of <-done.

internal/gcb/gcb.go
```diff
@@ -59,15 +60,39 @@ func (c *clientImpl) WaitForOperation(ctx context.Context, op *cloudbuild.Operat
...
+			done <- errors.Wrap(err, "fetching operation")
+		} else {
+			done <- nil
+		}
+	}()
+	select {
+	case err := <-done:
+		log.Printf("Finished op: %v", err)
+	case <-ctx.Done():
+		if err := ctx.Err(); err == context.DeadlineExceeded {
```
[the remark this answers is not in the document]
wbxyz: We decided this is an unlikely case, so it's not a big deal either way. Added a note in the comments.

internal/gcb/gcbtest/gcb_test.go
msuozzo, line 1: we can break the circular dep by moving the interface conformance assertion in mockclient.go into a test file

msuozzo (approved):

PR #406, opened 2025-03-28
Change CachedClient to store non-200 responses

wbxyz:
I don't really have a strong preference on whether these responses are actually stored. The more important thing is to avoid throwing away the non-200 responses (as is done currently by converting them to error strings). By doing so, the current implementation prevents any caller from parsing these non-200 response messages.

See this snippet from the standard library:

```
// An error is returned if caused by client policy (such as
// CheckRedirect), or failure to speak HTTP (such as a network
// connectivity problem). A non-2xx status code doesn't cause an
// error.
```

### DIFF
--- a/internal/httpx/http.go
+++ b/internal/httpx/http.go
@@ -7,7 +7,6 @@ package httpx
 import (
 	"bufio"
 	"bytes"
-	"errors"
 	"net/http"
 	"time"
 
@@ -55,9 +54,6 @@ func (cc *CachedClient) Do(req *http.Request) (*http.Response, error) {
 		resp, err := cc.BasicClient.Do(req)
 		if err != nil {
 			return nil, err
-		}
-		if resp.StatusCode != http.StatusOK {
-			return nil, errors.New(resp.Status)
 		}
 		defer resp.Body.Close()
 		foo := new(bytes.Buffer)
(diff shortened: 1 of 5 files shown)


internal/httpx/http.go
```diff
@@ -56,9 +55,6 @@ func (cc *CachedClient) Do(req *http.Request) (*http.Response, error) {
 		if err != nil {
 			return nil, err
 		}
-		if resp.StatusCode != http.StatusOK {
```
msuozzo, line 59: yeah I think this is basically a bug. that said, we really shouldn't be caching 5XX errors. additionally, we might also implement some minimal subset of the http caching spec i.e. https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Caching#dont_cache

msuozzo (approved):

internal/httpx/http.go
wbxyz: I'll punt on the behavioral changes, this PR is blocking for the debian epoch tests to work.

I filled https://github.com/google/oss-rebuild/issues/409 to track a better implementation of the cache.

wbxyz: Ah right, yes this would have added 5XX caching. I updated the behavior and added tests.

internal/httpx/http.go
```diff
@@ -51,22 +50,23 @@ func (cc *CachedClient) Do(req *http.Request) (*http.Response, error) {
...
 		defer resp.Body.Close()
 		foo := new(bytes.Buffer)
 		if err := resp.Write(foo); err != nil {
 			return nil, err
 		}
-		return foo.Bytes(), nil
-	})
-	if err != nil {
+		// Only cache if it isn't a server error
+		if resp.StatusCode < 500 || resp.StatusCode >= 600 {
```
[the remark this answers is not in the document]
wbxyz: Nice

PR #407, opened 2025-03-28
Fix timewarp support in rebuilder

msuozzo:


pkg/rebuild/rebuild/rebuildone.go
```diff
@@ -28,8 +28,12 @@ func RebuildOne(ctx context.Context, r Rebuilder, input Input, mux RegistryMux,
 		if hint, ok := input.Strategy.(*LocationHint); ok && hint != nil {
 			repoURI = hint.Repo
 		} else {
+			var timewarpHost string
+			if host := ctx.Value(TimewarpID); host != nil {
+				timewarpHost = host.(string)
```
wbxyz, line 33: nit: If you do the .(string) cast inside the if you can also check for succesful casts:

https://github.com/google/oss-rebuild/blob/5bde322b7e50338183856c1143ca0b5ed80a40ad/pkg/rebuild/rebuild/rebuildone.go#L93

wbxyz (approved):

pkg/rebuild/rebuild/rebuildone.go
msuozzo: nice. fixed

PR #410, opened 2025-03-31
Stabilize TUI diff inputs

msuozzo:
This isn't perfect as the actual stabilize comparison may have been different
but, for now, this better reflects the comparison made during the rebuild.

### DIFF
--- a/tools/ctl/ide/ui.go
+++ b/tools/ctl/ide/ui.go
@@ -20,6 +20,7 @@ import (
 	"time"
 
 	"github.com/gdamore/tcell/v2"
+	"github.com/google/oss-rebuild/pkg/archive"
 	"github.com/google/oss-rebuild/pkg/rebuild/rebuild"
 	"github.com/google/oss-rebuild/pkg/rebuild/schema"
 	"github.com/google/oss-rebuild/tools/benchmark"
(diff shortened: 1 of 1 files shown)


tools/ctl/ide/ui.go
```diff
@@ -140,15 +158,35 @@ func diffArtifacts(ctx context.Context, butler localfiles.Butler, example rundex
...
+				return errors.Wrap(err, "fetching upstream asset")
+			}
+		} else {
+			{
+				rba, err = butler.Fetch(ctx, example.RunID, example.WasSmoketest(), rebuild.RebuildAsset.For(t))
+				if err != nil {
+					return errors.Wrap(err, "fetching rebuild asset")
+				}
+				// TODO: We should use the version of Stabilize used in the rebuild.
+				stabilized := "stabilized-" + rba
```
wbxyz, line 172: Does this need to be a path? Or maybe stabilizeArtifact should make a tmpfile and return the path?

wbxyz (approved):

tools/ctl/ide/ui.go
msuozzo: oh you're totally right that name got broken when i tried to change it to a prefix. temp file also sounds better given i don't think the first attempt would have worked for serial diffs.

PR #411, opened 2025-03-31
Add Body helper to httpxtest

wbxyz:


internal/httpx/httpxtest/mockclient.go
```diff
@@ -17,6 +17,11 @@ type Call struct {
 	Error    error
 }
 
+func (c Call) Request() *http.Request {
```
msuozzo, line 20: move to pr where this is used

internal/httpx/httpxtest/mockclient.go
```diff
@@ -17,6 +17,11 @@ type Call struct {
...
+	req, _ := http.NewRequest(c.Method, c.URL, nil)
```
msuozzo, line 21: probably best to panic on this case, just to be sure

msuozzo (approved):

PR #413, opened 2025-04-02
fix: parse pom to get the file path of git properties

algomaster99:
Reference: https://github.com/google/oss-rebuild/pull/360#discussion_r2000186342

The changes in this PR parse POMs in the jar file to get the precise file name of the git properties file.

Basic configuration of the `git-commit-id-plugin` plugin is
```xml
<plugin>
    <groupId>pl.project13.maven</groupId>
    <artifactId>git-commit-id-plugin</artifactId>
    <version>4.9.10</version>
    <configuration>
        <generateGitPropertiesFile>true</generateGitPropertiesFile>
        <generateGitPropertiesFilename>${project.build.outputDirectory}/classes/git.json</generateGitPropertiesFilename>
    </configuration>
</plugin>
```
If `generateGitPropertiesFilename` is present, we know that the there is a custom path ot the git properties file. 

Now based on the XML schema of POM file, this plugin could be in three places according to [its schema ](https://maven.apache.org/xsd/maven-4.0.0.xsd)

1. `<build>` -> `<plugins>` (Most common)
2. `<profiles>` -> `<profile>` -> `<build>` -> `<plugins>`
3. `<build>` -> `<pluginManagement>` -> `<plugins>`
> We don't need to care about this because `pluginManagement` just configures the plugin and the module that inherit them, will actually execute the plugin.

I have added tests to check both 1 and 2. Apart from this, we also needed a parser for pom properties like `${project.build.outputDirectory}`. For now, I have just configured defaults
```go
map[string]string{
		"project.build.directory":       "",
		"project.build.outputDirectory": "classes",
	}
```


algomaster99:
@msuozzo review whenever you have time :)

algomaster99:
I found an edge case to this implementation. I have a jar file ([arthas-spring-boot-starter-3.3.9.jar](https://repo1.maven.org/maven2/com/taobao/arthas/arthas-spring-boot-starter/3.3.9/arthas-spring-boot-starter-3.3.9.jar)) that contains the following pom.xml file.

```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>

    <parent>
        <groupId>com.taobao.arthas</groupId>
        <artifactId>arthas-all</artifactId>
        <version>3.3.9</version>
        <relativePath>../pom.xml</relativePath>
    </parent>

	...
```

When this JAR is packaged, I see `arthas-git.properties` in the JAR as well. Now configuration to generate this `git properties` file is not there in the above POM, but in its parent POM. Now we don't have access to [parent POM](https://github.com/alibaba/arthas/blob/master/pom.xml) because it is published [as another maven artifact](https://repo1.maven.org/maven2/com/taobao/arthas/arthas-all/3.3.9/).
The only way I see to solve this here would be to cURL the parent pom using the information in `parent`.

Maven 4 should fix these problems as they want to seggregate these build details from the POM file.

pkg/archive/jar.go
```diff
@@ -177,23 +178,40 @@ func splitPreservingQuotes(s string, sep rune) []string {
...
+		// We find all the git properties files in the jar file by looking at the pom.xml files.
+		// We then remove the content of these files in order to stabilize the jar file.
+		var gitPropertiesFiles []string
+		for _, zf := range zr.File {
+			if strings.HasSuffix(zf.Name, "pom.xml") {
+				reader, err := zf.Open()
+				if err != nil {
+					continue
+				}
+				content, err := io.ReadAll(reader)
```
msuozzo, line 197: hmm I guess maybe log this error? I think you'd also want to return in this case, right? we don't have a great story around bubbling up stabilizer errors. come to think of it, maybe this should just panic... we wouldn't want this to output a value we couldn't successfully stabilize.

pkg/archive/pom.go
```diff
@@ -0,0 +1,78 @@
+package archive
+
+import (
+	"encoding/xml"
+	"regexp"
+)
+
+type Build struct {
```
msuozzo, line 8: nit: because this is in the package "archive" and public, we need to consider whether archive.Build and archive.Project are good names for these symbols. do we want to make these private instead to avoid adding a prefix?

pkg/archive/jar_test.go
```diff
@@ -528,3 +567,11 @@ func TestStableGitProperties(t *testing.T) {
 		})
 	}
 }
+
+func readFileAndHandleException(path string) []byte {
```
msuozzo, line 571: for these cases, we can use must() which I believe is already defined in this package but, otherwise, should be accessible elsewhere in the project:

data := must(os.ReadFile(...))

pkg/archive/archive_testdata/git/custom/path/in/profile/pom.xml
msuozzo, line 1: nit: I think pkg/archive/testdata is the better top-level dir name to avoid naming stutter.

Additionally, I think we should move these to the top-level in testdata since the additional path elements don't provide much added context IMO. that would mean renaming this to e.g. "testdata/custom_profile_path_pom.xml" or something like that.

pkg/archive/archive_testdata/git/custom/path/in/build/pom.xml
msuozzo, line 1: I'm concerned with including what looks like a build artifact from another project in this repo. is there some way to minimize this to just the element you're testing?

pkg/archive/pom.go
```diff
@@ -0,0 +1,78 @@
...
+	var gitPropertiesFiles []string
+	for _, plugin := range build.Plugins.Plugin {
+		if plugin.ArtifactID == "git-commit-id-plugin" && plugin.GroupID == "pl.project13.maven" {
+			gitPropertiesFiles = append(gitPropertiesFiles, parsePomProperties(plugin.Configuration.GenerateGitPropertiesFilename))
+		}
+	}
+	return gitPropertiesFiles
+}
+
+func parsePomProperties(property string) string {
```
msuozzo, line 60: I think this is the only function defined in this file which should remain here given its general name (pom.go). for now, I think its best to move the other functions to the file that defines the git-commit-id stabilizer.

pkg/archive/pom.go
```diff
@@ -0,0 +1,78 @@
...
+	// If they do, we would need to parse the properties in POM file.
+	defaultProperties := map[string]string{
+		"project.build.directory":       "",
+		"project.build.outputDirectory": "classes",
+	}
+	propertyRegex := regexp.MustCompile(`\$\{[^}]+\}/?`)
+	return propertyRegex.ReplaceAllString(property, defaultProperties[property])
+}
+
+func addDefaultPaths(paths *[]string) {
```
msuozzo, line 71: I'd probably define this as a constant like defaultGitPaths instead of a function.

pkg/archive/pom.go
```diff
@@ -0,0 +1,78 @@
...
+			Build Build `xml:"build"`
+		} `xml:"profile"`
+	} `xml:"profiles"`
+}
+
+func getAllGitPropertiesFileInPom(pom []byte) []string {
+	// This function returns all the git properties files that are generated by git-commit-id-maven-plugin
+	// in the given pom file.
+	// In theory, there can be multiple git properties files in a single pom file so we return a list of them.
+	mavenProject := Project{}
```
msuozzo, line 35: nit: I think parsing should probably be handled elsewhere and this function accept the already-parsed project. that way the caller can decide how to handle this error

msuozzo (commented):
thanks for handling this tricky case! this is a great

algomaster99:
I will take a look at this and the other PR soon :)

pkg/archive/jar.go
algomaster99: > maybe this should just panic... we wouldn't want this to output a value we couldn't successfully stabilize.

I agree. Having an unparseable `pom.xml` file should be notified.

pkg/archive/pom.go
algomaster99: > we need to consider whether archive.Build and archive.Project are good names for these symbols

they are not indeed :sweat_smile: Better name would be `archive.pom.Build` perhaps. 

> do we want to make these private instead to avoid adding a prefix?

Since these types are only used in `pom.go`, it does make sense to make them private. If we decide to make it public, we can change it to `archive.pom.Build`.

pkg/archive/archive_testdata/git/custom/path/in/build/pom.xml
algomaster99: I minimised it by removing unnecessary configurations. However, I  kept `groupId`, `artifactId`, and `version` so we could trace back this test resource to the original project. For an actual POM, these are compulsory properties, but our parsing logic ignores it. I could also remove these attributes and instead put a URL in order to trace back - https://repo1.maven.org/maven2/org/jrivard/xmlchai/xmlchai/0.1.0/xmlchai-0.1.0.pom.

pkg/archive/pom.go
algomaster99: I moved all functions and types related to pom parsing to `archive/manifest/pom.go`. Exported functions are now called `manifest.Build` which I believe is a much better name.

pkg/archive/pom.go
algomaster99: Handled in `archive/manifest/pom.go`.

algomaster99:
@msuozzo ping for review. Hopefully, we have not lost the context due to my travelling, paper submission, and now sick leave :sweat_smile: Let me know if you have more feedback.

I did not handle the edge case as I described [here](https://github.com/google/oss-rebuild/pull/413#issuecomment-2781514336). I can keep that for a separate PR.

PR #415, opened 2025-04-07
Change Translate to populate URL params not body

wbxyz:
The Handlers are all expecting the request in the params, I think this
was broken before.

### DIFF
--- a/internal/api/rpc.go
+++ b/internal/api/rpc.go
@@ -136,6 +136,7 @@ func Handler[I Message, O any, D Dependencies](initDeps InitT[D], handler Handle
 
 type Translator[O Message] func(io.ReadCloser) (O, error)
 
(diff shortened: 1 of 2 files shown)


internal/api/rpc.go
```diff
@@ -136,6 +136,7 @@ func Handler[I Message, O any, D Dependencies](initDeps InitT[D], handler Handle
 
 type Translator[O Message] func(io.ReadCloser) (O, error)
 
+// Translate uses a Translator on the Request.Body to populate the request.URL params.
```
msuozzo, line 139: want to add a test for this?

msuozzo (approved):

internal/api/rpc.go
wbxyz: Added

PR #418, opened 2025-04-08
Add ecosystem-specific stabilizer logic

msuozzo:


pkg/rebuild/stability/stabilizer.go
```diff
@@ -0,0 +1,36 @@
...
+	format := t.ArchiveType()
+	if format == archive.UnknownFormat {
+		return nil, errors.Errorf("unknown archive format for %s %s", t.Ecosystem, t.Artifact)
+	}
+	var stabilizers []archive.Stabilizer
+	switch format {
+	case archive.ZipFormat:
+		stabilizers = slices.Concat(archive.AllZipStabilizers)
+	case archive.TarFormat:
+		stabilizers = slices.Concat(archive.AllTarStabilizers)
```
wbxyz, line 25: slices.Clone()?

wbxyz (approved):

cmd/stabilize/main.go
```diff
@@ -134,6 +139,56 @@ func determinePasses(reg StabilizerRegistry, enableSpec, disableSpec string) ([]
 	return toRun, nil
 }
 
+func candidateEcosystems(filename string) []rebuild.Ecosystem {
+	ext := filepath.Ext(filename)
+	switch ext {
+	case ".jar":
```
algomaster99, line 145: For artifacts in Maven that have JVM bytecode, they can have more extensions that `.jar`.
```
jar  829
war  25
zip  20
nar  8
gz   12
tar  4
```
Source: https://github.com/chains-project/reproducible-central/blob/master/diffoscope_files_with_jvm.txt (each file is a path to difference between reference and rebuild artifact) (the diff is not yet uploaded)


But since the numbers of other extension is pretty low, we can expect the user to provide `ecosystem = MAVEN` to ensure that Jar stabilizer is enabled.

msuozzo: Ah TIL this is great! Handling war and potentially zip might be preferable..

algomaster99: Actually, if I enumerate all extensions of the [unreproducible artifacts](https://github.com/chains-project/reproducible-central/blob/master/java/unreproducible_maven_projects_to_releases.json) (are unreproducible because of any reason including changes in JVM bytecode) I have found, this is the correct number:
```
.jar        9702
.zip        581
.deb        7
.json       352
.xml        515
.sha512     9
.war        131
.gz         127
.kar        6
.pom        557
.bz2        4
.aar        2
.mar        16
.module     124
.nar        88
.html       5
.tgz        6
.buildinfo  5
.sh         5
.far        6
.opear      3
.rar        17
.tar        5
.exe        24
```
Now, of course, we cannot include all of them and it does not make sense too. For example, .exe is just an executable and can be stabilized by jar stabilizer anyway.
Right now, I can come up with a list of file extensions where I saw a diff in either `MANIFEST` or `git.properties`. And then we can add those extensions.

PR #421, opened 2025-04-08
Document network proxy

msuozzo:


cmd/proxy/README.md
```diff
@@ -0,0 +1,70 @@
+# Proxy
+
+Proxy is a transparent HTTP(S) proxy that intercepts and records network activity. It's primarily used within OSS Rebuild to monitor network interactions during the build process, helping to passively enumerate remote dependencies and to identify suspect build behavior.
```
wbxyz, line 3: Maybe mention TLS-terminating, because that's such a key feature?

"Proxy is a transparent, TLS-terminating, HTTP(S) proxy that..."

wbxyz (approved):

cmd/proxy/README.md
msuozzo: I think it's good to highlight more than was currently done. I bumped that section up in the order of appearance in the Overview section. I think it's a bit clunky to include in the one-line description, though, so i'm inclined to keep that as-is.

PR #422, opened 2025-04-08
Document stabilize

msuozzo:


cmd/stabilize/README.md
```diff
@@ -0,0 +1,57 @@
+# Stabilize
+
+Stabilize is a command-line tool that removes non-deterministic metadata from software packages to facilitate functional artifact comparison.
```
wbxyz, line 3: nit: Maybe format to use the actual command name in code format? Something like 

The `stabilize` command-line tool removes non-deterministic ...

cmd/stabilize/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Overview
+
+Software artifacts often contain non-deterministic metadata (timestamps, file ordering, compression levels, etc.) that can cause bit-for-bit differences in otherwise identical build results. Stabilize addresses this problem by applying transformations to archive files to make them consistently comparable.
```
wbxyz, line 7: Maybe worth mentioning that we only stabilize differences that aren't security significant?

wbxyz (approved):

cmd/stabilize/README.md
msuozzo: yeah done

cmd/stabilize/README.md
msuozzo: It's an option. the other READMEs don't use that form, though, so for the sake of consistency i'll use normal text for now.

msuozzo: ok given that "stabilize" is a much more common/real verb than Timewarp, using code text to distinguish it as the proper noun sounds good.

PR #423, opened 2025-04-10
rundex.FetchRebuilds should return slice not map

wbxyz:
This more closely matches rundex.FetchRuns. It also allows FetchRebuilds
to return multiple Rebuild objects for a given target.

### DIFF
--- a/tools/ctl/ctl.go
+++ b/tools/ctl/ctl.go
@@ -75,7 +75,7 @@ mkdir /out && cp /src/{{.OutputPath}} /out/
 `)[1:], // remove leading newline
 	))
 
-func buildFetchRebuildRequest(bench, run, prefix, pattern string, clean bool) (*rundex.FetchRebuildRequest, error) {
+func buildFetchRebuildRequest(bench, run, prefix, pattern string, clean, latestOnly bool) (*rundex.FetchRebuildRequest, error) {
 	var runs []string
 	if run != "" {
 		runs = strings.Split(run, ",")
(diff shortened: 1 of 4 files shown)


tools/ctl/ctl.go
```diff
@@ -75,7 +75,7 @@ mkdir /out && cp /src/{{.OutputPath}} /out/
 `)[1:], // remove leading newline
 	))
 
-func buildFetchRebuildRequest(bench, run, prefix, pattern string, clean bool) (*rundex.FetchRebuildRequest, error) {
+func buildFetchRebuildRequest(bench, run, prefix, pattern string, clean, latestOnly bool) (*rundex.FetchRebuildRequest, error) {
```
msuozzo, line 78: is this supposed to be in a separate PR?

tools/ctl/rundex/rundex.go
```diff
@@ -187,10 +187,11 @@ type FetchRebuildOpts struct {
...
 type FetchRebuildRequest struct {
-	Bench     *benchmark.PackageSet
-	Executors []string
-	Runs      []string
-	Opts      FetchRebuildOpts
+	Bench      *benchmark.PackageSet
+	Executors  []string
+	Runs       []string
+	Opts       FetchRebuildOpts
+	LatestOnly bool
```
msuozzo, line 194: nit: `Latest`?

tools/ctl/ctl.go
wbxyz: Clarified in the PR description. It's needed to replicate the previous implicit behavior.

msuozzo (approved):

PR #428, opened 2025-04-14
Inferring source distribution and python version

giacomobenedetti:
For source distribution:
* The source distribution can be manipulated/compromised the same way of bdists. Moreover, it may help as fallback when only platform-specific wheels are available.

For Python version:
* it may be not necessary, I didn't find any specific reproducibility issue caused by python version.
* it avoids some build failures caused by no longer supported libraries
* a step is added to the strategy for rebuilding Python packages (ce3555313de511d7755bfa1e8adad21a7ea657a5)

This PR requires the `Executor` logic proposed in PR https://github.com/google/oss-rebuild/pull/427
In case that PR isn't accepted I can decouple this dependency!

### DIFF
--- a/pkg/rebuild/pypi/infer.go
+++ b/pkg/rebuild/pypi/infer.go
@@ -4,8 +4,10 @@
 package pypi
 
 import (
+	"archive/tar"
 	"archive/zip"
 	"bytes"
+	"compress/gzip"
 	"context"
 	"fmt"
 	"io"
@@ -14,6 +16,7 @@ import (
 	re "regexp"
 	"slices"
 	"strings"
+	"time"
 
 	"github.com/go-git/go-billy/v5"
 	"github.com/go-git/go-git/v5"
@@ -149,6 +152,9 @@ func extractPyProjectRequirements(ctx context.Context, tree *object.Tree) ([]str
 		// TODO: Some of these requirements are probably already in rbcfg.Requirements, should we skip
 		// them? To even know which package we're looking at would require parsing the dependency spec.
 		// https://packaging.python.org/en/latest/specifications/dependency-specifiers/#dependency-specifiers
+		if strings.Contains(r, "python_version") {
+			r = strings.Split(r, ";")[0]
+		}
 		reqs = append(reqs, strings.ReplaceAll(r, " ", ""))
 	}
 	log.Println("Added these reqs from pyproject.toml: " + strings.Join(reqs, ", "))
@@ -187,29 +193,67 @@ func FindPureWheel(artifacts []pypireg.Artifact) (*pypireg.Artifact, error) {
 	return nil, fs.ErrNotExist
 }
 
-func inferRequirements(name, version string, zr *zip.Reader) ([]string, error) {
+func FindSourceDistribution(artifacts []pypireg.Artifact) (*pypireg.Artifact, error) {
+	for _, r := range artifacts {
+		if strings.HasSuffix(r.Filename, ".tar.gz") {
+			return &r, nil
+		}
+	}
+	return nil, fs.ErrNotExist
+}
+
+func inferRequirements(name, version string, zr interface{}) ([]string, error) {
 	// Name and version have "-" replaced with "_". See https://packaging.python.org/en/latest/specifications/recording-installed-packages/#the-dist-info-directory
-	// TODO: Search for dist-info in the gzip using a regex. It sounds like many tools do varying amounts of normalization on the path name.
+	// TODO: Search for dist-info in the archive using a regex. It sounds like many tools do varying amounts of normalization on the path name.
 	wheelPath := fmt.Sprintf("%s-%s.dist-info/WHEEL", strings.ReplaceAll(name, "-", "_"), strings.ReplaceAll(version, "-", "_"))
-	wheel, err := getFile(wheelPath, zr)
+
+	var wheel []byte
+	var err error
+	var reqs []string
+
+	switch reader := zr.(type) {
+	case *zip.Reader:
+		wheel, err = getFile(wheelPath, reader)
+	case *tar.Reader:
+		return reqs, nil
+		// wheel, err = getFileFromTarGz(wheelPath, reader)
+	default:
+		return nil, errors.New("[INTERNAL] Unsupported archive reader type")
+	}
+
 	if err != nil {
 		return nil, errors.Wrapf(err, "[INTERNAL] Failed to extract upstream %s", wheelPath)
 	}
-	reqs, err := getGenerator(wheel)
+
+	reqs, err = getGenerator(wheel)
 	if err != nil {
 		return nil, errors.Wrapf(err, "[INTERNAL] Failed to get upstream generator")
 	}
+
 	// Determine setuptools version.
 	if slices.ContainsFunc(reqs, func(s string) bool { return strings.HasPrefix(s, "setuptools==") }) {
 		// setuptools already set.
 		return reqs, nil
 	}
+
 	// TODO: Also find this with a regex.
 	metadataPath := fmt.Sprintf("%s-%s.dist-info/METADATA", strings.ReplaceAll(name, "-", "_"), strings.ReplaceAll(version, "-", "_"))
-	metadata, err := getFile(metadataPath, zr)
+
+	var metadata []byte
+
+	switch reader := zr.(type) {
+	case *zip.Reader:
+		metadata, err = getFile(metadataPath, reader)
+	case *tar.Reader:
+		metadata, err = getFileFromTarGz(metadataPath, reader)
+	default:
+		return nil, errors.New("[INTERNAL] Unsupported archive reader type")
+	}
+
 	if err != nil {
 		return nil, errors.Wrapf(err, "[INTERNAL] Failed to extract upstream dist-info/METADATA")
 	}
+
 	switch {
 	case !bytes.Contains(metadata, []byte("License-File")):
 		// The License-File value was introduced in later versions so this is the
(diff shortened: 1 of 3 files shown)


pkg/rebuild/pypi/infer.go
```diff
@@ -224,9 +268,44 @@ func inferRequirements(name, version string, zr *zip.Reader) ([]string, error) {
...
+	// case t.After(time.Date(2015, 9, 13, 0, 0, 0, 0, time.UTC)):
+	// 	pythonVersions = append(pythonVersions, "3.5.10") // Python 3.5 release
+	// case t.After(time.Date(2014, 3, 16, 0, 0, 0, 0, time.UTC)):
+	// 	pythonVersions = append(pythonVersions, "3.4.10") // Python 3.4 release
+	default:
+		pythonVersions = append(pythonVersions, "3.12.4") // Default to newest supported version
+	}
+
+	return pythonVersions[0]
+}
```
wbxyz, line 307: It might make sense to move this into the registry directory, we have something similar in some of the other ecosystems, it would be nice to keep them consistent if possible.

https://github.com/google/oss-rebuild/blob/028e28300b20e389762748e432cb67f3957d78aa/pkg/registry/npm/node.go#L33

https://github.com/google/oss-rebuild/blob/028e28300b20e389762748e432cb67f3957d78aa/pkg/registry/cratesio/rust.go#L40

pkg/rebuild/pypi/infer.go
```diff
@@ -187,29 +193,67 @@ func FindPureWheel(artifacts []pypireg.Artifact) (*pypireg.Artifact, error) {
...
+func FindSourceDistribution(artifacts []pypireg.Artifact) (*pypireg.Artifact, error) {
+	for _, r := range artifacts {
+		if strings.HasSuffix(r.Filename, ".tar.gz") {
+			return &r, nil
+		}
+	}
+	return nil, fs.ErrNotExist
+}
+
+func inferRequirements(name, version string, zr interface{}) ([]string, error) {
```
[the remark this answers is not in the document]
wbxyz: Or maybe this could accept a function that just reads files, and it could be independent of zip or tar dependencies?

pkg/rebuild/pypi/infer.go
```diff
@@ -254,10 +334,19 @@ func (Rebuilder) InferStrategy(ctx context.Context, t rebuild.Target, mux rebuil
...
 	}
+	zipFile := true
 	a, err := FindPureWheel(release.Artifacts)
 	if err != nil {
-		return cfg, errors.Wrap(err, "finding pure wheel")
+		zipFile = false
+		a, err = FindSourceDistribution(release.Artifacts)
+		if err != nil {
+			return scfg, errors.Wrap(err, "Failed to find pure wheel and source distribution")
+		}
```
wbxyz, line 344: Are any source distributions a different archive format (besides zip?)

pkg/rebuild/pypi/infer.go
```diff
@@ -235,6 +314,7 @@ func (Rebuilder) InferStrategy(ctx context.Context, t rebuild.Target, mux rebuil
 	}
 	// TODO: support different build types.
 	cfg := &PureWheelBuild{}
+	scfg := &SourceDistributionBuild{}
```
wbxyz, line 317: Do we still need these cfg and scfg variables to be defined at the beginning? I think we construct new ones at the end of this function.

pkg/rebuild/pypi/infer.go
```diff
@@ -365,3 +479,29 @@ func getFile(fname string, zr *zip.Reader) ([]byte, error) {
 	}
 	return nil, fs.ErrNotExist
 }
+
+// getFileFromTarGz extracts a file with the given name from a tar.Reader.
+func getFileFromTarGz(fname string, tr *tar.Reader) ([]byte, error) {
```
wbxyz, line 484: Now that we have both getFile for both .tar.gz and also .zip, maybe `getFile` should be more specific now? Something like `getFileFromZip`?

pkg/rebuild/pypi/strategy.go
```diff
@@ -54,15 +63,75 @@ func (b *PureWheelBuild) GenerateFor(t rebuild.Target, be rebuild.BuildEnv) (reb
 	return b.ToWorkflow().GenerateFor(t, be)
 }
 
+func (b *SourceDistributionBuild) ToWorkflow() *rebuild.WorkflowStrategy {
```
wbxyz, line 66: Any chance you could add some tests for this? It could probably be pretty similar to the TestPureWheelBuild in strategy_test.go (maybe without a timewarp example)

pkg/rebuild/pypi/strategy.go
```diff
@@ -95,10 +164,17 @@ var toolkit = []*flow.Tool{
...
+				Uses: "pypi/setup-python",
+				With: map[string]string{
+					"version": "{{.With.pythonVersion}}",
+				},
+			},
 			{
 				Uses: "pypi/setup-venv",
 				With: map[string]string{
-					"locator": "/usr/bin/",
+					// "locator": "/usr/bin/",
```
wbxyz, line 176: Delete?

pkg/rebuild/pypi/strategy.go
```diff
@@ -54,15 +63,75 @@ func (b *PureWheelBuild) GenerateFor(t rebuild.Target, be rebuild.BuildEnv) (reb
...
+// if ! command -v pyenv &> /dev/null; then
+//
+//		curl https://raw.githubusercontent.com/pyenv/pyenv-installer/master/bin/pyenv-installer | bash
+//		export PATH="/root/.pyenv/bin:$PATH"
+//		eval "$(pyenv init -)"
+//		eval "$(pyenv virtualenv-init -)"
+//	fi
+//
+// /root/.pyenv/bin/pyenv global {{.With.version}}`)[1:],
+// "bash", "clang", "curl", "build-base", "patch", "zip", "zlib-dev", "libffi-dev", "linux-headers", "readline-dev", "openssl", "openssl-dev", "sqlite-dev", "bzip2-dev"
```
wbxyz, line 115: Delete?

Hopefully it's possible to put something like this in the test file, that way it's both documented and testable that we're producing scripts like this. And we'll be forced to update those tests if we change how we implement this :)

pkg/rebuild/pypi/strategy.go
```diff
@@ -54,15 +63,75 @@ func (b *PureWheelBuild) GenerateFor(t rebuild.Target, be rebuild.BuildEnv) (reb
...
+	// Python versions before 3.6 cannot be compiled with openssl 3.x, they require 1.0.x
 	{
+		Name: "pypi/setup-python",
+		Steps: []flow.Step{{
+			Runs: textwrap.Dedent(`
+				if [ ! -d "/root/.pyenv" ]; then
+					if ! command -v curl &> /dev/null; then
+						apk add clang curl build-base patch zip zlib-dev libffi-dev linux-headers readline-dev openssl openssl-dev sqlite-dev bzip2-dev xz-dev
+					fi
+					curl https://raw.githubusercontent.com/pyenv/pyenv-installer/master/bin/pyenv-installer | bash
```
wbxyz, line 128: Do you think the pyenv-installer should be pinned?

I'd maybe lean towards yes but I'm open to suggestions there.

pkg/rebuild/pypi/strategy.go
```diff
@@ -54,15 +63,75 @@ func (b *PureWheelBuild) GenerateFor(t rebuild.Target, be rebuild.BuildEnv) (reb
...
 var toolkit = []*flow.Tool{
+	//  Use clang to avoid issues with openssl
+	// Python versions before 3.6 cannot be compiled with openssl 3.x, they require 1.0.x
 	{
+		Name: "pypi/setup-python",
+		Steps: []flow.Step{{
+			Runs: textwrap.Dedent(`
+				if [ ! -d "/root/.pyenv" ]; then
+					if ! command -v curl &> /dev/null; then
+						apk add clang curl build-base patch zip zlib-dev libffi-dev linux-headers readline-dev openssl openssl-dev sqlite-dev bzip2-dev xz-dev
```
wbxyz, line 126: This is quite a big set of dependencies, are these all required for installing python? I guess I shouldn't be surprised, maybe we should pin version numbers here?

wbxyz:
This is awesome, thanks for adding it! We've been wanting this for a while :)

pkg/rebuild/pypi/infer.go
```diff
@@ -224,9 +263,25 @@ func inferRequirements(name, version string, zr *zip.Reader) ([]string, error) {
 	default:
 		reqs = append(reqs, "setuptools==67.7.2")
 	}
+
 	return reqs, nil
 }
 
+func inferPythonVersion(artifact pypireg.Artifact) string {
```
wbxyz: If we don't need this for the source dist, would you mind splitting this into a separate PR? I think it will be helpful regardless, but worth separating the discussion.

pkg/rebuild/pypi/infer.go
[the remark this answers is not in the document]
wbxyz: I'm still interested to see if we can find a way to avoid this interface{} type. Could we use just a regular reader type?

pkg/registry/pypi/pypi.go
```diff
@@ -137,4 +138,449 @@ func (r HTTPRegistry) Artifact(ctx context.Context, pkg, version, filename strin
...
+	date, err := time.Parse(time.DateOnly, dateStr)
+	if err != nil {
+		panic(errors.Wrapf(err, "parsing date '%s'", dateStr))
+	}
+	return CPythonRelease{Version: sv, Date: date}
+}
+
+// Generated using:
+// curl -s https://www.python.org/api/v2/downloads/release/ | jq -r ' map(select(.is_published==true) | {ver: (.name | capture("(?<v>[0-9]+\\.[0-9]+\\.[0-9]+)").v), date: (.release_date | split("T")[0])}) | sort_by(.ver | split(".") | map(tonumber)) | .[] | "newCPythonRelease(\"" + .ver + "\", \"" + .date + "\"),"
+var PythonReleases = []CPythonRelease{
```
wbxyz: We should probably move this into the same PR as the python version inference?

pkg/rebuild/pypi/strategy.go
giacomobenedetti: Those dependencies were only for using `pyenv`. Even if we end up by using it, we can probably find a smarter way to deal with this!

pkg/rebuild/pypi/strategy.go
giacomobenedetti: If we don't infer the version of python it may be useful to pin it. But I don't have a strong opinion here, for what I saw so far the python version has a limited impact on the build. However, it may be useful from an attestation point of view.

PR #435, opened 2025-04-17
Add example Analyzer

msuozzo:


analyzer/example/analyzer/main.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+func EnqueueInit(ctx context.Context) (*analyzerservice.EnqueueDeps, error) {
+	queue, err := taskqueue.NewQueue(ctx, *taskQueuePath, *taskQueueEmail)
+	if err != nil {
+		return nil, errors.Wrap(err, "creating task queue")
+	}
+	return &analyzerservice.EnqueueDeps{
+		Tracker:  feed.TrackerFromFunc(func(schema.ReleaseEvent) (bool, error) { return true, nil }),
+		Analyzer: urlx.MustParse("/analyze"),
```
wbxyz, line 41: Is this correct? I would have thought we needed the hostname here.

analyzer/example/analyzer/main.go
```diff
@@ -0,0 +1,60 @@
...
+func AnalyzerInit(ctx context.Context) (*analyzerservice.AnalyzerDeps, error) {
+	return &analyzerservice.AnalyzerDeps{
+		BuildRepo:    urlx.MustParse(BuildRepo),
+		BuildVersion: BuildVersion,
+	}, nil
+}
+
+func main() {
+	flag.Parse()
+	http.HandleFunc("/enqueue", api.Handler(EnqueueInit, analyzerservice.Enqueue))
```
wbxyz, line 55: Does this need to be api.Translated?

analyzer/example/analyzerservice/enqueue.go
```diff
@@ -0,0 +1,82 @@
...
+	} else if !tracked {
+		return nil, nil
+	}
+	msg := schema.AnalyzeRebuildRequest{
+		Ecosystem: e.Ecosystem,
+		Package:   e.Package,
+		Version:   e.Version,
+		Artifact:  e.Artifact,
+	}
+	_, err := deps.Queue.Add(ctx, deps.Analyzer.JoinPath("analyze").String(), msg)
```
wbxyz, line 41: The URL in main.go already has analyze?

analyzer/example/deployments/main.tf
```diff
@@ -0,0 +1,317 @@
...
+  crypto_key = google_kms_crypto_key.signing-key.id
+}
+
+## Storage resources
+
+resource "google_project_service" "storage" {
+  service = "storage.googleapis.com"
+}
+resource "google_storage_bucket" "findings" {
+  name                        = "${var.analyzer}-rebuild-analyzer-attestations"
```
wbxyz, line 125: Maybe this should be a variable as well? Or maybe it should use the gcp project name as part of the bucket name? I imagine more than one group might want to run the same analyzer

analyzer/example/deployments/main.tf
```diff
@@ -0,0 +1,317 @@
...
+  depends_on = [google_project_service.storage]
+}
+
+## PubSub
+
+resource "google_project_service" "pubsub" {
+  service = "pubsub.googleapis.com"
+}
+data "google_pubsub_topic" "attestation-topic" {
+  #name       = "projects/${var.host}/topics/oss-rebuild-attestation-topic"
```
wbxyz, line 142: Delete?

analyzer/example/deployments/main.tf
```diff
@@ -0,0 +1,317 @@
...
+}
+
+## Storage resources
+
+resource "google_project_service" "storage" {
+  service = "storage.googleapis.com"
+}
+resource "google_storage_bucket" "findings" {
+  name                        = "${var.analyzer}-rebuild-analyzer-attestations"
+  location                    = "us-central1"
```
wbxyz, line 126: var.location, here and below?

analyzer/example/deployments/main.tf
```diff
@@ -0,0 +1,317 @@
...
+resource "google_storage_bucket_iam_binding" "analyzer-writes-findings" {
+  bucket  = google_storage_bucket.findings.name
+  role    = "roles/storage.objectCreator"
+  members = ["serviceAccount:${google_service_account.analyzer.email}"]
+}
+resource "google_project_iam_binding" "analyzer-runs-workloads-as-others" {
+  project = var.project
+  role    = "roles/iam.serviceAccountUser"
+  members = ["serviceAccount:${google_service_account.analyzer.email}"]
+}
```
wbxyz, line 292: I think this grants access to all accounts, not just the one account calling itself.

analyzer/example/analyzer/main.go
msuozzo: yeah i was assuming this was broken but didn't spend the time to figure out the right design to inject it. turns out it's really, really not fun to do so with terraform so i'm bailing and splitting out the subscriber into a different service.

analyzer/example/analyzerservice/enqueue.go
msuozzo: yep removed from there

analyzer/example/analyzer/main.go
msuozzo: good catch

analyzer/example/deployments/main.tf
msuozzo: oops there actually isn't a location so i'll just remove that one reference.

analyzer/example/deployments/main.tf
msuozzo: yep good call. fixed

analyzer/example/deployments/main.tf
wbxyz: It's not very pressing, I suppose if we run into this issue people can tweak the analyzer name they're using to avoid a naming collision.

wbxyz (approved):

PR #439, opened 2025-04-21
Add LLM function calling support

msuozzo:


internal/llm/chat.go
```diff
@@ -0,0 +1,129 @@
...
+				feedback = fmt.Sprintf("prompt feedback: %+v", resp.PromptFeedback)
+			}
+			return resp, errors.New(feedback)
+		}
+		candidate := resp.Candidates[0]
+		if cm.notify != nil && candidate.Content != nil {
+			cm.notify <- *candidate.Content
+		}
+		if calls := candidate.FunctionCalls(); len(calls) > 0 {
+			currentParts = currentParts[:0]
```
wbxyz, line 105: Why only keep the first of currentParts?

internal/llm/chat.go
```diff
@@ -0,0 +1,129 @@
...
+				if !found {
+					return resp, errors.Errorf("tool implementation not found for function call '%s'", call.Name)
+				}
+				responsePart := genai.Part(implFunc(call.Args))
+				currentParts = append(currentParts, responsePart)
+			}
+			continue
+		} else if candidate.FinishReason == genai.FinishReasonStop {
+			lastResponse = resp
+			return lastResponse, nil // graceful stop
```
wbxyz, line 117: Is lastResponse needed, or can this directly return?

internal/llm/chat.go
```diff
@@ -0,0 +1,129 @@
...
+func (cm *Chat) SendMessage(ctx context.Context, parts ...genai.Part) (*genai.GenerateContentResponse, error) {
+	if len(parts) == 0 {
+		return nil, errors.New("message parts cannot be empty")
+	}
+	currentParts := slices.Clone(parts)
+	var lastResponse *genai.GenerateContentResponse
+	for range cm.maxIter {
+		if cm.notify != nil {
+			cm.notify <- *genai.NewUserContent(currentParts...)
+		}
```
wbxyz, line 88: Is cm.notify expecting the prompt and the response as alternating values on the channel? Should we return the prompt and response together in a struct?

wbxyz (approved):

internal/llm/chat.go
msuozzo: this keeps none of current parts. it reuses the previous capacity of the slice across iterations.

internal/llm/chat.go
msuozzo: ah yeah that's a hold-over from a previous version. fixed

internal/llm/chat.go
msuozzo: imo alternating is fine since the role is present as Content.Role and comparable with llm.ModelRole and llm.UserRole. i'm not sure additional grouping provides a much better consumption experience (or, at the very least, i don't have a usecase that would be improved by that structure yet).

wbxyz: Ahh I didn't realize Content.Role was avialable. Makes sense now, thanks :)

PR #441, opened 2025-04-21
Add the per target explorer view

wbxyz:
This allows cross-run exploration of each package.

To start, there is a new option in the explorer tree, called "View per package". That button opens a table view, showing each target, the date of most recent attempt, and the success. If you select an individual target you can view a history of running that package. For each runid in that target's history, you can view the details, diff, or logs similarly to an individual package from the tree view previously.


wbxyz:
In a followup PR I'm going to refactor the ide library into smaller components. That work should make this slightly cleaner, and will also set us up for an AI agent integration.

tools/ctl/ide/ui.go
```diff
@@ -478,6 +532,209 @@ func (e *explorer) LoadTree() error {
...
+func (e *explorer) SelectTree() {
+	e.container.SwitchToPage(TreePageName)
+}
+
+func populateDetails(details *tview.TextView, node *tview.TreeNode) {
+	r, ok := node.GetReference().(rundex.Rebuild)
+	if !ok {
+		return
+	}
+	type detailsStruct struct {
```
msuozzo, line 544: `struct struct` 🙅

tools/ctl/ide/ui.go
```diff
@@ -478,6 +532,209 @@ func (e *explorer) LoadTree() error {
...
+	log.Println("Loading history for", r.ID())
+	t := r.Target()
+	rebuilds, err := e.dex.FetchRebuilds(e.ctx, &rundex.FetchRebuildRequest{
+		Target: &t,
+		Opts:   e.rundexOpts,
+	})
+	if err != nil {
+		return nil, errors.Wrap(err, "fetching rebuilds for target")
+	}
+	// Sort by runid.
```
msuozzo, line 576: this comment isn't doing much for me

tools/ctl/ide/ui.go
```diff
@@ -702,37 +969,48 @@ func (t *TuiApp) runBenchmark(bench string) {
...
 		if filepath.Ext(path) != ".json" {
 			return nil
 		}
-		name := strings.TrimSuffix(filepath.Base(path), ".json")
-		options.AddItem(name, "", 0, func() {
-			go t.runBenchmark(path)
+		res = append(res, path)
+		return nil
+	})
+	if err != nil {
```
msuozzo, line 988: this case is impossible if the walk cbk never returns err. you might just ignore the return value and remove this then (alternatively, you could log the error and eat it here instead of doing so in the cbk)

tools/ctl/ide/ui.go
```diff
@@ -702,37 +969,48 @@ func (t *TuiApp) runBenchmark(bench string) {
...
+		return nil, errors.Wrap(err, "walking benchmark dir")
+	}
+	return res, nil
+}
+
+func (t *TuiApp) selectBenchmark() (<-chan string, error) {
```
msuozzo, line 994: what about naming this something more like "benchmarkPicker" or "benchmarkSelector"?

tools/ctl/rundex/rundex.go
```diff
@@ -32,6 +34,14 @@ type Rebuild struct {
 	Created time.Time
 }
 
+func (r Rebuild) SuccessEmoji() string {
```
msuozzo, line 37: this is not the right place for display logic. this should live in the ui code.

msuozzo, line 37: "SuccessEmoji" describes a successful state to me. What would be a better name here?

msuozzo (approved):

tools/ctl/ide/ui.go
```diff
@@ -81,6 +85,8 @@ func showModal(app *tview.Application, container *tview.Pages, contents inputCap
 	contents.SetInputCapture(func(event *tcell.EventKey) *tcell.EventKey {
 		if event.Key() == tcell.KeyESC {
 			exitFunc()
+			contents.SetInputCapture(nil)
```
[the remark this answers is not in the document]
wbxyz: Added a comment, I also changed the behavior to restore any pre-existing input capture on the container.

tools/ctl/ide/ui.go
```diff
@@ -103,37 +109,54 @@ func showModal(app *tview.Application, container *tview.Pages, contents inputCap
...
 	e.tree.SetRoot(e.root).SetCurrentNode(e.root)
-	e.container.AddPage("explorer", e.tree, true, true)
+	e.table.SetInputCapture(func(event *tcell.EventKey) *tcell.EventKey {
+		if event.Key() == tcell.KeyESC {
+			e.SelectTree()
+			return nil
+		}
+		return event
+	})
+	e.container.AddPage(TablePageName, e.table, true, false)
```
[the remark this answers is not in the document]
wbxyz: I re-used the pattern from setting up the App.UI further down.

tools/ctl/rundex/rundex.go
wbxyz: I went with verdictAsEmoji

tools/ctl/ide/ui.go
```diff
@@ -103,37 +109,54 @@ func showModal(app *tview.Application, container *tview.Pages, contents inputCap
...
+	table        *tview.Table
+	tree         *tview.TreeView
+	root         *tview.TreeNode
+	rb           *Rebuilder
+	dex          rundex.Reader
+	rundexOpts   rundex.FetchRebuildOpts
+	runs         map[string]rundex.Run
+	buildDefs    rebuild.LocatableAssetStore
+	butler       localfiles.Butler
+	benchmarkDir string
```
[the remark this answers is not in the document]
wbxyz: Added in https://github.com/google/oss-rebuild/pull/451

tools/ctl/ide/ui.go
wbxyz: Resolved in the benchmarks PR

tools/ctl/ide/ui.go
wbxyz: I think those aren't quite right because they're nouns and would more accurately describe the UI component rather than this function which does the action of prompting the user and waiting for the response.

WDYT of "promptForBenchmark"

msuozzo: yeah that's fine

tools/ctl/ide/ui.go
wbxyz: Was copied from elsewhere in this file, I combined the two implementations and fixed the stutter

PR #446, opened 2025-04-23
Make App.Ctx private

wbxyz:


tools/ctl/ide/ui.go
```diff
@@ -486,7 +486,7 @@ type tuiAppCmd struct {
 
 // TuiApp represents the entire IDE, containing UI widgets and worker processes.
 type TuiApp struct {
-	Ctx          context.Context
+	ctx          context.Context
```
msuozzo, line 489: I'm not a fan of having this be present on the struct at all. contexts are really meant to match the lifecycle of the request/action. storing it on an object obscures that relationship and, as it's used here, can violate that original purpose of the context.

can you refactor this out either here or in a follow-on?

msuozzo (approved):

tools/ctl/ide/ui.go
wbxyz: I'm not really sure what the right way to handle contexts in this case is. The app is called with Run(), and while we could pass the context into Run(), the app has multiple callbacks triggered by UI elements and we don't get a context provided when those are called. Should I just use context.Background() in those cases?

Whatever we decide I'd prefer to do any major changes in a separate follow-on after the other ide refactor stuff lands.

msuozzo: yeah I think the requests at those trigger points is better modelled a request originating from the background. if there really was some state to pass down through each request, it should be internal to the app (so, regardless, this change is moving in the right direction).

wbxyz: Sounds good, I'll do that refactor in a followup. It's in line with my broader effort to encapsulate the application logic separate from the UI components. Probably pretty clean for the application logic functions to require a context, while the UI elements in TUI just provide a background context. If we ever switch to a web UI, having per-request contexts on the application logic will be helpful.

PR #451, opened 2025-04-24
Add benchmarks.Repository

wbxyz:


tools/ctl/ide/benchmarks/benchmarks.go
```diff
@@ -0,0 +1,49 @@
...
+			// Best effort reading, skip failures.
+			return nil
+		}
+		if filepath.Ext(path) != ".json" {
+			return nil
+		}
+		all = append(all, path)
+		return nil
+	})
+	if err != nil {
```
msuozzo, line 41: same comment as in other PR: if you suppress all the errors in the Walk callback, Walk can't return an err. I'd remove one or the other handlers

tools/ctl/ide/benchmarks/benchmarks.go
```diff
@@ -0,0 +1,49 @@
...
+		return nil, errors.Wrap(err, "walking dir")
```
msuozzo, line 42: nit: "walking repository"? or "path"? or "fs"? not sure "dir" is the right terminology to use for a fs abstraction.

tools/ctl/ide/benchmarks/benchmarks.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+	return all, nil
+}
+
+func (r *fsRepository) Load(path string) (*benchmark.PackageSet, error) {
```
msuozzo, line 47: comment? what is this supposed to do? or maybe comment on the interface?

msuozzo (approved):

tools/ctl/ide/benchmarks/benchmarks.go
wbxyz: Implementation was in the followup because that's where it was first used. Refactored that here.

tools/ctl/ide/benchmarks/benchmarks.go
wbxyz: After removing the outer error handler this is no longer necessary.

PR #452, opened 2025-04-24
Implement AllOf in Policy

andusy:
Adds AllOf ruleset which requires requests enforced by the policy to satisfy all of the applied rules. This is prioritized over the AnyOf ruleset

### DIFF
--- a/pkg/proxy/policy/policy.go
+++ b/pkg/proxy/policy/policy.go
@@ -28,6 +28,9 @@ func RegisterRule(rulename string, constructor func() Rule) {
 type Policy struct {
 	// AnyOf expects incoming requests to satisfy one of these Rules.
 	AnyOf []Rule `json:"anyOf"`
+
+	// AllOf expects incoming requests to satisfy all of these Rules.
+	AllOf []Rule `json:"allOf"`
 }
 
 // UnmarshalJSON implements the json.Unmarshaler interface for the Policy class.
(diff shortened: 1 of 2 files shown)


pkg/proxy/policy/policy.go
```diff
@@ -75,11 +86,23 @@ func newRuleFromJson(rule json.RawMessage) (Rule, error) {
 // Apply enforces the policy on the request. Returns http.StatusForbidden if the
 // request does not satisfy the policy rules.
 func (p Policy) Apply(req *http.Request, ctx *goproxy.ProxyCtx) (*http.Request, *http.Response) {
+	for _, rule := range p.AllOf {
+		if !rule.Allows(req) {
+			return blockedResponse(req)
+		}
+	}
+	if p.AllOf != nil && (p.AnyOf == nil || len(p.AnyOf) == 0) {
```
msuozzo, line 94: you should only need the len() check:

```
	var foo []string
	fmt.Println(foo == nil, len(foo))
```

```
true 0
```

pkg/proxy/policy/policy.go
```diff
@@ -129,7 +152,7 @@ func (rule URLMatchRule) Allows(req *http.Request) bool {
 		// Avoid matching partial domain names and only match full domain parts.
 		// That is, notgoogle.com must not match google.com, but is.google.com matches google.com.
 		host := rule.Host
-		if !strings.HasPrefix(".", host) {
+		if !strings.HasPrefix(host, ".") {
```
msuozzo, line 155: oof let's highlight this issue with a separate PR. better to have it be obvious in the git blame and changelog. should also add a test to confirm the fixed behavior

msuozzo (approved):

PR #454, opened 2025-04-24
Define rundex.Rebuild commands as data

wbxyz:


tools/ctl/ide/explorer/explorer.go
```diff
@@ -62,37 +62,46 @@ func tmuxWait(cmd string) error {
...
+	tree            *tview.TreeView
+	root            *tview.TreeNode
+	rb              *rebuilder.Rebuilder
+	dex             rundex.Reader
+	rundexOpts      rundex.FetchRebuildOpts
+	runs            map[string]rundex.Run
+	buildDefs       rebuild.LocatableAssetStore
+	butler          localfiles.Butler
+	benches         benchmark.Repository
+	exampleCommands []rebuildCmd
```
msuozzo, line 87: Why "example"? Maybe just "commands"?

msuozzo (approved):

tools/ctl/ide/explorer/explorer.go
wbxyz: In a followup CL I'm trying to wrangle a lot of this functionality out of the explorer UI element, and into a single MCP interface. In that case, there will be a few different types of commands using different arguments. I'd like to keep the variety of commands differentiated for now, but I'm open to renaming this once the MCP refactor is in.

PR #456, opened 2025-04-25
Add npm install flow step

msuozzo:


pkg/rebuild/npm/strategy.go
```diff
@@ -137,16 +137,30 @@ var toolkit = []*flow.Tool{
 			Needs: []string{},
 		}},
 	},
+	{
+		Name: "npm/install",
```
wbxyz, line 141: Can you add a test for this? It would be good to validate the template string does what we expect.

wbxyz (approved):

pkg/rebuild/npm/strategy.go
msuozzo: should now be covered by the existing test suite.

PR #459, opened 2025-04-25
Remove smoketest logic from diffoscope code

wbxyz:
This means we unconditionally stabilize artifacts, even if they're from a smoketest that already stabilized them.

### DIFF
--- a/tools/ctl/ide/explorer/explorer.go
+++ b/tools/ctl/ide/explorer/explorer.go
@@ -206,46 +206,34 @@ func diffArtifacts(ctx context.Context, butler localfiles.Butler, example rundex
 	}
 	var rba, usa string
 	{
-		var err error
-		if example.WasSmoketest() {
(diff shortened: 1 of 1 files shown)


tools/ctl/ide/explorer/explorer.go
```diff
@@ -206,46 +206,34 @@ func diffArtifacts(ctx context.Context, butler localfiles.Butler, example rundex
 	}
 	var rba, usa string
 	{
-		var err error
-		if example.WasSmoketest() {
-			rba, err = butler.Fetch(ctx, example.RunID, example.WasSmoketest(), rebuild.DebugRebuildAsset.For(t))
+		dir, err := os.MkdirTemp("", "*")
```
msuozzo, line 209: i'd mention that we're now stabilizing smoketest resources which, AFAIR, are already stabilized. i think this is fine behavior but I'd like to leave some indication of it on the commit / PR.

msuozzo (approved):

PR #461, opened 2025-04-25
Change butler to call diffoscope when needed

wbxyz:


pkg/rebuild/rebuild/storage.go
```diff
@@ -32,6 +32,8 @@ const (
 	DebugRebuildAsset AssetType = "rebuild"
 	// DebugUpstreamAsset is the upstream artifact we compared against.
 	DebugUpstreamAsset AssetType = "upstream"
+	// DebugDiffAsset captures the diff between rebuild and upstream.
+	DebugDiffAsset AssetType = "diff"
```
msuozzo, line 36: nit: we can move this to colocate it with the diffoscope code (maybe?) or with the butler. this set of assets i think fits best as those used in the service itself and we don't (yet) output diffs. we have some other examples in the codebase of a remote-defined AssetType

msuozzo (approved):

pkg/rebuild/rebuild/storage.go
wbxyz: Moved to diffoscope.

PR #471, opened 2025-04-28
Fix attestation of strategies that use timewarp

msuozzo:
When the strategy requires timewarp, the current workaround to extract the
Location during attestation creation will not work. A dummy URL is provided in
this change to at least get the instructions to render (but remain unused) so
the Location remains accessible.


wbxyz:
This looks good for now, but might be a hint that we should eventually expose the Location directly on the strategy, for two reasons:

1. There might eventually be some strategies that have entirely disjoint BuildEnvs they are valid for, in which case generating instructions with a static BuildEnv won't be possible.
2. We would like the ability to update strategies to use a new location (re-run the old strategy on a new commit, say).

PR #473, opened 2025-04-29
Make truststore detection logic more generic

bentekkie:


pkg/proxy/docker/linux.go
```diff
@@ -6,63 +6,66 @@ package docker
...
+type truststoreFS interface {
+	Exists(path string) bool
+	Read(path string) ([]byte, error)
+}
+
 var distroPattern = re.MustCompile(`\bID=["'']?([^\r\n]+?)["'']?[\r\n]`)
 
 // distro returns the given container's distribution identifier.
-func distro(dfs *dockerfs.Filesystem) (string, error) {
+func distro(dfs truststoreFS) (string, error) {
```
msuozzo, line 20: can this actually be just an fs.FS? Or more likely fs.ReadFileFS (and maybe fs.StatFS or the existence check could be done with an Open returning fs.ErrNotExist). you can consider it beyond the scope of the PR but probably a more sustainable path than rolling our own FS interfaces.

bentekkie:
Abandoning to recreate pr from a branch

PR #481, opened 2025-05-01
Treat HEAD requests like GET requests in the proxy

bentekkie:
This is required to use buildkit behind the proxy as it makes a HEAD request to the docker registry


bentekkie (commented):
Actually never mind I think this is not the issue

PR #485, opened 2025-05-02
Move tmux code to package

wbxyz:


tools/ctl/ide/tmux/tmux.go
```diff
@@ -0,0 +1,43 @@
...
+		return errors.Wrap(err, "opening tmux window")
+	}
+	// Wait to receive the tmux signal.
+	if _, err := exec.Command("tmux", "wait", done).Output(); err != nil {
+		return errors.Wrap(err, "failed to wait for tmux signal")
+	}
+	return nil
+}
+
+func Cat(path string) error {
```
msuozzo, line 41: I might lean towards omitting this since I don't necessarily know what to expect with tmux.Cat

msuozzo (approved):

PR #487, opened 2025-05-02
Log local benchmark execution to the rundex

wbxyz:


tools/ctl/ide/explorer/explorer.go
```diff
@@ -318,7 +318,7 @@ func (e *Explorer) RunBenchmark(ctx context.Context, bench string) {
 		BenchmarkName: filepath.Base(bench),
 		BenchmarkHash: hex.EncodeToString(set.Hash(sha256.New())),
 		Type:          string(schema.SmoketestMode),
-		Created:       ts.UnixMilli(),
+		Created:       ts.Unix(),
```
msuozzo, line 321: this seems like an independent fix. split out?

msuozzo (approved):

tools/ctl/ide/explorer/explorer.go
wbxyz: Done in #488. It ended up being a bit broader than I expected.

msuozzo (approved):

PR #488, opened 2025-05-05
Use native time.Time types in firestore and rundex

wbxyz:
We were inconsistent with timezone and Unix vs UnixMilli previously, causing a lot of bad data in our metadata storage.

After doing some data migrations, firestore contains the data as native time.Time types now.

I also removed the rundex.Rebuild.Created field, because that's now directly handled in embed schema.Attempt

### DIFF
--- a/pkg/rebuild/schema/schema.go
+++ b/pkg/rebuild/schema/schema.go
@@ -257,17 +257,17 @@ type RebuildAttempt struct {
 	RunID           string          `firestore:"run_id,omitempty"`
 	BuildID         string          `firestore:"build_id,omitempty"`
 	ObliviousID     string          `firestore:"oblivious_id,omitempty"`
-	Started         int64           `firestore:"started,omitempty"` // The time rebuild started
-	Created         int64           `firestore:"created,omitempty"` // The time this record was created
+	Started         time.Time       `firestore:"started,omitempty"` // The time rebuild started
+	Created         time.Time       `firestore:"created,omitempty"` // The time this record was created
 }
 
 // Run stores metadata on an execution grouping.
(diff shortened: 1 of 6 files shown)


msuozzo:
I think we should revisit using a structured type here. I believe time.Time has the support we need from firestore and it certainly presents the best programming interface for us. Does that sound feasible? We'll likely need some migration code but it might be worthwhile in the long run.

wbxyz:
Added migration capabilities in #490 and executed them to clean up most of our data. I changed this PR to use native types as well, once this is submitted and deployed I'll need to migrate any entries created in the interval inbetween, then I'll re-enable the index on "created" fields.

pkg/rebuild/schema/schema.go
```diff
@@ -257,17 +257,17 @@ type RebuildAttempt struct {
...
-	ID            string `firestore:"id,omitempty"`
-	BenchmarkName string `firestore:"benchmark_name,omitempty"`
-	BenchmarkHash string `firestore:"benchmark_hash,omitempty"`
-	Type          string `firestore:"run_type,omitempty"`
-	Created       int64  `firestore:"created,omitempty"`
+	ID            string    `firestore:"id,omitempty"`
+	BenchmarkName string    `firestore:"benchmark_name,omitempty"`
+	BenchmarkHash string    `firestore:"benchmark_hash,omitempty"`
+	Type          string    `firestore:"run_type,omitempty"`
+	Created       time.Time `firestore:"created,omitempty"` // The time this run was created
```
msuozzo, line 270: remove comment

msuozzo (approved):

PR #489, opened 2025-05-05
Add command to generate stabilizers from diff

wbxyz:
Using a content summary diff, this command populates the build
definition's CustomStabilizers with PathExclusions for all files
(including those that just have a small diff).

I also added a check to make sure this doesn't re-add paths that have
been excluded.

### DIFF
--- a/tools/ctl/ide/commands/commands.go
+++ b/tools/ctl/ide/commands/commands.go
@@ -19,7 +19,9 @@ import (
 	"strings"
 	"time"
 
+	"github.com/google/oss-rebuild/internal/glob"
 	"github.com/google/oss-rebuild/internal/llm"
+	"github.com/google/oss-rebuild/pkg/archive"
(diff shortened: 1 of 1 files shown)


tools/ctl/ide/explorer/explorer.go
```diff
@@ -157,6 +159,14 @@ func NewExplorer(app *tview.Application, modalFn modalFnType, dex rundex.Reader,
 				}
 			},
 		},
+		{
+			Name: "generate exclusions",
```
msuozzo, line 163: i think we want a clearer top-line description for this. i didn't immediately understand what this meant and its a bit trickier than the `edit and rebuild` case because there's no "rebuild" action. that means the user needs to understand what they're editing and where it should go which I think we need to work on.

wbxyz: How's "generate stabilizers from diff"?

msuozzo (approved):

PR #490, opened 2025-05-05
Add the migrate command to ctl.go

wbxyz:
Currently, the only transform it supports is "int-to-time" but other
transforms can be added in the future.


tools/ctl/ctl.go
```diff
@@ -548,6 +551,111 @@ var listRuns = &cobra.Command{
...
+		if *field == "" {
+			log.Fatal("field not provided")
+		}
+		if *transform == "" {
+			log.Fatal("transform not provided")
+		}
+		errSkip := errors.New("skip")
+		var transformFn func(interface{}) (interface{}, error)
+		switch *transform {
+		case "int-to-time":
```
msuozzo, line 578: i think each migration should actually be schema-aware i.e. hard-coded to do exactly one conversion. additionally, you could number them to ensure they are applied in sequence e.g. 01-native-creation-time

tools/ctl/ctl.go
```diff
@@ -548,6 +551,111 @@ var listRuns = &cobra.Command{
...
+					return nil, errSkip
+				default:
+					return nil, errors.New("unexpected type")
+				}
+			}
+		default:
+			log.Fatalf("Unknown transform: %s", *transform)
+		}
+		q := client.CollectionGroup(*collectionGroup).Query
+		bw := client.BulkWriter(cmd.Context())
```
msuozzo, line 598: for the conversion, I think you'll want to manually batch these to ensure you handle large portions of the writes in parallel without buffering too many in memory.

msuozzo: looking at the code, it looks like it already batches at 20. we'll see if that suffices.

tools/ctl/ctl.go
wbxyz: Done, wdyt of the new format

tools/ctl/ctl.go
```diff
@@ -548,6 +551,110 @@ var listRuns = &cobra.Command{
 	},
 }
 
+type migration struct {
```
msuozzo, line 554: pull out into a migrate.go

tools/ctl/ctl.go
```diff
@@ -548,6 +551,110 @@ var listRuns = &cobra.Command{
...
+	"01-native-creation-time": {
+		collectionGroup: "attempts",
+		transform: func(doc *firestore.DocumentSnapshot) ([]firestore.Update, error) {
+			const field = "created"
+			val, err := doc.DataAt(field)
+			if err != nil && strings.Contains(err.Error(), "no field") {
+				return nil, errSkip
+			} else if err != nil {
+				log.Printf("%+v", doc.Ref.Path)
+				log.Fatal(errors.Wrap(err, "getting field"))
```
msuozzo, line 570: combine

tools/ctl/ctl.go
```diff
@@ -548,6 +551,110 @@ var listRuns = &cobra.Command{
...
+			ag := q.NewAggregationQuery()
+			ag = ag.WithCount("total-count")
+			res, err := ag.Get(cmd.Context())
+			if err != nil {
+				log.Fatal(errors.Wrap(err, "getting count"))
+			}
+			totalV, ok := res["total-count"].(*firestorepb.Value)
+			if !ok {
+				log.Printf("%+v", res)
+				log.Fatalf("Couldn't get total count")
```
msuozzo, line 620: combine

msuozzo (approved):

PR #492, opened 2025-05-05
Create commands package, pull logic out of explorer

wbxyz:


tools/ctl/ide/commands/commands.go
```diff
@@ -0,0 +1,333 @@
...
+	butler      localfiles.Butler
+	asst        assistant.Assistant
+	buildDefs   rebuild.LocatableAssetStore
+	dex         rundex.Reader
+	benches     benchmark.Repository
+	rebuildCmds []RebuildCmd
+	globalCmds  []GlobalCmd
+}
+
+func NewCommander(app *tview.Application, rb *rebuilder.Rebuilder, modalFn modalFnType, butler localfiles.Butler, asst assistant.Assistant, buildDefs rebuild.LocatableAssetStore, dex rundex.Reader, benches benchmark.Repository) (*Commander, error) {
```
msuozzo, line 62: okay so overall, I think this is the right direction but we should split out the commands themselves into a separate constructor so the Commander's state is only the commands themselves. it then acts as a "registry" that can be initialized from ui or explorer while the commands can be more closely associated with the "component" that defines them. (the fudge here is rebuilds which are numerous enough that using real instantiated components for them is performance-prohibitive so special-casing them here is preferable.)

tools/ctl/ide/choice/choice.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/gdamore/tcell/v2"
+	"github.com/google/oss-rebuild/tools/ctl/ide/modal"
+	"github.com/rivo/tview"
+)
+
+const (
+	defaultBackground = tcell.ColorDarkCyan
+)
+
+func Choice(all []string) (modal.InputCaptureable, modal.ModalOpts, <-chan string) {
```
msuozzo, line 16: maybe add a description here

tools/ctl/ide/details/details.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/pkg/errors"
+	"github.com/rivo/tview"
+	"gopkg.in/yaml.v3"
+)
+
+const (
+	defaultBackground = tcell.ColorGray
+)
+
+func String(example rundex.Rebuild) (string, error) {
```
msuozzo, line 23: given this is sort of a component, not thrilled with exposing this as public. any chance the usage in the other file could reuse the textview here? or would there be a better place for this formatting snippet.

msuozzo, line 23: nit: might fit better as "details.Format" or something.

msuozzo (approved):

tools/ctl/ide/details/details.go
wbxyz: Re-using the textview would force the current users of String to add and remove elements to the ui very frequently, and I'm not sure how stable the UI would be with that behavior. If this interface becomes a problem we can revisit in the future.

PR #502, opened 2025-05-09
Add attestation storage docs

msuozzo:


docs/storage.md
```diff
@@ -0,0 +1,73 @@
...
+Attestations can be fetched from GCS using the `gsutil` CLI. For example, you can access Google OSS Rebuild attestations as follows:
+
+```bash
+# Using gsutil
+gsutil cat gs://google-rebuild-attestations/pypi/absl-py/2.0.0/absl_py-2.0.0-py3-none-any.whl/rebuild.intoto.jsonl
+```
+
+## Verification
+
+All attestations are signed using a host-specific KMS key. The `oss-rebuild` CLI can be used to verify attestation signatures:
```
wbxyz, line 64: Mention that requesting attestations are verified by default, and there are options on how to verify?

wbxyz (approved):

PR #509, opened 2025-05-12
Add dynamic refresh of TUI when rundex notifies

wbxyz:


tools/ctl/rundex/rundex.go
```diff
@@ -218,9 +218,15 @@ type FetchRunsOpts struct {
...
 
+type RundexUpdate struct {
+	Rebuild *Rebuild
+	Run     *Run
+}
+
 type Reader interface {
 	FetchRuns(context.Context, FetchRunsOpts) ([]Run, error)
 	FetchRebuilds(context.Context, *FetchRebuildRequest) ([]Rebuild, error)
+	Notify() <-chan RundexUpdate
```
msuozzo, line 229: I don't think this is the right place to put this abstraction. I'd recommend either

1) Adding a notify channel to Opts
2) converting the slice returns to iter returns and have the caller implement a notify channel

tools/ctl/ide/explorer/explorer.go
```diff
@@ -316,6 +352,96 @@ func (e *Explorer) SelectTree() {
...
+
+func (e *Explorer) handleRebuildUpdate(r *rundex.Rebuild) {
+	run, ok := e.runs[r.RunID]
+	if !ok {
+		log.Printf("rebuild update from unknown run %s", r.RunID)
+		return
+	}
+	rg := find(e.root, run.BenchmarkName)
+	if rg == nil {
+		log.Printf("Unable to find rungroup for  %s", run.BenchmarkName)
```
msuozzo, line 411: missing return here and below at 415. I think this means you might want have this return an error and log it at a higher level

tools/ctl/ide/explorer/explorer.go
```diff
@@ -353,24 +479,23 @@ func (e *Explorer) rebuildHistory(rebuilds []rundex.Rebuild) (modal.InputCapture
...
 		})
 		populateDetails := func(node *tview.TreeNode) {
 			if node == root {
 				details.SetText("")
 				return
 			}
-			r, ok := node.GetReference().(*rundex.Rebuild)
-			if !ok {
+			d, ok := node.GetReference().(nodeData)
+			if !ok || len(d.Rebuilds) < 1 {
```
msuozzo, line 494: == 0

tools/ctl/rundex/rundex.go
```diff
@@ -226,6 +226,11 @@ type Writer interface {
 	WriteRun(ctx context.Context, r Run) error
 }
 
+type Watcher interface {
+	WatchRuns() <-chan *Run
```
msuozzo, line 230: I do think these should take a very similar set of constraints to the Fetch* methods on Reader. not critical to implement now but maybe a placeholder param might be worthwhile.

tools/ctl/rundex/rundex.go
```diff
@@ -226,6 +226,11 @@ type Writer interface {
...
+	WatchRebuilds() <-chan *Rebuild
```
msuozzo, line 231: probably need to document the ownership and cleanup semantic of the returned channel. if the impl sends to a closed channel, that will panic.

tools/ctl/ctl.go
```diff
@@ -158,7 +159,9 @@ var tui = &cobra.Command{
 					log.Fatal(err)
 				}
 			} else {
-				dex = rundex.NewLocalClient(localfiles.Rundex())
+				lc := rundex.NewLocalClient(localfiles.Rundex())
+				dex = lc
```
msuozzo, line 163: hmm this strikes me as though it could be modelled as a new compound interface i.e. ReaderWatcher. Although maybe it would make sense for there to be separate readers and watchers for the same underlying storage.

msuozzo (approved):

tools/ctl/ide/explorer/explorer.go
wbxyz: It should probably have been != 1 like the check above.

tools/ctl/rundex/rundex.go
wbxyz: Added a TODO comment that we might want something like that. Currently though, I don't think we have a use case for filtering the Watcher methods. I'll leave the params out for now, if we do add params in the future it shouldn't be hard to find existing callsites and add the params that make sense there.

tools/ctl/rundex/rundex.go
wbxyz: Added.

tools/ctl/ctl.go
wbxyz: The problem is that the firestore rundex does not implement Watcher, so I can't just make `dex` be a `ReaderWatcher`.

Also, this gives us the freedom to easily decouple local reader and local watcher in the future.

PR #511, opened 2025-05-12
Move truncated message to beginning

wbxyz:
The truncation behavior is to take the last X bytes, so really the "...truncated..." message should go at the start not the end.


tools/ctl/ide/assistant/session.go
msuozzo, line 1: nit: you could also include the number of bytes removed, if you think it'd be useful.

msuozzo (approved):

tools/ctl/ide/assistant/session.go
wbxyz: Nice, I think that's a good idea.

PR #519, opened 2025-05-15
Detect non-zero exit status codes

msuozzo:


tools/agent/main.go
```diff
@@ -315,7 +315,7 @@ From the following llm response, extract only the shell commands to be run and e
 				log.Fatalf("Failed to stabilize upstream artifact: %v", err)
 			}
 			diffOutput, err := runDiffoscope(rb, up)
-			if err != nil {
+			if _, ok := err.(*exec.ExitError); err != nil && !ok {
```
wbxyz, line 318: Isn't this already handled on 413?

I guess if it's something other than ExitCode() 1, that would make its way up here. It might be cleaner to push any exec.ExitError decision making into runDiffoscope(). WDYT?

wbxyz (approved):

tools/agent/main.go
msuozzo: ah yeah. i expected two chunks here but the other was split off into the build failure recovery PR

PR #520, opened 2025-05-15
Add build failure recovery

msuozzo:


tools/agent/main.go
```diff
@@ -297,6 +297,47 @@ From the following llm response, extract only the shell commands to be run and e
 			log.Fatalf("Failed to generate strategy: %v", err)
 		}
 		err = execute(ctx, tg, *strat, store)
+		if _, ok := err.(*exec.ExitError); ok {
+			var response genai.Content
+			{
+				logsReader, err := store.Reader(ctx, rebuild.DebugLogsAsset.For(tg))
```
wbxyz, line 303: Should there be some cleanup of the logs and rebuild artifact between build attempts?

My first thought was that this should make a new store each time, but maybe we want to re-use the upstream artifact if it exists? Although looking further down the code I don't think we check to see if things exist before fetching new upstream or doing new stabilize steps.

WDYT?

wbxyz (approved):

tools/agent/main.go
msuozzo: if this step succeeds, we ignore the build output but the logs are guaranteed to be present.

if the diffoscope fails, we are guaranteed to have new build output and logs.

As for the upstream, additional fetches on each iteration should be optimized away but still don't compromise correctness.

wbxyz: In some incarnation of the agent I suspect a human will want to look back at the different attempts and understand what was happening, which isn't possible if we overwrite every time. However I don't think that's necessary for this PR.

It might also be nice to standardize the access in a way that TUI and other tools can read or re-use data the agent is producing. That could be some future PR though.

msuozzo: yeah for sure. we definitely want a sort of "history of experiments" but i think that's out of scope for now.

PR #522, opened 2025-05-17
refactor: add `Stabilize(any)` method to interface

### LINKED ISSUE #425: `Stabilizer` aliasing `any` prevents meaningful type checking
Notably, `[]Stabilizer` will accept the slice as an element: 

```
type Foo any

func main() {
	s := []Foo{0}
	fmt.Printf("%v\n", s)
	fmt.Printf("%v\n", append(s, 1))
	fmt.Printf("%v\n", append(s, []Foo{1}))
	fmt.Printf("%v\n", append(s, []Foo{1}...))
}
```

prints...

```
[0]
[0 1]
[0 [1]]
[0 1]
```

See https://github.com/google/oss-rebuild/pull/363#discussion_r2037874082

algomaster99:
Fixes #425 

The changes mark `Stabilizer` as an interface. For each of its implementations,  I added a method `Stabilize` for example
```go
func (g GzipStabilizer) Stabilize(arg any) {
	if g.Func != nil {
		g.Func(arg.(*MutableGzipHeader))
	}
}
```
Now instead of `Func`, `Stabilize` is invoked everywhere. We could also make `Func` and `Name` fields private but I let them be for now.

### DIFF
--- a/pkg/archive/gzip.go
+++ b/pkg/archive/gzip.go
@@ -48,9 +48,6 @@ type GzipStabilizer struct {
 }
 
 func (g GzipStabilizer) Stabilize(arg any) {
-	if g.Func == nil {
-		panic(g.Name + " stabilizer not implemented")
-	}
 	g.Func(arg.(*MutableGzipHeader))
 }
 
--- a/pkg/archive/gzip.go
+++ b/pkg/archive/gzip.go
@@ -48,9 +48,10 @@ type GzipStabilizer struct {
 }
 
 func (g GzipStabilizer) Stabilize(arg any) {
-	if g.Func != nil {
-		g.Func(arg.(*MutableGzipHeader))
+	if g.Func == nil {
+		panic(g.Name + " stabilizer not implemented")
 	}
+	g.Func(arg.(*MutableGzipHeader))
 }
 
 var AllGzipStabilizers = []Stabilizer{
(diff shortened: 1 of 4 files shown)


algomaster99:
@msuozzo ping for review

pkg/archive/gzip.go
```diff
@@ -47,6 +47,12 @@ type GzipStabilizer struct {
 	Func func(*MutableGzipHeader)
 }
 
+func (g GzipStabilizer) Stabilize(arg any) {
+	if g.Func != nil {
```
msuozzo, line 51: i'd probably just assume Func is set and have a "loud" outcome (i.e. a crash) if there's a probable misconfiguration rather than a "quiet" one where the entry is skipped.

msuozzo:
Yeah I think this approach will work. It's unfortunate that Stabilize(any) has such a brittle interface but we still do get some protection beyond the status quo.

msuozzo (approved):

algomaster99:
>  It's unfortunate that Stabilize(any) has such a brittle interface but we still do get some protection beyond the status quo.

Usage of generics in go is not very clear to me. It seems more like a [concatenation of types](https://go.dev/doc/tutorial/generics) rather than declaring any type you would want the method to accept. In other words, we would need to know all types that `Stabilize` can accept which is not a good design, hence `any`.

pkg/archive/tar.go
```diff
@@ -46,11 +46,25 @@ type TarArchiveStabilizer struct {
 	Func func(*TarArchive)
 }
 
+func (t TarArchiveStabilizer) Stabilize(arg any) {
+	if t.Func == nil {
```
msuozzo, line 50: ah I just meant let it panic when you try to call nil on line 53. the extra code likely doesn't cut appreciable time off of the debug process.

algomaster99: Cool. Made the change. `panic` statement just made it more explicit where the error is but the stack trace is also  clear.

msuozzo (approved):

PR #529, opened 2025-05-30
Revise attestation bundle API with filtering

msuozzo:
The new filter API allows accessing user-defined statement types from a bundle.
Similarly, this change reemoves user-specific attestation logic from the bundle
API.


pkg/attestation/filter.go
```diff
@@ -0,0 +1,95 @@
...
+	data, err := json.Marshal(from)
+	if err != nil {
+		return nil, err
+	}
+	var result T
+	if err := json.Unmarshal(data, &result); err != nil {
+		return nil, err
+	}
+	return &result, nil
+}
```
wbxyz: Would this be useful outside filtering in its own right? If so, maybe it should be in bundle.go? If not, maybe it should be private?

wbxyz (approved):

pkg/attestation/filter.go
msuozzo: yeah good call. made it private.

PR #549, opened 2025-06-05
Support GCB private pools

msuozzo:


internal/gcb/gcb.go
```diff
@@ -25,21 +33,44 @@ type Client interface {
...
+			build.Options = &cloudbuild.BuildOptions{}
+		}
+		build.Options.Pool = &cloudbuild.PoolOption{
+			Name: c.privatePoolConfig.Name,
+		}
+	}
+	// For private pools, we need to use the regional API endpoint if specified
+	if c.privatePoolConfig != nil && c.privatePoolConfig.Region != "" {
+		return c.service.Projects.Locations.Builds.Create(fmt.Sprintf("projects/%s/locations/%s", project, c.privatePoolConfig.Region), build).Context(ctx).Do()
+	}
```
wbxyz: Is the locaitons endpoint only valid for private pools? Or should this be an option for all build creations?

In the converse, is the location *required* for private pools? If so, should we raise an error if the pool name is specified but not the region?

msuozzo: it's not only valid for private pools but globally-scoped requests have an implicit global region. regardless of the current behavior of GCB, I think this faithfully fulfills what its API promises. Added some validation around Name, though, since we definitely need that to be set.

PR #556, opened 2025-06-10
Fix terraform output dependency

wbxyz:
Changing the outputs makes sure that terraform will wait for the images
to be built before proceeding.

### DIFF
--- a/deployments/modules/container_build_push/main.tf
+++ b/deployments/modules/container_build_push/main.tf
@@ -25,7 +25,7 @@ locals {
   build_args_str = join(" ", [for arg in var.build_args : "--build-arg ${arg}"])
 }
 
-resource "terraform_data" "image" {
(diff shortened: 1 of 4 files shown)


deployments/modules/container_build_push/main.tf
```diff
@@ -25,7 +25,7 @@ locals {
   build_args_str = join(" ", [for arg in var.build_args : "--build-arg ${arg}"])
 }
 
-resource "terraform_data" "image" {
+resource "terraform_data" "image_inputs" {
```
msuozzo: nit: maybe "image_deps"?

msuozzo: we should also update container_binary_upload main.tf to use the same terraform_data pattern

msuozzo (approved):

PR #561, opened 2025-06-11
Create local dev script for docs

msuozzo:


docs/_config.yml
```diff
@@ -4,3 +4,7 @@
 remote_theme: pages-themes/minimal@v0.2.0
 plugins:
   - jekyll-remote-theme
+exclude:
```
wbxyz, line 7: Should this also exclude Gemfile?

wbxyz (approved):

docs/_config.yml
msuozzo: It's already excluded natively by GitHub pages. This is just a list for our project.

PR #567, opened 2025-06-11
Change oss-rebuild commands to RunE

wbxyz:


cmd/oss-rebuild/main.go
```diff
@@ -57,9 +57,14 @@ var getCmd = &cobra.Command{
 	Long: `Get rebuild attestation for a specific ecosystem/package/version/artifact.
 The ecosystem is one of npm, pypi, or cratesio. For npm the artifact is the <package>-<version>.tar.gz file. For pypi the artifact is the wheel file. For cratesio the artifact is the <package>-<version>.crate file.`,
 	Args: cobra.MinimumNArgs(3),
-	Run: func(cmd *cobra.Command, args []string) {
+	// Silence errors because we want to pretty-print them.
+	SilenceErrors: true,
```
msuozzo, line 62: nit: I think this should be removed here and then enabled in the child PR

PR #575, opened 2025-06-16
Create new build abstraction

msuozzo:
The upshot is we can split and reuse methods of executing builds from the way
we compose the builds themselves. This should give us cleaner abstractions to
express our existing builds as well as allow for greater reuse in developing
future build types.

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -1,0 +1,78 @@
+// Copyright 2025 Google LLC
+// SPDX-License-Identifier: Apache-2.0
+
+package build
+
+import (
+	"context"
+	"io"
+
+	"github.com/google/oss-rebuild/pkg/rebuild/rebuild"
+)
+
+// ToolType represents types of tools used during builds
+type ToolType string
+
+const (
+	TimewarpTool ToolType = "timewarp"
+	ProxyTool    ToolType = "proxy"
+	GSUtilTool   ToolType = "gsutil_writeonly"
+)
+
+// Executor manages build execution for a specific backend
+type Executor interface {
+	Start(ctx context.Context, input rebuild.Input, opts Options) (Handle, error)
+	Status() ExecutorStatus
+	Close(ctx context.Context) error
+}
+
+// Handle represents an active or completed build
+type Handle interface {
+	BuildID() string
+	Wait(ctx context.Context) (Result, error)
+	OutputStream() io.Reader
+	Status() BuildState
+}
+
+// Result represents the completed build result
+type Result struct {
+	// Error represents a build-time failure (i.e. after build setup)
+	Error error
+	// Timings describes execution duration of various aspects of the build
+	Timings rebuild.Timings
+}
+
+// ExecutorStatus represents the overall executor status
+type ExecutorStatus struct {
+	// InProgress is the number of builds currently executing
+	InProgress int
+	// Capacity is the max number of builds that can be exected simultanously
+	Capacity int
+	// Healthy is whether the executor is accepting new builds
+	Healthy bool
+}
+
+// BuildState represents the current state of a build
+type BuildState int
+
+const (
+	BuildStateStarting BuildState = iota
+	BuildStateRunning
+	BuildStateCompleted
+	BuildStateCancelled
+)
+
+func (s BuildState) String() string {
+	switch s {
+	case BuildStateStarting:
+		return "starting"
+	case BuildStateRunning:
+		return "running"
+	case BuildStateCompleted:
+		return "completed"
+	case BuildStateCancelled:
+		return "cancelled"
+	default:
+		return "unknown"
+	}
+}
--- a/pkg/build/options.go
+++ b/pkg/build/options.go
@@ -1,0 +1,80 @@
+// Copyright 2025 Google LLC
+// SPDX-License-Identifier: Apache-2.0
+
+package build
+
+import (
+	"time"
+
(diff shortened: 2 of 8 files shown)


pkg/build/build.go
```diff
@@ -0,0 +1,73 @@
...
+// Result represents the completed build result
+type Result struct {
+	Error   error
+	Timings rebuild.Timings
+}
+
+// ExecutorStatus represents the overall executor status
+type ExecutorStatus struct {
+	InProgress int
+	Capacity   int
```
wbxyz: RemainingCapacity?
TotalCapacity?

pkg/build/options.go
```diff
@@ -0,0 +1,79 @@
...
+	// Resources configures URLs and authentication for build resources
+	Resources Resources
+}
+
+// CancelPolicy determines how build cancellation is handled
+type CancelPolicy int
+
+const (
+	// CancelImmediate terminates the build immediately
+	CancelImmediate CancelPolicy = iota
```
wbxyz, line 62: WDYT of the zero-th element being CancelUnspecified, with default behavior being defined at the executor level?

pkg/build/options.go
```diff
@@ -0,0 +1,79 @@
...
+	// BaseImageConfig defines the selection criteria for base images
+	BaseImageConfig BaseImageConfig
+}
+
+// Options configures build execution behavior
+type Options struct {
+	// CancelPolicy determines how cancellation is handled
+	CancelPolicy CancelPolicy
+	// Timeout for the build execution
+	Timeout time.Duration
```
wbxyz, line 30: Should this be a deadline instead of a timeout?

pkg/build/url.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+	u, err := url.Parse(gsURL)
+	if err != nil {
+		return "", fmt.Errorf("invalid gs:// URL: %w", err)
+	}
+	bucket := u.Host
+	object := strings.TrimPrefix(u.Path, "/")
+	httpURL := fmt.Sprintf("https://storage.googleapis.com/%s/%s", bucket, object)
+	return httpURL, nil
+}
```
wbxyz: Would this make sense in internal/gcb/?

pkg/build/build.go
msuozzo: `cap()` is a golang builtin meaning "capacity" and that means the total size of a data structure (versus `len()` which would be occupied capacity). I think the parallel in the stdlib means user expectation likely aligns with its semantics so i'm inclined to keep it.

pkg/build/url.go
msuozzo: nah it's not a GCB function since it's a property of the GCS API. I can see it as part of internal/uri but I don't feel strongly one way or the other.

pkg/build/options.go
msuozzo: i feel this is less important for a constant within our codebase. since we're not particularly concerned with distributed services figuring out how to interact over time (as is the case with the guidance around proto enums), I think it's preferable to reduce the undefined/implementation-defined behavior.

pkg/build/options.go
msuozzo: It could be. I leaned towards timeout since that's the field present in GCB's API. I think either could work, though.

pkg/build/url.go
wbxyz: Yeah I think internal/uri makes sense

pkg/build/options.go
wbxyz: I was leaning towards deadline in my code because we're bound by the lifetime of the top level handler, and timeouts don't account for any of the intermediate processing time before executing the build. I'd weakly vote for deadlines as far into our interfaces as we can, only switching to timeouts at the interface with GCB.

pkg/build/options.go
wbxyz: I was actually thinking it might be helpful for the caller to delegate cancellation policy to the implementation, and that different implementations might prefer different cancellation policies, but from your comment it sounds like thats intentionally avoided.

pkg/build/build.go
wbxyz: I see what you're saying, that an unqualified "Capacity" should be clear as total capacity based on the standard library. When I was reading it though I did actually interpret it the other way which is why that was my first suggestion.

Is this something that might get serialized to json at some point? Maybe just adding a comment would be enough?

pkg/build/options.go
msuozzo: i think it's only useful if the implementations have particular trouble with certain policies. I'm imagining that this isn't the case and so the consistency we gain exceeds the benefit of delegation.

pkg/build/url.go
msuozzo: eh i've made it private since it's really just an implementation detail of the conversion function.

pkg/build/build.go
msuozzo: added a comment

pkg/build/options.go
msuozzo: okay I think I also have a weak vote for Deadline but I think it's a small enough difference that we can defer it so i've added a TODO

pkg/build/url.go
wbxyz: Can the other URL helpers be private too?

wbxyz (approved):

PR #577, opened 2025-06-16
Add `docker run` build type

msuozzo:
This is the simplest build variant we currently use as it comprises a single
script. The general abstractions here will be reused in the more rigorous
`docker build` implementation which is forthcoming.


pkg/build/local/command.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+	if opts.Output != nil {
+		cmd.Stdout = opts.Output
+		cmd.Stderr = opts.Output
+	}
+	// Block and wait for completion.
+	return cmd.Run()
+}
+
+// LookPath implements CommandExecutor
```
wbxyz, line 52: Is there more detail this comment could add? Or maybe it should be LookPath implements CommandExecutor.LookPath?

pkg/build/local/command_options_test.go
```diff
@@ -0,0 +1,187 @@
...
+	"fmt"
+	"strings"
+	"testing"
+)
+
+func TestCommandExecutorOptions(t *testing.T) {
+	executor := NewRealCommandExecutor()
+	ctx := context.Background()
+
+	t.Run("Basic mode with output discarded", func(t *testing.T) {
```
wbxyz, line 18: Should each of these be a separate top-level test function?

As is, it's not clear if they depend on re-using state in the executor and ctx.

pkg/build/local/command_options_test.go
```diff
@@ -0,0 +1,187 @@
...
+
+		if err != nil {
+			t.Fatalf("Expected no error, got: %v", err)
+		}
+
+		output := strings.TrimSpace(buf.String())
+		if output != "combined test" {
+			t.Errorf("Expected 'combined test', got '%s'", output)
+		}
+	})
```
wbxyz: Are these tests the same?

pkg/build/local/run_executor.go
```diff
@@ -0,0 +1,231 @@
...
+		maxParallel:  maxParallel,
+		semaphore:    make(chan struct{}, maxParallel),
+		dockerCmd:    dockerCmd,
+		cmdExecutor:  cmdExecutor,
+		activeBuilds: syncx.Map[string, *localHandle]{},
+	}, nil
+}
+
+// NewDockerRunExecutorWithConfig creates a new Docker run executor with configuration
+func NewDockerRunExecutorWithConfig(config DockerRunExecutorConfig) (*DockerRunExecutor, error) {
```
wbxyz: Why not take DockerRunExecutorConfig in the other two constructors?

pkg/build/local/run_executor.go
```diff
@@ -0,0 +1,231 @@
...
+func (e *DockerRunExecutor) Start(ctx context.Context, input rebuild.Input, opts build.Options) (build.Handle, error) {
+	buildID := opts.BuildID
+	if buildID == "" {
+		buildID = fmt.Sprintf("docker-run-%d", time.Now().UnixNano())
+	}
+	planOpts := build.PlanOptions{
+		UseTimewarp:            opts.UseTimewarp,
+		UseNetworkProxy:        opts.UseNetworkProxy,
+		UseSyscallMonitor:      opts.UseSyscallMonitor,
+		PreferPreciseToolchain: true,
```
wbxyz: Can you add a comment describing why `true` is always used?

pkg/build/local/run_executor.go
```diff
@@ -0,0 +1,231 @@
...
+		for len(e.semaphore) > 0 {
+			time.Sleep(100 * time.Millisecond)
+		}
+		close(done)
+	}()
+	select {
+	case <-done:
+		return nil
+	case <-ctx.Done():
+		return ctx.Err()
```
wbxyz: Wrap with a note that the builds couldn't be cleaned up?

A benefit of wrap is that it adds to the stacktrace IIUC

pkg/build/local/run_executor.go
```diff
@@ -0,0 +1,231 @@
...
+func (e *DockerRunExecutor) executeBuild(ctx context.Context, handle *localHandle, plan *DockerRunPlan, input rebuild.Input, opts build.Options) {
+	defer func() {
+		e.activeBuilds.Delete(handle.id)
+	}()
+	// Acquire semaphore slot
+	select {
+	case e.semaphore <- struct{}{}:
+		defer func() { <-e.semaphore }()
+	case <-ctx.Done():
+		handle.setResult(build.Result{
```
wbxyz, line 167: Do these early exits need to call handle.updateStatus?

pkg/build/local/run_planner.go
```diff
@@ -0,0 +1,123 @@
...
+type dockerRunScriptArgs struct {
+	Inst           rebuild.Instructions
+	PackageManager build.PackageManagerCommands
+	UseTimewarp    bool
+	TimewarpURL    string
+	TimewarpAuth   bool
+}
+
+// dockerRunScriptTpl is the template for the script executed in the container
+var dockerRunScriptTpl = template.Must(
```
wbxyz, line 31: Does this script call docker run? Maybe it should be called "buildScript"?

pkg/build/local/run_executor.go
```diff
@@ -0,0 +1,231 @@
...
+		})
+		return
+	}
+	defer func() {
+		if err := os.RemoveAll(hostOutputPath); err != nil {
+			log.Printf("Failed to clean up temporary directory %s: %v", hostOutputPath, err)
+		}
+	}()
+	// Compose command args
+	runArgs := []string{"run", "--rm"} // Always remove container
```
wbxyz: Can removing the container be configurable? I've been considering container management at the TUI level, to allow inspecting why a build failed.

pkg/build/local/run_executor.go
wbxyz, line 1: Could you add some tests for this file? At least the executeBuild() function seems worth covering.

pkg/build/local/command.go
msuozzo: This is the same sort of docstring as in python that points to the interface. I don't think adding .LookPath will add any clarity. i think pointing out the parts of exec we're emulating in that interface would be good but that would be done in the base PR.

pkg/build/local/command_options_test.go
msuozzo: it's not a stateful interface so I think this is important

pkg/build/local/run_executor.go
msuozzo: yeah we should note that it's a legacy flag only non-true in smoketest

pkg/build/local/run_planner.go
msuozzo: It's the script used in docker run. the word "build" might be confused with docker build. the comment also clarifies the purpose well.

pkg/build/local/run_executor.go
msuozzo: yeah i'd be ok with that feature. i don't think we need it for worker in any of our current use (i think that would be the `docker build` mode instead) so I'm inclined to leave it as a TODO.

pkg/build/local/run_executor.go
msuozzo: oops! they were written but were in another change. done

pkg/build/local/command_options_test.go
msuozzo: oops yeah. output shouldn't have been in the first one

pkg/build/local/run_executor.go
msuozzo: eh actually it was easy enough to add. done

wbxyz (approved):

pkg/build/local/run_planner.go
wbxyz: Wdyt of rebuildScript?

pkg/build/local/command_options_test.go
wbxyz: Because it's not stateful, shouldn't we make that clear by destroying the executor and context between each test?

PR #578, opened 2025-06-16
Create BufferedPipe abstraction for log streaming

msuozzo:
This construct gives us a number of nice properties in implementing the build output streaming. For one, the ring buffer allows us to bound the memory cost of logs streaming for each build. Second, the line-based buffering helps us ensure logs remain human-readable even when eviction is necessary (mid-line breaks don't happen). Additionally, the use of the pipe means that the build invoker need not call `OutputStream()` to drain the output if they don't need it. And if they are interested in observing some failure after the build completes, the buffer is in a state to provide the last portion of the build output (although we currently close and free the buffer on termination).

### DIFF
--- a/internal/bufiox/linebuffer.go
+++ b/internal/bufiox/linebuffer.go
@@ -1,0 +1,201 @@
+// Copyright 2025 Google LLC
+// SPDX-License-Identifier: Apache-2.0
+
+package bufiox
+
+import (
+	"bytes"
+	"errors"
+	"sync"
+)
+
+// LineBuffer is a thread-safe ring buffer for lines of text.
+// Its capacity is defined in bytes, and it operates on a single underlying
+// byte slice to minimize memory allocations. It implements io.ReadWriter.
+type LineBuffer struct {
+	buf           []byte // The circular byte buffer.
+	capacity      int    // The total size of buf.
+	mu            sync.Mutex
+	size          int   // The number of bytes currently in use.
+	head          int   // The read pointer (start of the oldest line).
+	tail          int   // The write pointer (next available write position).
+	lineLengths   []int // A queue of line lengths, corresponding to data in buf.
+	pendingLength int   // The length of the unfinished line at the end of the buf.
+}
+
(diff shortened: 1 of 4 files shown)


internal/bufiox/linebuffer.go
```diff
@@ -0,0 +1,348 @@
...
+	// Read from the snapshot.
+	n, err = b.reader.Read(p)
+	if err == io.EOF {
+		// The snapshot has been fully read. Invalidate it so the next
+		// Read call will create a fresh one. We also return io.EOF to the caller.
+		b.reader = nil
+	}
+	return n, err
+}
+*/
```
wbxyz: Is this an old version? If so, can we delete it?

internal/bufiox/linebuffer.go
```diff
@@ -0,0 +1,348 @@
...
+)
+
+// LineBuffer is a thread-safe ring buffer for lines of text.
+// Its capacity is defined in bytes, and it operates on a single underlying
+// byte slice to minimize memory allocations. It implements io.ReadWriter.
+type LineBuffer struct {
+	buf           []byte // The circular byte buffer.
+	capacity      int    // The total size of buf.
+	mu            sync.Mutex
+	size          int   // The number of bytes currently in use.
```
wbxyz, line 19: Can this just be calculated from head and tail as a receiver method? Maybe Len()?

wbxyz: Oh Size() already exists, I guess it could just be a private size() that doesn't have mutex locks

wbxyz (approved):

internal/bufiox/linebuffer.go
msuozzo: Changed to Len. going through the impl, size currently serves to differentiate the full vs empty cases when tail == head. with the current variable semantics, we _could_ change size to a `full` bool but I don't love how that would change the logic. i like the semantics of head and tail so I'm inclined to leave it as-is and make sure the three locations we update pointers change both.

PR #580, opened 2025-06-16
Add `docker build` build type

msuozzo:
This is a local variant of the primary hosted build, although it does not
currently support the network proxy or the syscall monitor. The general
abstractions here are shared with the `docker run` build type but differ in
that this one is more faithful to the actual rebuild.


pkg/build/local/build_executor.go
```diff
@@ -0,0 +1,307 @@
...
+	outputDir := config.OutputDir
+	if outputDir == "" {
+		outputDir = "/tmp/oss-rebuild"
+	}
+	tempBase := config.TempDirBase
+	if tempBase == "" {
+		tempBase = os.TempDir()
+	}
+	// Check if docker is available
+	dockerCmd := "docker"
```
wbxyz, line 68: Should this also be a config?

pkg/build/local/build_executor.go
```diff
@@ -0,0 +1,307 @@
...
+func (e *DockerBuildExecutor) Start(ctx context.Context, input rebuild.Input, opts build.Options) (build.Handle, error) {
+	buildID := opts.BuildID
+	if buildID == "" {
+		buildID = fmt.Sprintf("docker-build-%d", time.Now().UnixNano())
+	}
+	planOpts := build.PlanOptions{
+		UseTimewarp:            opts.UseTimewarp,
+		UseNetworkProxy:        opts.UseNetworkProxy,
+		UseSyscallMonitor:      opts.UseSyscallMonitor,
+		PreferPreciseToolchain: true,
```
wbxyz: Copy the comment from docker run?

pkg/build/local/build_executor.go
```diff
@@ -0,0 +1,307 @@
...
+			Error: errors.Wrap(err, "docker build failed"),
+		})
+		return
+	}
+	// Generate host output path and create the directory.
+	hostOutputPath := filepath.Join(e.tempDirBase, fmt.Sprintf("oss-rebuild-%s", handle.id))
+	if err := os.MkdirAll(hostOutputPath, 0755); err != nil {
+		handle.setResult(build.Result{
+			Error: errors.Wrap(err, "failed to create output directory"),
+		})
```
wbxyz: For these early returns, should they also call handle.updateStatus(build.BuildStateCancelled)?

pkg/build/local/build_executor.go
```diff
@@ -0,0 +1,307 @@
...
+	// Set final successful result.
+	handle.updateStatus(build.BuildStateCompleted)
+	handle.setResult(build.Result{Error: nil})
+}
+
+// uploadAssets uploads build artifacts to the asset store.
+func (e *DockerBuildExecutor) uploadAssets(ctx context.Context, plan *DockerBuildPlan, hostOutputPath string, input rebuild.Input, opts build.Options, imageTag string, logs string) {
+	store := opts.Resources.AssetStore
+	if store == nil {
+		return
```
wbxyz, line 250: Is it worth logging that the AssetStore is missing? I could imagine that being a frustrating bug to happen silently.

pkg/build/local/build_executor.go
```diff
@@ -0,0 +1,307 @@
...
+		})
+		return
+	}
+	// Set final successful result.
+	handle.updateStatus(build.BuildStateCompleted)
+	handle.setResult(build.Result{Error: nil})
+}
+
+// uploadAssets uploads build artifacts to the asset store.
+func (e *DockerBuildExecutor) uploadAssets(ctx context.Context, plan *DockerBuildPlan, hostOutputPath string, input rebuild.Input, opts build.Options, imageTag string, logs string) {
```
wbxyz: Maybe use the buffer type for `logs` argument to avoid extra copies of a potentially really large string?

pkg/build/local/build_executor.go
```diff
@@ -0,0 +1,307 @@
...
+	defer file.Close()
+	writer, err := store.Writer(ctx, asset)
+	if err != nil {
+		return errors.Wrap(err, "failed to get asset store writer")
+	}
+	defer writer.Close()
+	if _, err := io.Copy(writer, file); err != nil {
+		return errors.Wrap(err, "failed to upload file to asset store")
+	}
+	return nil
```
wbxyz, line 291: Maybe re-use uploadContent?

```go
	content, err := os.ReadFile(filePath)
	if err != nil {
		return errors.Wrapf(err, "failed to open file %s", filePath)
	}
    return e.uploadContent(ctx, store, asset, content)
```

pkg/build/local/build_executor_test.go
```diff
@@ -0,0 +1,565 @@
...
+
+// Test case structure
+type buildTestCase struct {
+	name             string
+	plan             *DockerBuildPlan
+	planError        error
+	input            rebuild.Input
+	options          build.Options
+	maxParallel      int
+	dockerCmd        string
```
wbxyz: Does this get used in the test?

pkg/build/local/build_planner.go
```diff
@@ -0,0 +1,154 @@
...
+		"indent": func(s string) string { return strings.ReplaceAll(s, "\n", "\n ") },
+		"join":   func(sep string, s []string) string { return strings.Join(s, sep) },
+	}).Parse(
+		textwrap.Dedent(`
+			#syntax=docker/dockerfile:1.10
+			FROM {{.BaseImage}}
+			RUN {{if or .TimewarpAuth .ProxyAuth}}--mount=type=secret,id=auth_header {{end}}<<'EOF'
+			 set -eux
+			{{- if .UseTimewarp}}
+			 curl {{if .TimewarpAuth}}-H @/run/secrets/auth_header {{end}} {{.TimewarpURL}} > timewarp
```
wbxyz: Should we attempt to install curl from the PackageManager, the same as docker run?

https://github.com/msuozzo/oss-rebuild/blob/6b8607d840bd58209a660c0bc13501a7145dad1b/pkg/build/local/run_planner.go#L41

pkg/build/local/build_planner.go
```diff
@@ -0,0 +1,154 @@
...
+	var buf bytes.Buffer
+	if err := dockerBuildDockerfileTpl.Execute(&buf, args); err != nil {
+		return "", errors.Wrap(err, "failed to execute Dockerfile template")
+	}
+
+	return buf.String(), nil
+}
+
+// Helper methods to extract configuration from options (matching GCB planner)
+func (p *DockerBuildPlanner) getToolURL(toolType build.ToolType, opts build.PlanOptions) (string, bool, error) {
```
wbxyz: Maybe name the output args or just describe them in the docstring? The bool in particular isn't obvious without reading other usage or the implementation.

wbxyz (approved):

pkg/build/local/build_executor.go
msuozzo: discussed offline. yes it could be but we don't need it now. this is the easiest we can make it to change that decision in the future if we need it, though.

pkg/build/local/build_executor.go
msuozzo: meh we'd need to add another case to buffer the full asset into memory instead of streaming it with io.Copy as we do here. i think this is tolerable

pkg/build/local/build_executor.go
msuozzo: good call. done

PR #581, opened 2025-06-17
Benchmark combine tool

wbxyz:


tools/benchmark/combine/main.go
```diff
@@ -0,0 +1,125 @@
...
+func main() {
+	flag.Parse()
+
+	inFiles := flag.Args()
+	if len(inFiles) == 0 {
+		log.Fatal("Error: At least one input benchmark file path must be provided as an argument.")
+	}
+
+	latestUpdated := time.Time{}
+	// mergedPackagData[Ecosystem][Package][Version][Artifact] == true
```
msuozzo, line 32: can you do Target.ID() with a flat map instead?

tools/benchmark/combine/main.go
```diff
@@ -0,0 +1,125 @@
...
+func sortedKeys[V any](m map[string]V) []string {
+	keys := make([]string, 0, len(m))
+	for k := range m {
+		keys = append(keys, k)
+	}
+	slices.Sort(keys) // Sort versions
+	return keys
+}
+
+func main() {
```
msuozzo: want to just split out the bulk of the functionality into a non-main function?

tools/benchmark/combine/main.go
wbxyz: Is there a Target.ID()? The only one I can think of rn is rundex.Rebuild.ID()? 

I wanted to avoid using string concatenation because I wanted to avoid escaping, and re-parsing the strings back into structs later.

wbxyz: I wrote and re-wrote this a few ways, and while there are probably better ways to write it I think the current is a local minimum.

What we're trying to do here is group things by <ecosystem, package name> so we can make benchmark.Package objects which themselves contain lists of version and artifacts.

For that reason, ID() isn't the right key. We could use a concatenation of the ecosystem and package name, then keep a list of <version, artifact> pairs, but I wanted to de-duplicate version and artifact pairs, so I would probably need another map with concatenation of <version, artifact> to bools.

Because we'll eventually need all four values anyway for constructing the final benchmark, I decided to keep the values simple and separated in these maps which I think is probably worth the cost of the extra six or so lines for extra layers of maps.

PR #582, opened 2025-06-17
Add ctl command to put tracked packages in bucket

wbxyz:


tools/ctl/ctl.go
```diff
@@ -819,6 +821,56 @@ var infer = &cobra.Command{
...
+			return errors.Wrap(err, "reading benchmark file")
+		}
+		packageMap := make(map[string]map[string]bool)
+		for _, p := range set.Packages {
+			if _, ok := packageMap[p.Ecosystem]; !ok {
+				packageMap[p.Ecosystem] = make(map[string]bool)
+			}
+			packageMap[p.Ecosystem][p.Name] = true
+		}
+		data := make(map[string][]string)
```
msuozzo: hmm could we just keep the benchmark format? or define a special type for this instead? thinking ahead to consumption, it would be good to share it between the two.

wbxyz: Yeah I can define a type for this. Benchmark isn't a good fit because specifically do not want version numbers or artifact names. Only ecosystem and package (with maybe some other filters later but that could be on the tracker level). I was also trying to optimize a bit for space, which is why it's not map[string]map[string]bool, because that would result in a whole bunch of ": True" line suffix in the json which I think meaningfully increases the file size.

wbxyz: I added TrackedPackageSet in https://github.com/google/oss-rebuild/pull/584

tools/ctl/ctl.go
```diff
@@ -819,6 +822,56 @@ var infer = &cobra.Command{
...
+func logFailure(f func() error) {
+	if err := f(); err != nil {
+		log.Println(err)
+	}
+}
+
+var setTrackedPackagesCmd = &cobra.Command{
+	Use:   "set-tracked --bench <benchmark.json> <gcs-bucket>",
+	Short: "Set the list of tracked packages",
+	Args:  cobra.MinimumNArgs(1),
```
msuozzo: i think you want exactly 1 arg here

tools/ctl/ctl.go
```diff
@@ -819,6 +822,56 @@ var infer = &cobra.Command{
...
+		ctx := cmd.Context()
+		if *bench == "" {
+			return errors.New("bench must be provided")
+		}
+		bucket := args[len(args)-1]
+		set, err := benchmark.ReadBenchmark(*bench)
+		if err != nil {
+			return errors.Wrap(err, "reading benchmark file")
+		}
+		packageMap := make(map[string]map[string]bool)
```
msuozzo: use new type

deployments/main.tf
```diff
@@ -211,6 +211,14 @@ resource "google_storage_bucket" "bootstrap-tools" {
   default_event_based_hold = true
   depends_on               = [google_project_service.storage]
 }
+# Stores the list of "tracked" packages that are elligible for automatic rebuilds.
+resource "google_storage_bucket" "tracked-packages" {
```
msuozzo: can you break this out into a parent or child change

wbxyz: Done: https://github.com/google/oss-rebuild/pull/588

PR #583, opened 2025-06-17
Add GCS-based Tracker to feed package

wbxyz:


pkg/feed/gcstracker.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+func (tracker *GCSTracker) fetchFromGCS(ctx context.Context) error {
+	log.Println("Fetching new tracked packages list from GCS...")
+	r, err := tracker.obj.NewReader(ctx)
+	if err != nil {
+		return errors.Wrap(err, "opening GCS reader")
+	}
+	defer logFailure(r.Close)
+	w, err := os.Open(tracker.localCache)
```
msuozzo: nah. if we need this, let's take an fs abstraction or a Cache object

pkg/feed/gcstracker.go
```diff
@@ -0,0 +1,128 @@
...
+	return &GCSTracker{obj: obj, localCache: cachePath}
+}
+
+func logFailure(f func() error) {
+	if err := f(); err != nil {
+		log.Println(err)
+	}
+}
+
+func (tracker *GCSTracker) fetchFromGCS(ctx context.Context) error {
```
msuozzo: i think we should eagerly pre-populate prior to the first request. and we need mutual exclusion given inits can race.

pkg/feed/gcstracker.go
```diff
@@ -0,0 +1,128 @@
...
+		eco := rebuild.Ecosystem(ecoStr)
+		if _, ok := trackedPackages[eco]; !ok {
+			trackedPackages[eco] = make(map[string]bool)
+		}
+		for _, pkgName := range pkgs {
+			trackedPackages[eco][pkgName] = true
+		}
+	}
+	tracker.trackedPackages = trackedPackages
+	log.Println(trackedPackages)
```
msuozzo: We probably want to remove this. I like the idea of logging when the cache is refreshed tho (and maybe some stats about the change i.e. old # vs new #)

pkg/feed/gcstracker.go
```diff
@@ -0,0 +1,128 @@
...
+		tracker.fetchFromGCS(ctx)
+	} else if err != nil {
+		return err
+	}
+	return tracker.readCache()
+}
+
+// IsTracked checks if the given TargetEvent corresponds to a tracked package.
+func (lft *GCSTracker) IsTracked(e schema.TargetEvent) (bool, error) {
+	if err := lft.maybeUpdate(context.Background()); err != nil {
```
msuozzo: why does this take a context.Background()?

pkg/feed/gcstracker.go
```diff
@@ -0,0 +1,128 @@
...
+	tracker.trackedPackages = trackedPackages
+	log.Println(trackedPackages)
+	return nil
+}
+
+func (tracker *GCSTracker) maybeUpdate(ctx context.Context) error {
+	if !tracker.lastUpdate.IsZero() && time.Now().Before(tracker.lastUpdate.Add(UpdatePollingInterval)) {
+		return nil
+	}
+	tracker.lastUpdate = time.Now()
```
msuozzo: this should go at the end of the function (on success)

pkg/feed/gcstracker.go
```diff
@@ -0,0 +1,128 @@
...
+	if err != nil {
+		return errors.Wrap(err, "opening cache file")
+	}
+	defer logFailure(r.Close)
+	gzr, err := gzip.NewReader(r)
+	if err != nil {
+		return errors.Wrap(err, "opening zip reader")
+	}
+	defer logFailure(gzr.Close)
+	var rawTracked map[string][]string
```
msuozzo: this should definitely be defined somewhere if this is shared between a producer and a consumer

pkg/feed/gcstracker.go
```diff
@@ -0,0 +1,128 @@
...
+
+	gcs "cloud.google.com/go/storage"
+	"github.com/google/oss-rebuild/pkg/rebuild/rebuild"
+	"github.com/google/oss-rebuild/pkg/rebuild/schema"
+	"github.com/pkg/errors"
+)
+
+const TrackedPackagesFile = "tracked.json.gz"
+
+const UpdatePollingInterval = 15 * time.Minute
```
msuozzo: i think this should be configurable at the top-level, as well

pkg/feed/gcstracker.go
wbxyz: Oops that was a debug line

pkg/feed/gcstracker.go
wbxyz: IsTracked the interface does not include a context, and updating might be making network requests that require a context, so it needs to come from somewhere. We could make a TODO to change the IsTracked interface but probably out of scope for this PR.

pkg/feed/gcstracker.go
wbxyz: I want to rate-limit failures, so it's intentionally on all cases, not just success.

pkg/feed/gcstracker.go
wbxyz: Not opposed to it, was just easier to define as a constant initially

pkg/feed/gcstracker.go
msuozzo: i'd rename "lastUpdateAttempt" then.

pkg/feed/gcstracker.go
msuozzo: does this need to be synchronous then? you could update periodically in the background regardless which would be a lot clearer than doing so lazily

pkg/feed/gcstracker.go
wbxyz: Sent https://github.com/google/oss-rebuild/pull/584

pkg/feed/gcstracker.go
wbxyz: Done. I changed to a billy.Filesystem, and I re-used the TrackedPackagesFile from above

pkg/feed/gcstracker.go
msuozzo: yeah constant default is fine.

pkg/feed/gcstracker.go
wbxyz: I moved to an expectation that we provide the generation-specific object (using a flag for the gs uri that includes generation) and now we don't do background refreshes.

pkg/feed/gcstracker.go
wbxyz: Added eager pre-population, but there's still a danger for multiple inits to race and fetch the GCS object multiple times.

pkg/feed/gcstracker.go
```diff
@@ -0,0 +1,120 @@
...
+		}
+	}
+	tracker.trackedPackages = trackedPackages
+	return nil
+}
+
+// IsTracked checks if the given TargetEvent corresponds to a tracked package.
+func (lft *GCSTracker) IsTracked(e schema.TargetEvent) (bool, error) {
+	// Tracked packages are loaded at construction time via NewGCSTracker.
+	if lft.trackedPackages == nil {
```
msuozzo: meh if this is a post-condition of initialization, I'd be fine omitting this check

pkg/feed/gcstracker.go
wbxyz: I added a mutex, so they should avoid multiple fetches .

pkg/feed/gcstracker.go
wbxyz: I rolled the check into initialization code.

pkg/feed/gcstracker.go
wbxyz: Removed polling, any update in the generation I think should be handled one generation above.

pkg/feed/gcstracker.go
```diff
@@ -0,0 +1,112 @@
...
+		return nil, errors.Wrap(err, "getting GCS object attributes")
+	}
+	gcsLock.Lock()
+	defer gcsLock.Unlock()
+	info, err := fsys.Stat(TrackedPackagesFile)
+	if errors.Is(err, fs.ErrNotExist) || info.ModTime().Before(attrs.Updated) {
+		if err := copyFromGCS(ctx, fsys, obj); err != nil {
+			return nil, errors.Wrap(err, "fetching tracked packages list from GCS")
+		}
+	} else if err != nil {
```
wbxyz: I don't think this is the correct check. If we decide to "roll back" to an older generation, this code would not respect the selected generation. I probably need to encode the generation inside the file. Or maybe in the filename?

Inside the file: It means the file on disk is not the same format as the file in GCS which would be odd but not the end of the world.
Using the filename makes it annoying to delete.

Maybe I should just store the TrackedPackageIndex as a package level var? That's probably how I'm leaning rn.

wbxyz: Moved to *only* a static variable that's shared across all instances. We kick off the population during initialization and then I used a chan to force readers to wait until it's ready.

pkg/feed/datatracker.go
```diff
@@ -0,0 +1,192 @@
...
+		totalPackages += len(pkgs)
+	}
+
+	log.Printf("Loaded index of %d tracked packages\n", totalPackages)
+	sharedIndex.Index = trackedPackages
+	return nil
+}
+
+// GCSDataSource implements DataSource for Google Cloud Storage using int64 generations
+type GCSDataSource struct {
```
msuozzo: nit: Want to call this GCSObjectDataSource? This is specific to a single object and narrowing the naming allows for potentially other versioning strategies built on GCS.

pkg/feed/datatracker.go
```diff
@@ -0,0 +1,192 @@
...
+	Index   TrackedPackageIndex
+	Version T
+	Ready   chan struct{}
+	Err     error
+}
+
+var (
+	// This in-memory cache makes sure multiple init calls don't need to re-read the index.
+	// We use a map keyed by DataSource identifier to support multiple different sources.
+	// TODO: Find a way to avoid storing the entire thing in memory, we can probably find a disk-based search index.
```
msuozzo, line 42: do we still want this TODO?

PR #584, opened 2025-06-17
Add the TrackedPackageSet schema type

wbxyz:
This is the structure of data as it will be stored in GCS and
distributed broadly.

### DIFF
--- a/pkg/feed/tracker.go
+++ b/pkg/feed/tracker.go
@@ -4,6 +4,7 @@
 package feed
 
 import (
+	"github.com/google/oss-rebuild/pkg/rebuild/rebuild"
 	"github.com/google/oss-rebuild/pkg/rebuild/schema"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/rebuild/schema/schema.go
```diff
@@ -310,3 +310,5 @@ const (
 	SmoketestMode ExecutionMode = "smoketest" // No attestations, faster.
 	AttestMode    ExecutionMode = "attest"    // Creates attestations, slower.
 )
+
+type TrackedPackageSet map[string][]string // Map from ecosystem to list of packages.
```
msuozzo: could this go in pkg/feed instead? feel like it doesn't have a strong affinity to our API schema since it's consumed out-of-band through storage.

wbxyz: Moved. Now that it's in pkg/feed I made two, one for storage and one for access

pkg/feed/tracker.go
```diff
@@ -24,3 +25,9 @@ var _ Tracker = &funcTracker{}
 func TrackerFromFunc(isTracked func(schema.TargetEvent) (bool, error)) Tracker {
 	return &funcTracker{isTracked: isTracked}
 }
+
+// TrackedPackageSet is a map from ecosystem to list of packages. Optimized for storage, not access.
+type TrackedPackageSet map[string][]string
```
msuozzo: let's use rebuild.Ecosystem as the key here, too

PR #588, opened 2025-06-18
Create bucket for tracked package set

wbxyz:


deployments/main.tf
```diff
@@ -211,6 +211,17 @@ resource "google_storage_bucket" "bootstrap-tools" {
   default_event_based_hold = true
   depends_on               = [google_project_service.storage]
 }
+# Stores the list of "tracked" packages that are elligible for automatic rebuilds.
```
msuozzo: elligible -> eligible

PR #590, opened 2025-06-18
Rename remaining uses of 'normalize' to 'stabilize'

msuozzo:


wbxyz (approved):
Is it worth adding a github actions check to warn for this? I could definitely see myself or an external contributor accidentally adding something named "normalize" instead of stabilize.

msuozzo:
> Is it worth adding a github actions check to warn for this? I could definitely see myself or an external contributor accidentally adding something named "normalize" instead of stabilize.

eh I think we should be good with this cleanup. once this is in, there are essentially no remaining uses of the old terminology so it would mostly be a risk to existing contributors.

PR #591, opened 2025-06-18
Add analyze option to run-one

msuozzo:


tools/ctl/ctl.go
```diff
@@ -480,16 +480,16 @@ var runBenchmark = &cobra.Command{
...
 var runOne = &cobra.Command{
-	Use:   "run-one smoketest|attest --api <URI> --ecosystem <ecosystem> --package <name> --version <version> [--artifact <name>] [--strategy <strategy.yaml>] [--strategy-from-repo]",
+	Use:   "run-one smoketest|attest|analyze --api <URI> --ecosystem <ecosystem> --package <name> --version <version> [--artifact <name>] [--strategy <strategy.yaml>] [--strategy-from-repo]",
 	Short: "Run benchmark",
 	Args:  cobra.ExactArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
 		if *ecosystem == "" || *pkg == "" || *version == "" {
 			log.Fatal("ecosystem, package, and version must be provided")
 		}
 		mode := schema.ExecutionMode(args[0])
```
wbxyz, line 495: Wdyt of defining a local analyzeMode:

```go
analyzeMode := schema.ExecutionMode("analyze")
```

Then all the comparisons can use `mode == analyzeMode` rather than `args[0] == "analyze"`, and the string literal is defined once rather than inline everywhere.

msuozzo: oh yeah that's a better idea. can clean up the control-flow, too, now.

PR #598, opened 2025-06-25
Catch build creation failure

msuozzo:


pkg/rebuild/rebuild/rebuildremote.go
```diff
@@ -508,6 +508,9 @@ func doCloudBuild(ctx context.Context, client gcb.Client, build *cloudbuild.Buil
 		buildCtx = bctx
 	}
 	build, err := gcb.DoBuild(buildCtx, client, opts.Project, build, gcb.DoBuildOpts{TerminateOnTimeout: terminateOnTimeout})
+	if build == nil {
+		return errors.Wrap(err, "creating build")
+	}
 	bi.BuildID = build.Id
 	bi.Steps = build.Steps
```
wbxyz: If build==nil AND err==nil, this will  return nil I think?

What about:
```go
if build != nil {
  	bi.BuildID = build.Id
	bi.Steps = build.Steps
}
if err != nil {
  return errors.Wrap(err, "doing build")
} else if build == nil {
  return errors.New("unexpected empty build result")
}
```

msuozzo: done

pkg/rebuild/rebuild/rebuildremote.go
```diff
@@ -508,8 +508,11 @@ func doCloudBuild(ctx context.Context, client gcb.Client, build *cloudbuild.Buil
...
-	bi.BuildID = build.Id
-	bi.Steps = build.Steps
+	if build != nil {
+		// Set all the values of BuildInfo we can, regardless of the error response.
+		bi.BuildID = build.Id
+		bi.Steps = build.Steps
+	}
 	if err != nil {
 		return errors.Wrap(err, "doing build")
 	}
```
wbxyz, line 518: You'll want to add a separate case for `build == nil` because we're going to attempt to use it below.

msuozzo: nah the contract of the function is `err != nil` -> `build != nil`. if we have an err, we exit. If we don't, it's fine to assume build != nil.

wbxyz: I weakly lean towards not trusting that contract, but I'll defer to your decision there.

msuozzo: I think tests are the appropriate venue for such validation, not additional in-code checks.

PR #601, opened 2025-06-25
Add VPC support for pools

msuozzo:


deployments/main.tf
```diff
@@ -491,6 +496,76 @@ data "google_artifact_registry_docker_image" "network-subscriber" {
...
+  network = google_compute_network.vpc[0].name
+  allow {
+    protocol = "tcp"
+    ports    = ["0-65535"]
+  }
+  allow {
+    protocol = "udp"
+    ports    = ["0-65535"]
+  }
+  source_ranges = ["10.0.0.0/20"]
```
wbxyz: When I deployed this in testing, I had to change this line:

```
source_ranges = ["${google_compute_global_address.private_service_access[0].address}/${google_compute_global_address.private_service_access[0].prefix_length}"]
```

deployments/main.tf
```diff
@@ -491,6 +496,76 @@ data "google_artifact_registry_docker_image" "network-subscriber" {
...
+  network   = google_compute_network.vpc[0].name
+  direction = "EGRESS"
+  allow {
+    protocol = "tcp"
+  }
+  allow {
+    protocol = "udp"
+  }
+  destination_ranges = ["0.0.0.0/0"]
+}
```
wbxyz, line 574: I think we also need a subnetwork for any shared services to be deployed into:

```
resource "google_compute_subnetwork" "rebuild_subnet" {
  count     = var.enable_vpc ? 1 : 0
   name          = "${var.host}-rebuild-subnet"
   ip_cidr_range = "10.10.1.0/24"
   region        = "us-central1"
   network       = google_compute_network.vpc[0].name
}
```

PR #602, opened 2025-06-26
Add UDS support for docker proxy

andusy:


cmd/proxy/main.go
```diff
@@ -28,7 +28,7 @@ var (
 	httpProxyAddr              = flag.String("http_addr", "localhost:3128", "address for HTTP proxy")
 	tlsProxyAddr               = flag.String("tls_addr", "localhost:3129", "address for TLS proxy")
 	ctrlAddr                   = flag.String("ctrl_addr", "localhost:3127", "address for administrative endpoint")
-	dockerAddr                 = flag.String("docker_addr", "", "address for docker proxy endpoint")
+	dockerAddr                 = flag.String("docker_addr", "", "address for docker proxy endpoint in the format host:port for tcp, or unix:///file for unix domain sockets.")
```
msuozzo: we might also support tcp://... since I believe DOCKER_HOST does the same: https://docs.docker.com/reference/cli/docker/#host

cmd/proxy/main.go
```diff
@@ -28,7 +28,7 @@ var (
...
 	dockerSocket               = flag.String("docker_socket", "/var/run/docker.sock", "path to the docker socket")
```
msuozzo, line 33: honestly, it would be nice to have this flag support the same things. it would require tweaking the client code for the socket so feel free to leave it be for now but you might add a TODO to bring this up to parity.

pkg/proxy/docker/docker.go
```diff
@@ -469,9 +469,21 @@ func (d *ContainerTruststorePatcher) leasePatchSet(container string) *patchSet {
 
 // Proxy serves the Docker API while patching the container truststore.
 func (d *ContainerTruststorePatcher) Proxy(srvAddr, dockerAddr string) {
-	tcpChan := make(chan net.Conn, 1)
+	dockerProxyChan := make(chan net.Conn, 1)
```
msuozzo: nit: maybe "proxyChan" and "realChan" or something like that

pkg/proxy/docker/docker.go
```diff
@@ -469,9 +469,21 @@ func (d *ContainerTruststorePatcher) leasePatchSet(container string) *patchSet {
...
 	udsChan := make(chan net.Conn, 1)
-	tl, err := net.Listen("tcp", srvAddr)
+
+	network := "tcp"
+	address := srvAddr
+	if strings.HasPrefix(address, "unix://") {
+		network = "unix"
+		address = strings.TrimPrefix(address, "unix://")
+		if _, err := os.Stat(address); err == nil {
+			if err := os.RemoveAll(address); err != nil {
```
msuozzo: woah... I think maybe we should fail instead in this case. it seems like net.Listen should unlink the socket on exit so this would only be an issue if there were an outstanding proxy (in which case we'd break the other proxy by removing it) or another process on the system created/is using it (in which case we'd really want the caller to clean it up). Best just to leave it

andusy: Good point. I'll leave it up to the client/user to remove the dangling socket file if it exists.

cmd/proxy/main.go
andusy: I left a TODO, but only mentioned SSH since _I don't think_ the proxy works on Windows?

msuozzo: ah no i meant that we don't support "tcp" here. note the hard-coded "unix" bind below on line 522.

I don't think we have a compelling need to support ssh so you can leave that off.

andusy: Oh I mixed up the line that was being highlighted. Fixed.

PR #604, opened 2025-06-27
Add abstraction to fetch GCB logs

msuozzo:


pkg/gcb/gcb.go
```diff
@@ -31,6 +33,55 @@ type Client interface {
...
+	ReadBuildLogs(ctx context.Context, bucket, buildID string) (io.ReadCloser, error)
+	// ReadStepLogs reads logs for a specific step within a build from the specified bucket
+	ReadStepLogs(ctx context.Context, bucket, buildID string, stepIndex int) (io.ReadCloser, error)
+	// ListStepLogs returns the available step log files for a build in the specified bucket
+	ListStepLogs(ctx context.Context, bucket, buildID string) (int, error)
+}
+
+// gcsLogsClient implements LogsClient using Google Cloud Storage.
+type gcsLogsClient struct {
+	gcsClient *gcs.Client
```
wbxyz, line 49: Should this just store a BucketHandle? That way read call locations don't need to be aware of which bucket name they're reading from.

wbxyz (approved):

pkg/gcb/gcb.go
msuozzo: yeah the main justification i think is resonant is that we could read using the GCB native logging interface by dropping the bucket param (which wouldn't be necessary)

wbxyz (approved):

PR #605, opened 2025-07-01
stabilize: fix error when stabilizing tars

winterqt:
Before this patch, `stabilize /path/to/a/tar.gz` would fail:

```
panic: runtime error: comparing uncomparable type archive.TarArchiveStabilizer

goroutine 1 [running]:
slices.Equal[...](...)
        slices/slices.go:25
main.eligiblePasses({0x7fffffffe1ae, 0x10})
        github.com/google/oss-rebuild/cmd/stabilize/main.go:185 +0x2fb
main.run()
        github.com/google/oss-rebuild/cmd/stabilize/main.go:212 +0x198
main.main()
        github.com/google/oss-rebuild/cmd/stabilize/main.go:246 +0x13
```

### DIFF
--- a/cmd/stabilize/main.go
+++ b/cmd/stabilize/main.go
@@ -182,7 +182,9 @@ func eligiblePasses(filename string) ([]archive.Stabilizer, error) {
 		}
 		if i == 0 {
 			result = stabs
-		} else if !slices.Equal(result, stabs) {
+		} else if !slices.EqualFunc(result, stabs, func(s1, s2 archive.Stabilizer) bool {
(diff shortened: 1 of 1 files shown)


winterqt:
Thanks for the quick review!

I admit, I didn’t look too closely at the various code paths at play here: would this only trigger in certain cases (presumably those without an ecosystem?), which would explain why it wasn’t caught till now?

> Maybe we could change the stabilizers to be pointer types instead?

If we’re sure that the various pointers compared will always be identical, then I think that’s a good solution if this isn’t preferable.

PR #607, opened 2025-07-10
Move modalFunc to modal package

wbxyz:


tools/ctl/ide/modal/modal.go
```diff
@@ -14,6 +14,9 @@ const (
 	defaultModalBackground = tcell.ColorDarkCyan
 )
 
+// ModalFnType can be used to show an InputCaptureable. It returns an exit function that can be used to close the modal.
+type ModalFnType func(InputCaptureable, ModalOpts) func()
```
msuozzo: maybe just Fn then? it'll be referred to as modal.Fn then which feels ok

PR #614, opened 2025-07-10
Move explorer historical view into dedicated rebuildhistory package

wbxyz:


tools/ctl/ide/rebuildhistory/rebuildhistory.go
```diff
@@ -0,0 +1,119 @@
...
+	})
+	populateDetails := func(node *tview.TreeNode) {
+		if node == root {
+			details.SetText("")
+			return
+		}
+		d, ok := node.GetReference().(*rundex.Rebuild)
+		if !ok {
+			log.Println("Node has unexpected reference")
+			details.SetText("ERROR")
```
msuozzo: for a follow-on: might be nice to include more detail about the error in the text area instead of in the log (which may not be visible?)

msuozzo (approved):

PR #617, opened 2025-07-10
Change FirestoreClient to handle benchmark filters faster

wbxyz:
Previously when filtering by a benchmark set, the FirestoreClient would
query all records global only filtering clientside. This created
enormous wait times that conintued to grow as we add more documents.

### DIFF
--- a/tools/ctl/ide/commands/commands.go
+++ b/tools/ctl/ide/commands/commands.go
@@ -463,8 +463,9 @@ func NewBenchmarkCmds(app *tview.Application, rb *rebuilder.Rebuilder, modalFn m
 					return
 				}
 				tracked := make(map[string]bool)
+				var set benchmark.PackageSet
 				{
-					set, err := benches.Load(benchPath)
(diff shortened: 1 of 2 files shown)


tools/ctl/ide/commands/commands.go
```diff
@@ -501,9 +501,16 @@ func NewBenchmarkCmds(app *tview.Application, rb *rebuilder.Rebuilder, modalFn m
...
 					rebuilds = slices.DeleteFunc(rebuilds, func(r rundex.Rebuild) bool {
-						return !tracked[r.ID()]
+						if tracked[r.ID()] {
+							return false
+						} else {
+							extranious += 1
+							return true
+						}
 					})
+					log.Printf("Throwing away %d extranious rebuilds", extranious)
```
msuozzo: while theoretically useful for debugging performance issues with the TUI, i'm not sure this is worth keeping around now that you've implemented the dedup.

msuozzo (approved):

PR #620, opened 2025-07-10
Add a csv format to the get-results command

wbxyz:


tools/ctl/ctl.go
msuozzo, line 1: add csv ref to `Use` description

PR #621, opened 2025-07-11
Add package diff to AI clustering context

wbxyz:


tools/ctl/ide/commands/commands.go
```diff
@@ -329,6 +328,23 @@ func NewRebuildGroupCmds(app *tview.Application, rb *rebuilder.Rebuilder, modalF
 						{Text: "Please summarize this rebuild failure in one sentence."},
 						{Text: logs},
 					}
+					if strings.Contains(in.Message, "content mismatch") {
```
msuozzo, line 331: do we include the logs in the context if the failure is a diff? I guess it could maybe help suggest an alternative but I'm less sure it would improve the performance of the clustering. input tokens are cheap, i guess....

wbxyz: Yeah I figured any error messages from the logs could be helpful for the AI deciding why there's a diff. I suppose we can experiment but I'll err on the side of too much info for now and trim back later.

tools/ctl/ide/commands/commands.go
```diff
@@ -329,6 +328,23 @@ func NewRebuildGroupCmds(app *tview.Application, rb *rebuilder.Rebuilder, modalF
...
+						diffPath, err := butler.Fetch(ctx, in.RunID, in.WasSmoketest(), diffoscope.DiffAsset.For(in.Target()))
+						if err != nil {
+							log.Println(errors.Wrap(err, "fetching diff"))
+						} else {
+							diffContent, err := os.ReadFile(diffPath)
+							if err != nil {
+								log.Println(errors.Wrap(err, "reading diff"))
+							} else {
+								diffStr := string(diffContent)
+								if len(diffStr) > uploadBytesLimit {
```
msuozzo, line 341: we might want to tune the args for diffoscope for this use-case since we likely only want a few chucks per file. could be done in a follow-on, though

msuozzo (approved):

tools/ctl/ide/commands/commands.go
wbxyz: Yeah I'll leave it for a later PR. Unless we want *all* uses of the diff to have the same setting, we might need to explore breaking the asset type into two different types (ShortDiff and CompleteDiff)?

PR #625, opened 2025-07-14
Add new GCB build abstraction

msuozzo:


pkg/build/gcb/executor.go
```diff
@@ -0,0 +1,284 @@
...
+	client           gcb.Client
+	logsClient       gcb.LogsClient
+	project          string
+	serviceAccount   string
+	logsBucket       string
+	privatePool      *gcb.PrivatePoolConfig
+	outputBufferSize int
+
+	activeBuildsMu sync.RWMutex
+	activeBuilds   map[string]*gcbHandle
```
wbxyz: syncx.Map?

pkg/build/gcb/executor.go
```diff
@@ -0,0 +1,284 @@
...
+	outputBufferSize := config.OutputBufferSize
+	if outputBufferSize <= 0 {
+		outputBufferSize = 1024 * 1024 // 1MB default
+	}
+	return &GCBExecutor{
+		planner:          planner,
+		client:           config.Client,
+		logsClient:       config.LogsClient,
+		project:          config.Project,
+		serviceAccount:   config.ServiceAccount,
```
wbxyz: Should we check these for empty values?

pkg/build/gcb/executor.go
```diff
@@ -0,0 +1,284 @@
...
+}
+
+// Status implements build.Executor
+func (e *GCBExecutor) Status() build.ExecutorStatus {
+	e.activeBuildsMu.RLock()
+	activeCount := len(e.activeBuilds)
+	e.activeBuildsMu.RUnlock()
+	return build.ExecutorStatus{
+		InProgress: activeCount,
+		Capacity:   -1, // TODO: Allow user to configure sizing
```
wbxyz, line 129: Maybe just configure this in in GCBExecutorConfig? It would be cool to figure it out dynamically though.

pkg/build/gcb/executor.go
```diff
@@ -0,0 +1,284 @@
...
+	"fmt"
+	"io"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/google/oss-rebuild/internal/bufiox"
+	"github.com/google/oss-rebuild/internal/gcb"
+	"github.com/google/oss-rebuild/pkg/build"
+	"github.com/google/oss-rebuild/pkg/rebuild/rebuild"
```
wbxyz, line 16: Is this going to cause a dependency loop if we try to use this executor in rebuildremote?

pkg/build/gcb/executor.go
msuozzo: idk we could but i'm not sure it's worthwhile. i think each of them does represent something a user might want to do so i'd probably omit it for now

pkg/build/gcb/executor.go
msuozzo: eh yeah i figured i'd just punt. it's no worse than what we do currently. and i think we only really need small-scale / serial use of each object anyway so no urgent need for the feature.

pkg/build/gcb/executor.go
msuozzo: yeah as discussed offline, we'd probably need to get rid of that and reimplement each RebuildRemote ecosystem impl using an Executor. should be manageable but i don't think there's a good way to break this dependence. that can come in a follow-on when we integrate.

pkg/build/gcb/executor.go
```diff
@@ -0,0 +1,302 @@
...
+	go e.executeBuild(ctx, handle, gcbBuild, input, opts, plan)
+	return handle, nil
+}
+
+// Status implements build.Executor
+func (e *Executor) Status() build.ExecutorStatus {
+	activeCount := 0
+	for range e.activeBuilds.Values() {
+		activeCount++
+	}
```
wbxyz, line 126: This pattern seems common, maybe add Len() to syncx.Map? Out of scope for this PR but maybe a followup.

pkg/build/gcb/executor.go
```diff
@@ -0,0 +1,302 @@
...
+	// Upload assets to asset store if configured
+	var finalErr error
+	if opts.Resources.AssetStore != nil {
+		if err := e.uploadContent(ctx, opts.Resources.AssetStore, rebuild.DockerfileAsset.For(input.Target), []byte(plan.Dockerfile)); err != nil {
+			finalErr = errors.Wrap(err, "uploading Dockerfile")
+		}
+		if err := e.uploadBuildInfo(ctx, opts.Resources.AssetStore, rebuild.BuildInfoAsset.For(input.Target), buildInfo); err != nil {
+			finalErr = errors.Wrap(err, "uploading build info")
+		}
+	}
```
wbxyz: The buildInfo is frequently an important part of debugging after a build fails. We should try to write the BuildInfo into the AssetStore in most failure cases, with whatever info we have available.

pkg/build/gcb/executor.go
```diff
@@ -0,0 +1,302 @@
...
+	handle.updateStatus(build.BuildStateCompleted)
```
wbxyz, line 214: optional: Add a deferred call to set the state as completed to make sure none of the early returns accidentally leave the build in an invalid state.

pkg/build/gcb/handle.go
```diff
@@ -0,0 +1,110 @@
...
+
+// writeOutput writes data to the output stream
+func (h *gcbHandle) writeOutput(data []byte) (n int, err error) {
+	return h.output.Write(data)
+}
+
+// Write writes data to the output stream (implements io.Writer)
+func (h *gcbHandle) Write(data []byte) (n int, err error) {
+	return h.output.Write(data)
+}
```
wbxyz: Is there a reason to include the private writeOutput if we also have the public Write()?

pkg/build/gcb/executor.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+	// Generate the execution plan
+	planOpts := build.PlanOptions{
+		UseTimewarp:            opts.UseTimewarp,
+		UseNetworkProxy:        opts.UseNetworkProxy,
+		UseSyscallMonitor:      opts.UseSyscallMonitor,
+		PreferPreciseToolchain: true,
+		Resources:              opts.Resources,
+	}
+	plan, err := e.planner.GeneratePlan(ctx, input, planOpts)
```
wbxyz, line 98: Should the dockerfile and build steps be written to the opts.Resources.AssetStore here?

wbxyz (approved):

pkg/build/gcb/executor.go
msuozzo: I have a PR for it but it substantially increases the complexity of the impl in exchange for correctness. At the very least, the absence of an API means a user who really needs accuracy needs to put some effort in. We don't care about it (for the moment) in any of our implementations so we can shelve it until we have a use-case.

pkg/build/gcb/executor.go
msuozzo: oof. yeah i forgot about that. fixed

pkg/build/gcb/executor.go
msuozzo: the refactor means we only called is once so it's less of a concern

pkg/build/gcb/handle.go
msuozzo: nah this was from the previous impl where it was always private.

pkg/build/gcb/executor.go
msuozzo: i like the idea of keeping the uploads in the same place. additionally, the steps are different pre- vs post-build since they have success/execution metadata in the latter so there is some benefit in delaying upload to executeBuild.

PR #626, opened 2025-07-14
Add equivalence test for GCB builders

msuozzo:


pkg/build/gcb/equivalence_test.go
```diff
@@ -0,0 +1,423 @@
...
+//
+// The core property being tested: "For any valid input, both implementations
+// must produce identical Dockerfiles and Cloud Build steps"
+//
+// Usage Examples:
+//   # Basic fuzz test (comprehensive parameters)
+//   go test -fuzz FuzzImplementationEquivalence -fuzztime 30s
+//
+//   # Run all fuzz tests briefly
+//   go test -fuzz . -fuzztime 5s
```
wbxyz, line 23: Do you need to enable this in our CI?

msuozzo: nah no need to run it in CI. just a nice artifact that we can use as an example while we migrate over.

PR #637, opened 2025-07-16
Add tui flag for shared asset store

wbxyz:
This can be used for collaborating on rebuild datasets

### DIFF
--- a/tools/ctl/ctl.go
+++ b/tools/ctl/ctl.go
@@ -139,7 +139,7 @@ func makeShellScript(input rebuild.Input) (string, error) {
 }
 
 var tui = &cobra.Command{
-	Use:   "tui [--project <ID>] [--debug-storage <bucket>] [--benchmark-dir <dir>] [--clean] [--llm-project] [--rundex-gcs-path <path>]",
+	Use:   "tui [--project <ID>] [--debug-storage <bucket>] [--benchmark-dir <dir>] [--clean] [--llm-project] [--rundex-gcs-path <path>] [--merged-asset-store <path>]",
 	Short: "A terminal UI for the OSS-Rebuild debugging tools",
 	Args:  cobra.NoArgs,
 	Run: func(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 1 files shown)


tools/ctl/ctl.go
```diff
@@ -139,7 +139,7 @@ func makeShellScript(input rebuild.Input) (string, error) {
 }
 
 var tui = &cobra.Command{
-	Use:   "tui [--project <ID>] [--debug-storage <bucket>] [--benchmark-dir <dir>] [--clean] [--llm-project] [--rundex-gcs-path <path>]",
+	Use:   "tui [--project <ID>] [--debug-storage <bucket>] [--benchmark-dir <dir>] [--clean] [--llm-project] [--rundex-gcs-path <path>] [--shared-asset-store <path>]",
```
msuozzo: nit: merged-asset-store?

tools/ctl/ctl.go
```diff
@@ -207,7 +207,30 @@ var tui = &cobra.Command{
...
 		}
-		butler := localfiles.NewButler(*metadataBucket, *logsBucket, *debugStorage, mux, localfiles.AssetStore)
+		var assetStoreFn func(runID string) (rebuild.LocatableAssetStore, error)
+		if *sharedAssetStore != "" {
+			u, err := url.Parse(*sharedAssetStore)
+			if err != nil {
+				log.Fatal(errors.Wrap(err, "parsing --shared-asset-store"))
+			}
+			if u.Scheme != "gs" {
+				log.Fatal("--shared-asset-store must be a gs:// URL")
```
msuozzo: not sure we want this constraint. having a local one would be fine, no? if we want to put off adding it, let's add a TODO and an "unimplemented" message

msuozzo (approved):

tools/ctl/ctl.go
wbxyz: Yeah I think that's better

tools/ctl/ctl.go
wbxyz: I'll punt on local fs based storage for now, the localfiles.AssetStore kind of fills that role already using a cached asset store. There's some room for improving these asset store abstractions more broadly, which I drafted locally but aren't quite polished yet.

Added a comment and about future support for other schemes, and updated the error message.

PR #638, opened 2025-07-16
Include the uploadpath in DebugStore error

wbxyz:


pkg/rebuild/rebuild/storage.go
```diff
@@ -123,7 +123,7 @@ func DebugStoreFromContext(ctx context.Context) (AssetStore, error) {
 			os.MkdirAll(path, 0755)
 			return NewFilesystemAssetStore(osfs.New(path)), nil
 		}
-		return nil, errors.New("unsupported upload path")
+		return nil, errors.Errorf("unsupported upload path: '%s'", uploadpath)
```
msuozzo: maybe specify that it's an unsupported scheme and that we expect a URI instead of adding in the full path. that's not the part that makes debugging tricky, it's that the message itself doesn't actually explain what element of the path is unsupported.

wbxyz: I ran into this problem because the uploadpath was just empty because of issues elsewhere in the calling path. I can add that as an explict check if you don't want to include the uploadpath

PR #641, opened 2025-07-17
Create initial agent framework setup

msuozzo:


cmd/agent-api/main.go
```diff
@@ -0,0 +1,116 @@
...
+		gcbClient = gcb.NewClientWithPrivatePool(svc, privatePoolConfig)
+	} else {
+		gcbClient = gcb.NewClient(svc)
+	}
+	executorConfig := buildgcb.ExecutorConfig{
+		Project:            *project,
+		ServiceAccount:     *buildRemoteIdentity,
+		LogsBucket:         *logsBucket,
+		Client:             gcbClient,
+		BuilderName:        os.Getenv("K_REVISION"),
```
wbxyz: Should the BuilderName be different from the regular api builder? Maybe

`"agent-api@" + os.Getenv("K_REVISION")`

cmd/agent/main.go
```diff
@@ -0,0 +1,80 @@
...
+
+	"cloud.google.com/go/storage"
+	"github.com/google/oss-rebuild/internal/api"
+	"github.com/google/oss-rebuild/pkg/rebuild/schema"
+)
+
+var (
+	sessionID      = flag.String("session-id", "", "Session ID for this agent run")
+	agentAPIURL    = flag.String("agent-api-url", "", "URL of the agent API service")
+	sessionsBucket = flag.String("sessions-bucket", "", "GCS bucket for session data")
```
wbxyz: Does the agent need to directly access the sessions bucket? I figured necessary info would be returned from status requests

cmd/agent/main.go
```diff
@@ -0,0 +1,80 @@
...
+
+	// TODO: Implement actual agent logic here
+	// For now, just complete immediately as a no-op
+	req := schema.AgentCompleteRequest{
+		SessionID:  *sessionID,
+		StopReason: "NO_OP",
+		Summary:    "Agent completed as no-op stub",
+	}
+
+	_, err = completeStub(ctx, req)
```
wbxyz, line 85: Should we do a defer call to complete before the "main logic" that's currently TODO? Just in case something returns an error in the loop, it might be good to signal that a session has terminated.

deployments/main.tf
```diff
@@ -706,6 +770,37 @@ resource "google_cloud_run_v2_service" "network-subscriber" {
...
+        "--prebuild-version=${var.prebuild_version}",
+        "--prebuild-auth=${var.public ? "false" : "true"}",
+        ], var.enable_private_build_pool ? [
+        "--gcb-private-pool-name=${google_cloudbuild_worker_pool.private-pool[0].id}",
+        "--gcb-private-pool-region=us-central1",
+      ] : [])
+      resources {
+        limits = {
+          cpu    = "1000m"
+          memory = "1G"
```
wbxyz, line 885: This seems like it might be too small?

internal/api/agentapiservice/complete.go
```diff
@@ -0,0 +1,56 @@
...
+		docSnap, err := t.Get(sessionDoc)
+		if err != nil {
+			return errors.Wrap(err, "fetching session")
+		}
+		var session schema.AgentSession
+		if err := docSnap.DataTo(&session); err != nil {
+			return errors.Wrap(err, "parsing session data")
+		}
+		// Check if already completed
+		if session.Status == "COMPLETED" {
```
wbxyz: can this be a const somewhere? Maybe in schema?

deployments/main.tf
msuozzo: yeah depending on how many get scheduled for a build, things like copying logs might end up being too expensive but none of those would need to be buffered in memory (nor are we reading artifacts or any other large files). It's good to note but I think we should just go ahead and see how this provisioning matches our requirements.

cmd/agent-api/main.go
msuozzo: yeah sg

internal/api/agentapiservice/iteration.go
```diff
@@ -0,0 +1,152 @@
...
+import (
+	"context"
+	"io/fs"
+	"path"
+	"time"
+
+	"cloud.google.com/go/firestore"
+	"github.com/google/oss-rebuild/internal/api"
+	"github.com/google/oss-rebuild/pkg/build"
+	buildgcb "github.com/google/oss-rebuild/pkg/build/gcb"
```
wbxyz: Does this need the alias? Or maybe it can just be "gcb"?

cmd/agent/main.go
msuozzo: nope it's an oblivious write. for now, just llm session logs.

internal/api/agentapiservice/complete.go
msuozzo: yeah i don't love this but, atm, i think it's tolerable. a good TODO but we're currently only writing them so we haven't hit necessity yet.

internal/api/agentapiservice/iteration.go
```diff
@@ -0,0 +1,152 @@
...
+		} else if err == nil {
+			return errors.Wrap(fs.ErrExist, "checking for existing iteration")
+		}
+		// Create iteration record
+		iteration = schema.AgentIteration{
+			ID:        iterationID,
+			SessionID: req.SessionID,
+			Number:    req.IterationNumber,
+			Strategy:  req.Strategy,
+			BuildID:   buildID,
```
wbxyz: If we eventually want to have both the oblivious ID and the GCB ID, it might be worth giving this value in firestore a clearly distinct name. In the attestation build infra it's referred to as ObliviousID and BuildID: https://github.com/google/oss-rebuild/blob/ad9d2bfce2d3778c8b74cde628a4095565404715/pkg/rebuild/schema/schema.go#L259

internal/api/agentapiservice/iteration.go
```diff
@@ -0,0 +1,152 @@
...
+			Status:    "PENDING",
```
wbxyz: Const?

internal/api/agentapiservice/iteration.go
```diff
@@ -0,0 +1,152 @@
...
+	// Extract strategy and use GCB executor to plan and execute build
+	if req.Strategy == nil {
+		return nil, api.AsStatus(codes.InvalidArgument, errors.New("strategy is required"))
+	}
+	strategy, err := req.Strategy.Strategy()
+	if err != nil {
+		return nil, api.AsStatus(codes.InvalidArgument, errors.Wrap(err, "invalid strategy"))
+	}
+	// Use GCB executor to plan and execute the build using Target from session
+	store, err := rebuild.NewGCSStore(ctx, path.Join(deps.MetadataBucket, buildID))
```
wbxyz: I think you need to set the rebuild.RunID context var to be buildID?

cmd/agent-api/main.go
```diff
@@ -0,0 +1,117 @@
...
+	d.BuildProject = *project
+	d.BuildServiceAccount = *buildRemoteIdentity
+	d.LogsBucket = *logsBucket
+	d.MetadataBucket = *metadataBucket
+	if *prebuildVersion != "" {
+		prebuildRepo, err := serviceid.ParseLocation(BuildRepo, *prebuildVersion)
+		if err != nil {
+			return nil, errors.Wrap(err, "parsing prebuild location")
+		}
+		d.PrebuildConfig.Dir = prebuildRepo.Ref
```
wbxyz, line 93: Carry-over the comment from API?

```
	// NOTE: The subdir will match the version identifier used for the service version.
```

pkg/rebuild/schema/schema.go
```diff
@@ -310,3 +310,103 @@ const (
...
+type AgentCreateRequest struct {
+	Target         rebuild.Target `form:",required"`
+	MaxIterations  int            `form:""`
+	TimeoutSeconds int            `form:""`
+	Context        *AgentContext  `form:""`
+}
+
+var _ api.Message = AgentCreateRequest{}
+
+func (AgentCreateRequest) Validate() error { return nil }
```
wbxyz: Check that Target is fully specified?

pkg/rebuild/schema/schema.go
```diff
@@ -310,3 +310,103 @@ const (
...
+	Strategy        *StrategyOneOf `form:",required"`
+}
+
+var _ api.Message = AgentCreateIterationRequest{}
+
+func (AgentCreateIterationRequest) Validate() error { return nil }
+
+// AgentBuildConfig contains build configuration
+type AgentBuildConfig struct {
+	Substitutions map[string]string `json:"substitutions,omitempty"`
```
wbxyz: What are substitutions?

wbxyz (approved):

cmd/agent/main.go
msuozzo: ah actually it's not really feasible at the moment since ~all the errors here are on the critical path of setting up the client to all complete in the first place. i've added a TODO to make sure to do think when the implementation is added.

internal/api/agentapiservice/complete.go
msuozzo: eh i ended up added it

pkg/rebuild/schema/schema.go
msuozzo: left in accidentally

PR #643, opened 2025-07-21
Bubble up DeadlineExceeded error from gcb

msuozzo:
Previously, this error suppression led to parsing errors when the caller
assumed the returned Build would be complete. Since we know we're
returning an incomplete value in this case, we need to pass on the
timeout error.


internal/gcb/gcb_test.go
```diff
@@ -72,8 +73,8 @@ func TestDoBuildTimeoutNoTerminate(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
 	defer cancel()
 	b, err := DoBuild(ctx, c, "project", &cloudbuild.Build{}, DoBuildOpts{TerminateOnTimeout: false})
-	if err != nil {
-		t.Errorf("DoBuild unexpected error: %v", err)
+	if err == nil || !errors.Is(err, context.DeadlineExceeded) {
+		t.Errorf("DoBuild expected DeadlineExceeded: got %v", err)
```
loosebazooka, line 77: seems like you were expecting DeadlineExceeded or *no error*, not just DeadlineExceeded

internal/gcb/gcb.go
```diff
@@ -143,8 +143,8 @@ func DoBuild(ctx context.Context, client Client, project string, build *cloudbui
 		}
 	} else if errors.Is(err, context.DeadlineExceeded) {
 		log.Printf("GCB deadline exceeded, allowing the build to continue")
-		// Note: This is the Build metadata returned by CreateBuild
-		return bm.Build, nil
+		// NOTE: This is the Build metadata returned by CreateBuild
```
loosebazooka, line 146: Can you explain this to me cause I'm curious. If the GCB build ends prematurely, what does it mean that you are `allowing the build to continue`?

msuozzo: well, it's more like "our GCB client returns prematurely." the build itself persists in GCP beyond the deadline error we encounter client-sided

internal/gcb/gcb_test.go
msuozzo: nah it's negated. admittedly I could (and should) just remove the nil check since that's captured by the errors.Is check

internal/gcb/gcb.go
```diff
@@ -143,8 +143,8 @@ func DoBuild(ctx context.Context, client Client, project string, build *cloudbui
 		}
 	} else if errors.Is(err, context.DeadlineExceeded) {
 		log.Printf("GCB deadline exceeded, allowing the build to continue")
```
loosebazooka: update this error message to indicate the client timed out?

msuozzo: yep good call. done

loosebazooka (approved):
I don't know who is the target of this error message (just your team?). Does it make sense to add in that you can check the status of the build somewhere else? or does the caller of DoBuild handle that?

msuozzo:
It's an internal note that we should have enough info to react to from elsewhere. I do, however, notice that we mention an ID when cancelling the build and that the ID should really be a different value so I'll update them both in a follow-on.

PR #645, opened 2025-07-21
Update known rust compiler versions

msuozzo:


pkg/registry/cratesio/rust.go
```diff
@@ -36,8 +36,23 @@ func RustVersionAt(t time.Time) (string, error) {
 }
 
 // List generated with the following script:
-// curl -L https://github.com/rust-lang/rust/raw/4739948/RELEASES.md | rg --replace='newRelease("$1", "$2"),' -o 'Version (\S+) \(([^)]+)\)'
+// curl -L https://github.com/rust-lang/rust/raw/3ca41e2/RELEASES.md | rg --replace='newRelease("$1", "$2"),' -o 'Version (\S+) \(([^)]+)\)'
```
loosebazooka, line 39: does `3ca41e2` point to a tag or any human friendly id? I imagine it doesn't really matter.

loosebazooka (approved):

pkg/registry/cratesio/rust.go
msuozzo: commit. it's the 7-character sommit SHA prefix that GitHub supports elsewhere in its URLs/APIs

loosebazooka: sorry what I meant was how easy was it for someone to audit this quickly?

msuozzo: what do you mean audit? we can either use commits or tags and only one is immutable / reproducible and that's commits.

loosebazooka: audit was the wrong word. just a quick lookup I guess on updates like this. I could just compare position in the commit history of the two commits which is fine. But I would maybe ideally want the commit to be on a "main"/"release" branch?

msuozzo: eh I get where you're coming from. I guess I wasn't viewing this as anything more than a convenience. you could look at the main branch version of this file (the release notes) and view them as plain text. additionally, the output is itself human-readable and pretty easy to manually verify against the official release notes.

PR #646, opened 2025-07-21
Assert rust toolchain has MUSL support

msuozzo:
While not ideal, this approach is a quick way to ensure the generated
strategies are compatible with the default alpine container used in
crate builds.


pkg/rebuild/cratesio/infer.go
```diff
@@ -239,6 +239,14 @@ func (Rebuilder) InferStrategy(ctx context.Context, t rebuild.Target, mux rebuil
 			return nil, errors.New("rust version heuristic failed")
 		}
 	}
+	// TODO: This should be moved to build-time since strategies are intended to be, at least notionally, arch-independent.
+	hasMUSLBuild, err := reg.HasMUSLBuild(rustVersion)
+	if err != nil {
+		return nil, errors.New("rust version heuristic failed")
```
loosebazooka: how does one debug off the error in 245? its the same error as 239. Should you add the error returned from HasMUSLBuild in here? Should it say "while checking HasMUSLBuild"?

pkg/registry/cratesio/rust.go
```diff
@@ -11,16 +11,17 @@ import (
...
 type release struct {
-	Version string
-	Date    time.Time
+	Version   string
+	Date      time.Time
+	MUSLBuild bool
 }
 
-func newRelease(version, dateStr string) release {
+func newRelease(version, dateStr string, muslBuild bool) release {
```
loosebazooka, line 19: `bool` in method parameters is a bit odd. Would it make sense to have both
- `newRelease(version,data)`
- `newReleseWithMUSL(version,data)`
which both populate the release struct differently?

loosebazooka (commented):
It's not really clear to me what the rust versioning strategy is

pkg/registry/cratesio/rust.go
msuozzo: meh it's generated by the command referenced in the comment below so I figure it's not necessarily meant for actual use, just codegen ease

PR #647, opened 2025-07-21
Change dir for crates package build

msuozzo:
This diverges from the prior attempts to use the --package flag after
too many issues with cargo falling back to building the wrong package(s)
in the presence of virtual manifests.


loosebazooka:
is there a rust person someone that can double check this?

msuozzo:
> is there a rust person someone that can double check this?

nope! fundamentally, the rebuild outcomes are the main indicators of success and this appears to address the issues with the status quo.

PR #654, opened 2025-07-24
Add RAVE as a related project

johnbillion:
I maintain the [RAVE for WordPress](https://github.com/johnbillion/rave-wordpress) project which reproduces WordPress from source and verifies that various official and unofficial packages, source code mirrors, and builds match it exactly.

RAVE in its current form is specific to WordPress and runs only as a set of GitHub Actions workflows, but perhaps others will be interested in its approach.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,7 @@ Check out these related projects contributing to the reproducible builds effort:
   Java, Kotlin reproducibility.
 - [kpcyrd/rebuilderd](https://github.com/kpcyrd/rebuilderd): Rebuild scheduler
   with support for several distros.
+- [johnbillion/rave-wordpress](https://github.com/johnbillion/rave-wordpress): Reproduces and verifies official and unofficial WordPress packages.
 
 ## Disclaimer
 


johnbillion:
This is for the core WordPress software. It's specific to all of its oddities, source code mirrors, package mirrors, and distribution channels.

Essentially RAVE checks out the source code, runs the build process, then verifies that packages of WordPress which are distributed via official and unofficial platforms are all legit. There's a fair bit more info in the project readme about why it's needed.

msuozzo:
Gotcha and thanks for taking the time to explain. I am a huge fan of the work and am not opposed to adding this link. I do, however, have some hesitation about this 'related work' section currently being scoped to reproducing package ecosystems and this new addition departing from that pattern.

Looking elsewhere, I found [this page](https://reproducible-builds.org/tools/#orchestrators) (or perhaps a [Success Stories](https://reproducible-builds.org/success-stories/) entry) which could be a cleaner venue to present the project ([repo](https://salsa.debian.org/reproducible-builds/reproducible-website), [contributing](https://reproducible-builds.org/contribute/salsa/). Admittedly, it's a bit overwhelming as-is so there's probably room to carve off a separate "Projects" section especially since, I now am noticing, OSS Rebuild itself is not mentioned there either :)

PR #655, opened 2025-07-24
Use key resource name as signer-verifier param

msuozzo:
Requiring the key object requires external users to have the
cloudkms.cryptoKeyVersion.get permission which is not part of the
cloudkms.verifier role we grant to all users.

Accepting the resource name alone means the interface only need fetch
the public key (cloudkms.cryptoKeyVersion.getPublicKey).

### DIFF
--- a/pkg/kmsdsse/kmsdsse.go
+++ b/pkg/kmsdsse/kmsdsse.go
@@ -10,6 +10,7 @@ import (
 	"crypto/sha256"
 	"crypto/x509"
 	"encoding/pem"
+	"regexp"
 
 	kms "cloud.google.com/go/kms/apiv1"
 	"cloud.google.com/go/kms/apiv1/kmspb"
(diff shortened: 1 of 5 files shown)


pkg/kmsdsse/kmsdsse.go
```diff
@@ -19,14 +20,23 @@ import (
...
-func NewCloudKMSSignerVerifier(ctx context.Context, c *kms.KeyManagementClient, k *kmspb.CryptoKeyVersion) (*CloudKMSSignerVerifier, error) {
+// keyVersionRegex is a compiled regular expression to validate the format of a
+// Google Cloud KMS key version resource name.
+var keyVersionRegex = regexp.MustCompile(
+	`^projects/[^/]+/locations/[^/]+/keyRings/[^/]+/cryptoKeys/[^/]+/cryptoKeyVersions/[^/]+$`,
+)
+
+func NewCloudKMSSignerVerifier(ctx context.Context, c *kms.KeyManagementClient, key string) (*CloudKMSSignerVerifier, error) {
+	if !keyVersionRegex.MatchString(key) {
+		return nil, errors.Errorf("invalid key format: %q; expected format: %s", key, keyVersionRegex.String())
```
loosebazooka:
```diff
- invalid key format:
+ invalid key version:
```

maybe?

pkg/kmsdsse/kmsdsse.go
```diff
@@ -40,7 +50,12 @@ func NewCloudKMSSignerVerifier(ctx context.Context, c *kms.KeyManagementClient,
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to parse PEM public key")
 	}
-	return &CloudKMSSignerVerifier{c, k, pubpb, pub}, nil
+	return &CloudKMSSignerVerifier{
+		client: c,
```
loosebazooka: what do you need the client for if you've already obtained the key?

pkg/kmsdsse/kmsdsse.go
```diff
@@ -19,14 +20,23 @@ import (
 
 type CloudKMSSignerVerifier struct {
 	client *kms.KeyManagementClient
-	key    *kmspb.CryptoKeyVersion
+	key    string
```
loosebazooka: rename to keyVersion (and maybe everywhere else?)

pkg/kmsdsse/kmsdsse.go
loosebazooka: nvm, it's also the signer

pkg/kmsdsse/kmsdsse.go
```diff
@@ -19,14 +20,23 @@ import (
...
 	pubpb  *kmspb.PublicKey
 	pub    crypto.PublicKey
 }
 
-func NewCloudKMSSignerVerifier(ctx context.Context, c *kms.KeyManagementClient, k *kmspb.CryptoKeyVersion) (*CloudKMSSignerVerifier, error) {
+// keyVersionRegex is a compiled regular expression to validate the format of a
+// Google Cloud KMS key version resource name.
+var keyVersionRegex = regexp.MustCompile(
+	`^projects/[^/]+/locations/[^/]+/keyRings/[^/]+/cryptoKeys/[^/]+/cryptoKeyVersions/[^/]+$`,
```
loosebazooka, line 30: add test for regex

PR #668, opened 2025-07-31
Use the proper pathsep for different AssetStores

### LINKED ISSUE #666: Bug: `get` command fails with `storage: object doesn't exist` even for packages and versions listed by `list`
Hello,

I've been using the oss-rebuild tool and have found a bug where the get command consistently fails with a "storage: object doesn't exist" error, even for packages and versions that are successfully listed by the list command.

This issue has been reproduced on multiple different packages, including a popular one like requests, suggesting a problem with the tool's path resolution or data consistency.

Steps to Reproduce

Use the list command to find packages and versions that have attestations. The command successfully returns a list of paths.
    Bash

    oss-rebuild list pypi requests

(The output confirms the existence of attestations for various versions.)

Select one of the listed versions (e.g., 2.32.4) and run the get command.
Bash

    oss-rebuild get pypi requests 2.32.4

Additionally, I have tried the official example from the documentation, which also failed.
Bash

    oss-rebuild get pypi absl-py 2.0.0

Expected Behavior

The get command should successfully retrieve and print the attestation file (rebuild.intoto.jsonl) for the specified package and version.

Actual Behavior

Both get commands fail with the following error, despite the list command indicating the requests asset exists:

NOTE:  artifact is being inferred as "requests-2.32.4-py3-none-any.whl"
Error: creating attestation reader: creating GCS reader for pypi\requests\2.32.4\requests-2.32.4-py3-none-any.whl\rebuild.intoto.jsonl: storage: object doesn't exist
asset not found

msuozzo:
Fixes #666

GCS paths should unconditionally use / as pathsep while filesystem-based
stores can defer to the local machine's default.

### DIFF
--- a/pkg/rebuild/rebuild/storage.go
+++ b/pkg/rebuild/rebuild/storage.go
@@ -10,6 +10,7 @@ import (
 	"io/fs"
 	"net/url"
 	"os"
+	"path"
 	"path/filepath"
 	"strings"
 
@@ -71,12 +72,12 @@ type Asset struct {
 }
 
 // assetPath describes the general layout of assets shared by most hierarchy-based AssetStore types. The runID
-func assetPath(a Asset, runID string) string {
+func assetPath(a Asset, runID string) []string {
 	name := string(a.Type)
 	if a.Type == RebuildAsset {
 		name = a.Target.Artifact
 	}
-	return filepath.Join(string(a.Target.Ecosystem), a.Target.Package, a.Target.Version, a.Target.Artifact, runID, name)
+	return []string{string(a.Target.Ecosystem), a.Target.Package, a.Target.Version, a.Target.Artifact, runID, name}
 }
 
 // ReadOnlyAssetStore is a storage mechanism for debug assets.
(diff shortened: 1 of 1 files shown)


pkg/rebuild/rebuild/storage.go
```diff
@@ -210,7 +211,7 @@ type FilesystemAssetStore struct {
 
 // TODO: Maybe this should include a runID?
 func (s *FilesystemAssetStore) resourcePath(a Asset) string {
-	return assetPath(a, s.runID)
+	return filepath.Join(assetPath(a, s.runID)...)
```
wbxyz, line 218: How does this work with memfs? Separator match the underlying system separator? I see that the impl for memfs.Join() just uses filepath.Join anyway, so maybe it doesn't matter?

wbxyz, line 218: Maybe this should use fs.Join()? https://pkg.go.dev/github.com/go-git/go-billy/v5#Basic

msuozzo: yeah i think that's desirable. if it's an fs, we should consider it an fs according to the local system

msuozzo: I considered this but it notably does not appear to be part of `io/fs.FS` which I'd hope we could eventually use. i know billy has planned to converge their implementation to the builtin abstraction but, in lieu of that being released, it feels safer to rely on the builtin interfaces where we can do so.

Issue #110, opened 2024-09-26
Improve support for Cargo.lock file rebuilds


msuozzo:
Our current support for Rust crate rebuilds is a bit hand-wavy when it comes to lockfiles: We make the `Cargo.lock` content an input to our build (at substantial cost to build definition size) so as to avoid trying to reproduce it.

However discussing this issue with @kpcyrd, we may have stumbled on to a way to potentially make lockfile generation stable:

---

The cargo registry is effectively just a [git repo](https://github.com/rust-lang/crates.io-index) with a newer [index API](https://doc.rust-lang.org/nightly/cargo/reference/registry-index.html) that provides a caching layer.

* _Before the introduction of the index_: The registry repo was present on disk and infrequently updated.
* _After the index_: The registry state is accessed via the index API and is close to the repo head.

At lockfile construction time, the registry's git repo state is accessed (directly or via the index API) to resolve a crate's dependency graph and store it in `Cargo.lock`.

So what we should be able to do is **look for a point in the git repo's history where this resolution should have been possible**.

For our purposes, we can somewhat arbitrarily pick the **earliest commit** in the registry git repo which contains all relevant dependency versions.

---

As for technical feasibility, the repo is currently 160MB bare and ~2GB checked out and has a linear history of ~15k commits. These aren't particularly big numbers aside from the checkout size but, luckily, we should be able to avoid materializing the full repo while still quickly searching the history.

An unoptimized prototype shows us able to search ~20 commits per second which, at the current rate of ~1000 commits/day, would put us at ~40s for a day of registry state. We can almost certainly improve this but, even now, it should be viable to introduce this into our inference pipeline, especially considering we expect the crate publish time to be quite close (well within 1d) to the registry state time in the common case.

msuozzo:
Quick update: #140 improved the search rate such that we should be able to find the target registry commit in about a tenth the time initially quoted (~7s).

bjorn3:
> and has a linear history of ~15k commits

PR #673, opened 2025-08-06
Add library exposing cargo registry inference algorithm

msuozzo:
Port of the algorithm from [./tools/registryscan/](https://github.com/google/oss-rebuild/blob/main/tools/registryscan/main.go)

Work towards #110


pkg/cargolock/cargolock.go
```diff
@@ -0,0 +1,275 @@
...
+		// Update best result
+		bestResult = result
+		// If we found a boundary within this repo, we're done
+		if result.PriorCommit != nil {
+			break
+		}
+		lastResult = result
+	}
+	if bestResult == nil || bestResult.ResolutionCommit == nil {
+		return nil, nil
```
wbxyz: Maybe this should be an error?

pkg/cargolock/cargolock.go
```diff
@@ -0,0 +1,275 @@
...
+			Path:    getPackageFilePath(pkg.Name),
+		}
+	}
+	// Use the existing implementation
+	result, err := findCommitWithVersions(index, internalPackages, cratePublished)
+	if err != nil {
+		return nil, err
+	}
+	if result == nil || result.ResolutionCommit == nil {
+		return nil, nil
```
wbxyz: Same, maybe error?

pkg/cargolock/cargolock.go
```diff
@@ -0,0 +1,275 @@
...
+}
+
+// Package represents a crate dependency from a Cargo.lock file.
+type Package struct {
+	Name    string
+	Version string
+}
+
+// Parse extracts package information from Cargo.lock content.
+func Parse(content string) ([]Package, error) {
```
wbxyz: Should the lock file parsing be separate from the index search algorithm?

wbxyz (approved):

pkg/cargolock/cargolock.go
msuozzo: removed the error case

pkg/cargolock/cargolock.go
msuozzo: removed the error case

wbxyz (approved):

PR #674, opened 2025-08-07
Add different types of diff that cause reproducibility differences in Java

algomaster99:
One of the goal of such a documentation is to build context for an AI model that would help explain the diff and what caused it. The documentation aims to contain only some examples of diff [as the rest (~12K)](https://arxiv.org/abs/2504.21679) is expected to be inferred by the model.

### DIFF
--- a/pkg/rebuild/anonymous_class.json
+++ b/pkg/rebuild/anonymous_class.json
@@ -1,0 +1,276 @@
+{
+  "Count": 27,
+  "Updated": "2025-10-16T16:03:16.469407",
+  "Packages": [
+    {
+      "Ecosystem": "maven",
+      "Name": "org.apache.tika:tika-core",
+      "Versions": [
+        "2.9.2"
+      ],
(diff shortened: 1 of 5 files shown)


algomaster99:
Thinking to reorganize the doc so that the causes are the headers (rather than artifact ID). It won't make a difference to the AI model, but I am also trying to make the doc as human readable friendly as possible.

algomaster99:
Write a line on what the issue is and what is the fix. Helps to give more context.

algomaster99:
Used the following regex to find the specific pattern in diff

1. `^.*(-|\+)[^#]*String.valueOf`
2. `^.*(-|\+)[^#]*\$values\(\);`
3. `^.*(-|\+)[^#]*static final class`
4. `^.*(-|\+)[^#]*Objects.requireNonNull\(`

Don't have a benchmark for duplicate class entries because I could not think of an easy regex expression.

PR #676, opened 2025-08-07
fix: release URL of a certain Maven artifact

algomaster99:
`ResolveReference` API omitted the the path present in the registry URL (`/maven2` of `https://repo1.maven.org/maven2`) so instead we use `JoinPath` to retain it.

### DIFF
--- a/pkg/registry/maven/maven.go
+++ b/pkg/registry/maven/maven.go
@@ -170,6 +170,6 @@ func (r HTTPRegistry) ReleaseURL(ctx context.Context, pkg, version, typ string) 
 		return "", errors.New("package identifier not of form 'group:artifact'")
 	}
 	artifactPath := path.Join(strings.ReplaceAll(g, ".", "/"), a, version, fmt.Sprintf("%s-%s%s", a, version, typ))
-	artifactURL := releaseURL.JoinPath(artifactPath).String()
-	return artifactURL, nil
+	artifactURL := releaseURL.JoinPath(artifactPath)
+	return artifactURL.String(), nil
 }


pkg/registry/maven/maven_test.go
```diff
@@ -0,0 +1,21 @@
...
+package maven
+
+import (
+	"context"
+	"testing"
+)
+
+func TestReleaseURL(t *testing.T) {
+	r := HTTPRegistry{}
+	url, err := r.ReleaseURL(context.Background(), "com.google.guava:guava", "33.4.8-jre", TypePOM)
```
msuozzo: would you mind making this a parameterized test (i.e. loop through a list of test cases)? Even if it's just 1 or 2 examples, it makes it a lot easier to incrementally add new cases.

pkg/registry/maven/maven.go
```diff
@@ -170,6 +170,6 @@ func (r HTTPRegistry) ReleaseURL(ctx context.Context, pkg, version, typ string)
 		return "", errors.New("package identifier not of form 'group:artifact'")
 	}
 	artifactPath := path.Join(strings.ReplaceAll(g, ".", "/"), a, version, fmt.Sprintf("%s-%s%s", a, version, typ))
-	artifactURL := releaseURL.ResolveReference(urlx.MustParse(artifactPath))
-	return artifactURL.String(), nil
+	artifactURL := releaseURL.JoinPath(artifactPath).String()
```
msuozzo: nit: i think the idea of having the "artifactURL" variable name was that it was of URL type. stringifying on this line isn't wrong but I might prefer keeping the String() where it was

msuozzo (approved):

pkg/registry/maven/maven.go
algomaster99: Makes sense. I will fix.

algomaster99:
@msuozzo done ready for merge.

PR #683, opened 2025-08-08
feat: infer java version

algomaster99:
`MANIFEST` attributes may not always be present so we need some way to still get the Java version. In this PR, I propose to infer the the class file version from the bytecode itself. This ensures that we get the compiler version which can atleast compile but does not guarantee reproducibility.

Try this command out where none of the attributes were present:
```
go run tools/ctl/ctl.go infer --ecosystem maven --package com.github.ldapchai:ldapchai --version 0.8.6 --artifact ldapchai-0.8.6.jar
```
> The actual JDK this project is[ built with is 22](https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/com/github/ldapchai/ldapchai-0.8.6.buildspec), but that is determined by [conversation with the author](https://github.com/ldapchai/ldapchai/issues/32).

Please note that I also added an attribute `Build-Jdk-Spec:` which is mentioned in the [AROMA paper](https://dl.acm.org/doi/pdf/10.1145/3643764).  (let me know if I should split them up)


algomaster99:
Will make it ready for review when I add tests for this.

pkg/rebuild/maven/infer.go
```diff
@@ -168,12 +168,63 @@ func getJarJDK(ctx context.Context, name, version string, mux rebuild.RegistryMu
...
+	if len(classBytes) < 8 {
+		return "", errors.New("class file too short")
+	}
+
+	// Check magic number (0xCAFEBABE)
+	if classBytes[0] != 0xCA || classBytes[1] != 0xFE || classBytes[2] != 0xBA || classBytes[3] != 0xBE {
+		return "", errors.New("invalid class file magic number")
+	}
+
+	// Skip minor version (bytes 4-5) and read major version (bytes 6-7)
```
msuozzo: i think this needs a bit more information. mainly, it'd be useful to mention here that the minor bytecode version is always 0.

pkg/rebuild/maven/infer.go
```diff
@@ -168,12 +168,63 @@ func getJarJDK(ctx context.Context, name, version string, mux rebuild.RegistryMu
...
+
+			return majorVersion, nil
+		}
+	}
+
+	return "", errors.New("no .class files found in jar")
+}
+
+// getClassFileMajorVersion extracts the major version from Java class file bytes
+func getClassFileMajorVersion(classBytes []byte) (string, error) {
```
msuozzo: nit: might be worth returning an int here and formatting as a string at the higher level? notably, i think the choice of minor version (or omission, if that's the choice) is useful to document in the above function moreso than this one.

pkg/rebuild/maven/infer.go
```diff
@@ -168,12 +168,63 @@ func getJarJDK(ctx context.Context, name, version string, mux rebuild.RegistryMu
...
+	}
+
+	// Check magic number (0xCAFEBABE)
+	if classBytes[0] != 0xCA || classBytes[1] != 0xFE || classBytes[2] != 0xBA || classBytes[3] != 0xBE {
+		return "", errors.New("invalid class file magic number")
+	}
+
+	// Skip minor version (bytes 4-5) and read major version (bytes 6-7)
+	// Reference: https://docs.oracle.com/javase/specs/jvms/se21/html/jvms-4.html
+	offsetBetweenBytecodeAndJavaVersion := uint16(44)
```
msuozzo: `bytecodeToVersionOffset` is a bit shorter

pkg/rebuild/maven/infer.go
```diff
@@ -168,12 +168,63 @@ func getJarJDK(ctx context.Context, name, version string, mux rebuild.RegistryMu
...
 			return strings.TrimSpace(value), nil
 		}
 	}
-	return "", nil
+	return approximateJarJDK(zipReader)
+}
+
+func approximateJarJDK(jarZip *zip.Reader) (string, error) {
+	// This is a fallback if the JDK version cannot be determined from the JAR manifest.
+	// It looks for a .class file in the JAR to analyze the bytecode version.
```
msuozzo: might be useful to give a brief summary of bytecode version and how it corresponds to JDK version (i.e. you can target a lower bytecode version from a higher JDK version but not vice versa).

pkg/rebuild/maven/infer.go
```diff
@@ -168,12 +168,63 @@ func getJarJDK(ctx context.Context, name, version string, mux rebuild.RegistryMu
...
+		if strings.HasPrefix(line, "Build-Jdk:") || strings.HasPrefix(line, "Build-Jdk-Spec:") {
 			_, value, _ := strings.Cut(line, ":")
 			return strings.TrimSpace(value), nil
 		}
 	}
-	return "", nil
+	return approximateJarJDK(zipReader)
+}
+
+func approximateJarJDK(jarZip *zip.Reader) (string, error) {
```
msuozzo: nit: this function name describes that we're guessing the jdk but we're doing that above, as well, so this serves as *another* approximation and so shouldn't be named by its purpose but its function. one option would be `inferJDKFromBytecode` as that communicates the content of the heuristic.

pkg/rebuild/maven/infer.go
```diff
@@ -168,12 +168,63 @@ func getJarJDK(ctx context.Context, name, version string, mux rebuild.RegistryMu
...
+	// This is a fallback if the JDK version cannot be determined from the JAR manifest.
+	// It looks for a .class file in the JAR to analyze the bytecode version.
+
```
msuozzo: nit: we tend to avoid whitespace within a function (although llms feel differently)

pkg/rebuild/maven/infer.go
```diff
@@ -168,12 +168,63 @@ func getJarJDK(ctx context.Context, name, version string, mux rebuild.RegistryMu
...
+			defer classFile.Close()
+
+			classBytes, err := io.ReadAll(classFile)
+			if err != nil {
+				continue
+			}
+
+			majorVersion, err := getClassFileMajorVersion(classBytes)
+			if err != nil {
+				continue
```
msuozzo: we might consider logging this case since it feels exceptional enough that we'd like some evidence that it failed. i suppose we could log the others, as well, but if we were to pick one, my vote would be this.

algomaster99:
@msuozzo ping for review

algomaster99:
More ways to infer:
1. if it is a [multi release jar](https://www.baeldung.com/java-multi-release-jar), see if there are folders names `META-INF/versions/<jdk-version>/**/*.class`. The JDK version here would be greater than version of classes in `**/*.class`.
2. Another idea is to rely on date of release so infer the version. JDK version = Most recent LTS JDK before the release date.

msuozzo:
let's avoid including JAR files as test data. we can create synthetic cases to pass in for now and, if we need anything more complex, can put together a small test library.

algomaster99:
@msuozzo I added toy libraries to test it. I git tracked the source (A.java), build manifest (pom.xml), and the jar created. `mvn package` creates many other files in the `target` directory, but I ignore them. I keep the source and manifest to reproduce the jar required for the test. Note that the jar may vary even if you use the exact JDK as I did 21.0.8 Temurin because there will be a difference in modification time.

algomaster99:
A better idea is to mimic the tests done in `pkg/rebuild/cratesio/infer_test.go` to keep similar structure.

algomaster99:
@msuozzo ready for your review.

pkg/rebuild/maven/infer.go
```diff
@@ -168,12 +168,69 @@ func getJarJDK(ctx context.Context, name, version string, mux rebuild.RegistryMu
...
+}
+
+func inferJDKFromBytecode(jarZip *zip.Reader) (int, error) {
+	// This is a fallback if the JDK version cannot be determined from the JAR manifest.
+	// We look for the major version of the first .class file found in the JAR as it corresponds to the *lowest* JDK version that can run the bytecode.
+	// This JDK version can be then used to run the build that would ensure that the sources at least compile successfully.
+	// If the we choose a version lower than this, we risk compilation failure during rebuild.
+
+	for _, file := range jarZip.File {
+		if strings.HasSuffix(file.Name, ".class") && !strings.Contains(file.Name, "$") {
```
msuozzo: can you document why you're excluding $?

pkg/rebuild/maven/infer.go
```diff
@@ -168,12 +168,69 @@ func getJarJDK(ctx context.Context, name, version string, mux rebuild.RegistryMu
...
-	return "", nil
+	jdkInt, err := inferJDKFromBytecode(zipReader)
+	if err != nil {
+		return "", err
+	}
+	return fmt.Sprintf("%d", jdkInt), nil
+}
+
+func inferJDKFromBytecode(jarZip *zip.Reader) (int, error) {
+	// This is a fallback if the JDK version cannot be determined from the JAR manifest.
```
msuozzo: > This is a fallback

this sort of context doesn't really belong _inside_ the function definition since this is about an external use of the function. the function documentation should share what the function does and, briefly, how. the second sentence captures "how" well but I think there's more you could cut here.

additionally, whole-function comments in go are placed on the line above the function and begin with the function name:

// inferJDKFromBytecode identifies the lowest JDK version that can run the provided JAR's bytecode.

you can continue additional content on the following lines of content, if necessary

pkg/rebuild/maven/infer.go
algomaster99: That's a good point. It actually is not necessary. $ in class names is used to indicate inner (anonymous) classes which also encode the bytecode version. Let me fix it.

pkg/rebuild/maven/infer.go
algomaster99: Thanks for these documentation tips. I went on thinking more about it and related it to how I write code in Java. We prioritize adding doc strings to public method and optionally add docstring to private methods.

For `infer.go`, the public methods are documented in the interface so that is okay. I also documented the private methods including `getJarJDK`. It was easier to document it if I split the method into two inference strategies - inference from manifest and bytecode  respectively so please note the refactoring in [`afc71c8` (#683)](https://github.com/google/oss-rebuild/pull/683/commits/afc71c8f34bb065eeb6f534e6f3c0a3c44b888a4). I hope it is okay if I did it here. Did not want to make another PR.

pkg/rebuild/maven/infer.go
```diff
@@ -157,6 +158,22 @@ func getJarJDK(ctx context.Context, name, version string, mux rebuild.RegistryMu
...
+	jdk, err := inferJDKFromManifest(zipReader)
+	if err != nil {
+		return "", err
+	}
+	if jdk != "" {
+		return jdk, nil
+	}
+	jdkInt, err := inferJDKFromBytecode(zipReader)
+	if err != nil {
+		return "", err
```
msuozzo: nit: i think we want to wrap this error and the one above it to give context about where the error came from. we tend to phrase these wraps in the present progressive tense e.g. "doing X" so the error ends up formatted "doing X: no .class files found"

msuozzo (approved):

PR #685, opened 2025-08-08
feat: setup Java environment for MavenBuild strategy

algomaster99:
Parsing only the major version of JDK and then using Alpine's package manager to download the JDK. `apk add openjdk{{.With.version}}` works for the LTS 8, 11, 17, and 21, not sure about the rest. If we want to be more fine-grained, we can use links in https://wiki.openjdk.org/display/JDKUpdates/Archived+Releases to get JDK for a specific version (including patch version).

### DIFF
--- a/pkg/rebuild/maven/strategy.go
+++ b/pkg/rebuild/maven/strategy.go
@@ -4,6 +4,9 @@
 package maven
 
 import (
+	"strconv"
+	"strings"
+
 	"github.com/google/oss-rebuild/pkg/rebuild/flow"
 	"github.com/google/oss-rebuild/pkg/rebuild/rebuild"
 )
@@ -24,14 +27,52 @@ func (b *MavenBuild) ToWorkflow() *rebuild.WorkflowStrategy {
 			Uses: "git-checkout",
 		}},
 		Deps: []flow.Step{{
-			Runs: "true",
+			Uses: "maven/deps/basic",
+			With: map[string]string{
+				"version": getOnlyMajorVersion(b.JDKVersion),
+			},
 		}},
 		Build: []flow.Step{{
-			Uses: "true",
+			Runs: "echo 'Building Maven project'",
 		}},
 	}
 }
 
 func (b *MavenBuild) GenerateFor(t rebuild.Target, be rebuild.BuildEnv) (rebuild.Instructions, error) {
 	return b.ToWorkflow().GenerateFor(t, be)
+}
+
+func init() {
+	for _, t := range toolkit {
+		flow.Tools.MustRegister(t)
+	}
+}
+
+var toolkit = []*flow.Tool{
+	{
+		Name: "maven/setup-java",
+		Steps: []flow.Step{{
+			Runs: "apk add openjdk{{.With.version}}",
+		}},
+	},
+
+	{
+		Name: "maven/deps/basic",
+		Steps: []flow.Step{
+			{
+				Uses: "maven/setup-java",
+				With: map[string]string{
+					"version": "{{.With.version}}",
+				},
+			},
+		},
+	},
+}
+
+func getOnlyMajorVersion(version string) string {
+	if major, err := strconv.Atoi(version); err == nil {
+		return strconv.Itoa(major)
+	}
+	parts := strings.Split(version, ".")
+	return parts[0]
 }


algomaster99:
Found [Java 8 download URLs here](https://www.oracle.com/java/technologies/javase/javase8-archive-downloads.html) but they are behind an authentication wall.

algomaster99:
Temurin [GitHub release page](https://github.com/adoptium/temurin8-binaries/releases) hosts Java 8 versions but they are mostly early access.

pkg/rebuild/maven/strategy.go
```diff
@@ -24,14 +29,63 @@ func (b *MavenBuild) ToWorkflow() *rebuild.WorkflowStrategy {
...
+		Name: "maven/setup-java",
+		Steps: []flow.Step{{
+			Runs: textwrap.Dedent(`
+			mkdir -p /opt/jdk
+			wget -q -O - "{{.With.versionURL}}" | tar -xzf - --strip-components=1 -C /opt/jdk
+			export JAVA_HOME=/opt/jdk
+			export PATH=$JAVA_HOME/bin:$PATH`),
+			Needs: []string{"wget"},
+		}},
+	},
```
wbxyz: Can you add a test case in strategy_test.go to demonstrate the evaluated instructions?

pkg/rebuild/maven/strategy.go
```diff
@@ -24,14 +29,63 @@ func (b *MavenBuild) ToWorkflow() *rebuild.WorkflowStrategy {
...
+				With: map[string]string{
+					"versionURL": "{{.With.versionURL}}",
+				},
+			},
+		},
+	},
+}
+
+func getVersionURL(version string) string {
+	versionToURLCSV, err := os.ReadFile("pkg/rebuild/maven/versionToUrl.csv")
```
wbxyz: Is this portable? I would imagine this could break depending on where you run this code. WDYT of making that CSV file a golang file that's a map constant. That way it's compiled straight into the binary.

pkg/rebuild/maven/strategy_test.go
```diff
@@ -0,0 +1,57 @@
...
+					Dir:  "ldapchai",
+				},
+				SystemDeps: []string{"git", "wget"},
+				Source:     "git clone https://github.com/ldapchai/ldapchai.git .\ngit checkout --force 'a9de4ccc8db9a4862f3819f3dfb63e57a6450bdf'",
+				Deps: `mkdir -p /opt/jdk
+wget -q -O - "https://adoptium.net/download?link=https%3A%2F%2Fgithub.com%2Fadoptium%2Ftemurin8-binaries%2Freleases%2Fdownload%2Fjdk8u462-b08%2FOpenJDK8U-jdk_x64_linux_hotspot_8u462b08.tar.gz&vendor=Adoptium" | tar -xzf - --strip-components=1 -C /opt/jdk
+export JAVA_HOME=/opt/jdk
+export PATH=$JAVA_HOME/bin:$PATH`,
+				Build:      "echo 'Building Maven project'", // TODO: Replace with actual Maven build command
+				OutputPath: "ldapchai-0.8.6.jar",
```
algomaster99, line 43: These are still dummy values as I am yet to infer the build command.

pkg/rebuild/maven/strategy.go
```diff
@@ -24,14 +26,55 @@ func (b *MavenBuild) ToWorkflow() *rebuild.WorkflowStrategy {
...
+}
+
+var toolkit = []*flow.Tool{
+	{
+		Name: "maven/setup-java",
+		Steps: []flow.Step{{
+			Runs: textwrap.Dedent(`mkdir -p /opt/jdk
+wget -q -O - "{{.With.versionURL}}" | tar -xzf - --strip-components=1 -C /opt/jdk
+export JAVA_HOME=/opt/jdk
+export PATH=$JAVA_HOME/bin:$PATH`),
```
wbxyz: If you want to use textwrap.Dedent you can add a newline at the beginning and then slice that char off using subscript. Something like this:

```
    textwrap.Dedent(`
          foobar
          baz`[1:])
```

As is, I think dedent isn't really necessary because there's no indentation.

pkg/rebuild/maven/strategy.go
```diff
@@ -24,14 +26,55 @@ func (b *MavenBuild) ToWorkflow() *rebuild.WorkflowStrategy {
...
+		},
+	},
+}
+
+func getVersionURL(version string) string {
+	url, exists := JDKDownloadURLs[version]
+	if exists {
+		return url
+	}
+	return ""
```
wbxyz: If we don't know where the URL is, maybe that should be an error somewhere? Returning an empty string and letting the strategy evaluate further seems like it's going to bury the failure deep in a GCB execution.

One option would be to do this URL evaluation during inference and have the MavenBuild struct store the URL instead of just the version number. Another option is to check during GenerateFor, and have that return an error saying it can't find an appropriate JDK? WDYT?

pkg/rebuild/maven/strategy_test.go
wbxyz: Yup dummy values lgtm.

pkg/rebuild/maven/jdk_urls.go
```diff
@@ -0,0 +1,68 @@
...
+	"11.0.2":   "https://download.java.net/java/GA/jdk11/9/GPL/openjdk-11.0.2_linux-x64_bin.tar.gz",
+	"11.0.1":   "https://download.java.net/java/GA/jdk11/13/GPL/openjdk-11.0.1_linux-x64_bin.tar.gz",
+	"11":       "https://download.java.net/java/ga/jdk11/openjdk-11_linux-x64_bin.tar.gz",
+	"10.0.2":   "https://download.java.net/java/GA/jdk10/10.0.2/19aef61b38124481863b1413dce1855f/13/openjdk-10.0.2_linux-x64_bin.tar.gz",
+	"10.0.1":   "https://download.java.net/java/GA/jdk10/10.0.1/fb4372174a714e6b8c52526dc134031e/10/openjdk-10.0.1_linux-x64_bin.tar.gz",
+	"10":       "https://download.java.net/java/GA/jdk10/10/binaries/openjdk-10_linux-x64_bin.tar.gz",
+	"9.0.4":    "https://download.java.net/java/GA/jdk9/9.0.4/binaries/openjdk-9.0.4_linux-x64_bin.tar.gz",
+	"9.0.1":    "https://download.java.net/java/GA/jdk9/9.0.1/binaries/openjdk-9.0.1_linux-x64_bin.tar.gz",
+	"9":        "https://download.java.net/java/GA/jdk9/9/binaries/openjdk-9_linux-x64_bin.tar.gz",
+	"8":        "https://adoptium.net/download?link=https%3A%2F%2Fgithub.com%2Fadoptium%2Ftemurin8-binaries%2Freleases%2Fdownload%2Fjdk8u462-b08%2FOpenJDK8U-jdk_x64_linux_hotspot_8u462b08.tar.gz&vendor=Adoptium",
```
wbxyz: Is this link going through a proxy?

I think it eventually resolves to this one: https://github.com/adoptium/temurin8-binaries/releases/download/jdk8u462-b08/OpenJDK8U-jdk_x64_linux_hotspot_8u462b08.tar.gz

pkg/rebuild/maven/jdk_urls.go
```diff
@@ -0,0 +1,68 @@
...
+	"8u411":    "https://cfdownload.adobe.com/pub/adobe/coldfusion/java/java8/java8u411/jdk/jdk-8u411-linux-x64.tar.gz",
+	"8u401":    "https://cfdownload.adobe.com/pub/adobe/coldfusion/java/java8/java8u401/jdk/jdk-8u401-linux-x64.tar.gz",
+	"8u391":    "https://cfdownload.adobe.com/pub/adobe/coldfusion/java/java8/java8u391/jdk/jdk-8u391-linux-x64.tar.gz",
+	"8u381":    "https://cfdownload.adobe.com/pub/adobe/coldfusion/java/java8/java8u381/jdk/jdk-8u381-linux-x64.tar.gz",
+	"8u371":    "https://cfdownload.adobe.com/pub/adobe/coldfusion/java/java8/java8u371/jdk/jdk-8u371-linux-x64.tar.gz",
+	"8u361":    "https://cfdownload.adobe.com/pub/adobe/coldfusion/java/java8/java8u361/jdk/jdk-8u361-linux-x64.tar.gz",
+	"8u351":    "https://cfdownload.adobe.com/pub/adobe/coldfusion/java/java8/java8u351/jdk/jdk-8u351-linux-x64.tar.gz",
+	"8u341":    "https://cfdownload.adobe.com/pub/adobe/coldfusion/java/java8/java8u341/jdk/jdk-8u341-linux-x64.tar.gz",
+	"8u333":    "https://cfdownload.adobe.com/pub/adobe/coldfusion/java/java8/java8u333/jdk/jdk-8u333-linux-x64.tar.gz",
+	"8u331":    "https://cfdownload.adobe.com/pub/adobe/coldfusion/java/java8/java8u331/jdk/jdk-8u331-linux-x64.tar.gz",
```
algomaster99: document how we got them

pkg/rebuild/maven/strategy.go
algomaster99: I let `MavenBuild` have the version number, and not its URL because I think it is better to keep the inference logic independent of that. It is under strategy that we should decide whether we want to use a tarball or an `apt/apk` package.

pkg/rebuild/maven/jdk_urls.go
algomaster99: @msuozzo documented it by linking it to an #686 .

pkg/rebuild/maven/jdk_urls.go
```diff
@@ -0,0 +1,70 @@
...
+	"11":       "https://download.java.net/java/ga/jdk11/openjdk-11_linux-x64_bin.tar.gz",
+	"10.0.2":   "https://download.java.net/java/GA/jdk10/10.0.2/19aef61b38124481863b1413dce1855f/13/openjdk-10.0.2_linux-x64_bin.tar.gz",
+	"10.0.1":   "https://download.java.net/java/GA/jdk10/10.0.1/fb4372174a714e6b8c52526dc134031e/10/openjdk-10.0.1_linux-x64_bin.tar.gz",
+	"10":       "https://download.java.net/java/GA/jdk10/10/binaries/openjdk-10_linux-x64_bin.tar.gz",
+	"9.0.4":    "https://download.java.net/java/GA/jdk9/9.0.4/binaries/openjdk-9.0.4_linux-x64_bin.tar.gz",
+	"9.0.1":    "https://download.java.net/java/GA/jdk9/9.0.1/binaries/openjdk-9.0.1_linux-x64_bin.tar.gz",
+	"9":        "https://download.java.net/java/GA/jdk9/9/binaries/openjdk-9_linux-x64_bin.tar.gz",
+	"8":        "https://github.com/adoptium/temurin8-binaries/releases/download/jdk8u462-b08/OpenJDK8U-jdk_x64_linux_hotspot_8u462b08.tar.gz",
+
+	// See https://github.com/google/oss-rebuild/issues/686 for how we ge this list and how to improve it further.
```
wbxyz:
```suggestion
	// See https://github.com/google/oss-rebuild/issues/686 for how we got this list and how to improve it further.
```

algomaster99:
@wbxyz CI is green. Let's merge? A follow up task here is to add strategy for Maven.

pkg/rebuild/maven/jdk_urls.go
msuozzo: Can you move a short version of this inline as a comment? Might also be nice to see if there's a more authoritative source for the adobe links.

reason: I'd prefer to keep at least the main explanation of things inside the git repo itself (i.e. not in issues).

pkg/rebuild/maven/strategy.go
```diff
@@ -17,21 +20,71 @@ type MavenBuild struct {
...
+				Uses: "maven/setup-java",
+				With: map[string]string{
+					"versionURL": "{{.With.versionURL}}",
+				},
+			},
+		},
+	},
+}
+
+func getVersionURL(version string) (string, error) {
```
msuozzo: Move to infer.go (and maybe inline since this is no simpler than a check-fail for map access)

algomaster99: So I thought about that but went with implementing in strategy  because I wanted the inference to return the Java number and keep it away from the implementation detail of how we are going to install JDK. Only strategy should care about the implementation. In future, in case we decide to install JDK via `apt`, we only need to change the strategy.

algomaster99:
The tests are failing. It is because we infer JDK 8 for ldapchai, but we don't have 8 anymore. I know that ldapchai 0.8.6 was built with 22. Let me look around the jar and see how I can infer a higher JDK.

pkg/rebuild/maven/strategy.go
```diff
@@ -17,21 +20,63 @@ type MavenBuild struct {
...
+
+var toolkit = []*flow.Tool{
+	{
+		Name: "maven/setup-java",
+		Steps: []flow.Step{{
+			Runs: textwrap.Dedent(`
+				mkdir -p /opt/jdk
+				wget -q -O - "{{.With.versionURL}}" | tar -xzf - --strip-components=1 -C /opt/jdk
+				export JAVA_HOME=/opt/jdk
+				export PATH=$JAVA_HOME/bin:$PATH`[1:]),
```
msuozzo, line 68: add a TODO to confirm whether these persist to the docker run part of the 2-phase docker build->run. it seems like it may not in which can we can split these out into a stub like "maven/export-java" (or even keep it as "setup-java") which can then be used from _both_ Deps and Build phases.

msuozzo (approved):

PR #689, opened 2025-08-14
Add crates index manager abstraction

msuozzo:
The full crates.io index corpus is too large and growing too fast to
realistically assume we can have access to it all at once for inference
purposes. This new component enables us to keep a fixed-size cache of
snapshots on disk while rotating the least used snapshots out when other
snapshots are requested. We unconditionally keep the current (i.e.
non-archived) index cloned given it is the newest and most likely to be
requested during inference.

The interface also supports loading index portions from an existing
filesystem which means it can also be used transiently to mediate access
to these resources.


pkg/registry/cratesio/index/manager.go
```diff
@@ -0,0 +1,332 @@
...
+// initialize ensures the repository is fetched and ready, using sync.Once for coalescing
+func (c *cachedRepository) initialize(ctx context.Context, m *IndexManager) error {
+	c.initOnce.Do(func() {
+		c.initErr = c.doInitialize(ctx, m)
+	})
+	return c.initErr
+}
+
+// doInitialize performs the actual initialization work
+func (c *cachedRepository) doInitialize(ctx context.Context, m *IndexManager) error {
```
wbxyz: Should this take a write lock on the readersMu?

pkg/registry/cratesio/index/manager.go
```diff
@@ -0,0 +1,332 @@
...
+	}
+}
+
+// getRepositoryPath returns the filesystem path for a repository
+func (m *IndexManager) getRepositoryPath(key RepositoryKey) string {
+	switch key.Type {
+	case CurrentIndex:
+		return "current"
+	case SnapshotIndex:
+		return "snapshots/" + key.Name
```
wbxyz: nit: path.join?

pkg/registry/cratesio/index/manager.go
```diff
@@ -0,0 +1,332 @@
...
+// maybeEvictSnapshot evicts the least recently used snapshot if at capacity
+func (m *IndexManager) maybeEvictSnapshot(ctx context.Context) error {
+	// Count current snapshots
+	snapshotCount := 0
+	var lruSnapshot *cachedRepository
+	var lruKey RepositoryKey
+	m.repositories.Range(func(key RepositoryKey, repo *cachedRepository) bool {
+		if key.Type == SnapshotIndex {
+			snapshotCount++
+			if lruSnapshot == nil || repo.lastAccess.Load() < lruSnapshot.lastAccess.Load() {
```
wbxyz: Should this be taking a read lock on these repositories?

Although if there's a write lock inside doInitialize then that wouldn't work.

Would it be possible to maybeEvict from the IndexManager?

pkg/registry/cratesio/index/manager.go
```diff
@@ -0,0 +1,332 @@
...
+func (m *IndexManager) updateRepository(ctx context.Context, repo *cachedRepository) error {
+	// NOTE: If readers are active, this attempt to write lock will block
+	repo.readersMu.Lock()
+	defer repo.readersMu.Unlock()
+	// Update using fetcher
+	repoFs, err := m.fs.Chroot(repo.path)
+	if err != nil {
+		return err
+	}
+	defer repo.lastUpdate.Store(time.Now().UnixMilli())
```
wbxyz: I think this needs to be wrapped in a func to make sure the time is evaluated only once Update is done?

defer func() {repo.lastUpdate.Store(time.Now().UnixMilli())}()

wbxyz (approved):
Baring race conditions in the IndexManager, this lgtm.

pkg/registry/cratesio/index/manager.go
msuozzo: changed sufficiently that I think this is no longer a concern

internal/safememfs/safememfs.go
```diff
@@ -0,0 +1,138 @@
...
+	defer s.mu.Unlock()
+	return s.fs.TempFile(dir, prefix)
+}
+
+// --- Read Operations (use shared RLock) ---
+
+func (s *SafeMemory) Open(filename string) (billy.File, error) {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	return s.fs.OpenFile(filename, os.O_RDONLY, 0)
```
wbxyz: Add a comment about why OpenFile() instead of just Open() (I think to guarantee that mu.RLock is correct)

internal/safememfs/safememfs.go
wbxyz, line 1: Move to separate PR?

internal/safememfs/safememfs.go
```diff
@@ -0,0 +1,138 @@
...
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	return s.fs.TempFile(dir, prefix)
+}
+
+// --- Read Operations (use shared RLock) ---
+
+func (s *SafeMemory) Open(filename string) (billy.File, error) {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
```
wbxyz: Does this lock need to remain until the returned file is closed?

pkg/registry/cratesio/index/manager.go
```diff
@@ -0,0 +1,452 @@
...
+// loadExistingRepo checks a path for a valid git repository and adds it
+func (m *IndexManager) loadExistingRepo(key RepositoryKey) error {
+	path := m.getRepositoryPath(key)
+	repoFs, err := m.fs.Chroot(path)
+	if err != nil && errors.Is(err, fs.ErrNotExist) {
+		return nil // Path doesn't exist, skip
+	}
+	storer := filesystem.NewStorage(repoFs, cache.NewObjectLRUDefault())
+	if _, err := git.Open(storer, nil); err != nil {
+		return nil // Not a valid repo, skip
```
wbxyz: Shouldn't this be a real error? If the path exists and the contents aren't a valid repo, I would expect loadExistingRepo to return an error. It's ok if the caller decides to ignore all the errors I think, but it should probably be returned.

pkg/registry/cratesio/index/manager.go
```diff
@@ -0,0 +1,452 @@
...
+	}
+	return nil
+}
+
+// loadExistingRepo checks a path for a valid git repository and adds it
+func (m *IndexManager) loadExistingRepo(key RepositoryKey) error {
+	path := m.getRepositoryPath(key)
+	repoFs, err := m.fs.Chroot(path)
+	if err != nil && errors.Is(err, fs.ErrNotExist) {
+		return nil // Path doesn't exist, skip
```
wbxyz, line 156: Also maybe return ErrNotExists when appropriate?

pkg/registry/cratesio/index/manager.go
```diff
@@ -0,0 +1,452 @@
...
+
+// handleAcquisitionRequest processes a request to acquire repositories
+func (m *IndexManager) handleAcquisitionRequest(req acquisitionRequest) {
+	ctx, cancel := context.WithCancel(req.ctx)
+	defer cancel() // ensure the context is cleaned up
+	// Construct the sets of existing and needed repositories
+	var existing, missing []*managedRepository
+	seen := make(map[RepositoryKey]bool, len(req.keys))
+	for _, key := range req.keys {
+		if _, exists := seen[key]; exists {
```
wbxyz, line 207: is seen updated somewhere?

pkg/registry/cratesio/index/manager.go
```diff
@@ -0,0 +1,452 @@
...
+			repo.rwMutex.RUnlock()
+		}
+	}
+	// Get read locks for existing
+	for _, repo := range existing {
+		repo.rwMutex.RLock()
+		locked.Store(repo.key, repo)
+		// TODO: This should be executed in parallel with the fetches below
+		if repo.key.Type == CurrentIndex && time.UnixMilli(repo.lastUpdate.Load()).Before(time.Now().Add(-m.currentUpdateInterval)) {
+			if err := m.doUpdate(ctx, repo); err != nil {
```
wbxyz: Does this code need to take a write lock before calling doUpdate?

pkg/registry/cratesio/index/manager.go
```diff
@@ -0,0 +1,452 @@
...
+				return
+			}
+		}
+	}
+	// Fetch the new repositories and get read locks
+	var causeErr error
+	for err := range pipe.ParInto(len(missing), pipe.FromSlice(missing), func(in *managedRepository, out chan<- error) {
+		in.rwMutex.Lock()
+		var err error
+		if foo, exists := m.repositories.LoadOrStore(in.key, in); exists {
```
wbxyz: nit: Is there a better description than foo?

pkg/registry/cratesio/index/manager.go
```diff
@@ -0,0 +1,452 @@
...
+	return nil
+}
+
+// acquireLoop is the main event loop that handles all state changes
+func (m *IndexManager) acquireLoop() {
+	defer m.acquireWg.Done()
+	for {
+		select {
+		case req := <-m.acquisitionCh:
+			m.handleAcquisitionRequest(req)
```
wbxyz, line 192: It might be worth adding a comment mentioning that all handleAcquisitionRequests are serialized so eviction doesn't need to worry about race conditions.

pkg/registry/cratesio/index/manager.go
```diff
@@ -0,0 +1,452 @@
...
+		req.response <- acquisitionResponse{err: causeErr}
+		return
+	}
+	// Get handles
+	var handles []*RepositoryHandle
+	for _, repo := range locked.Iter() {
+		handle, err := m.createRepositoryHandle(repo)
+		if err != nil {
+			unlock()
+			req.response <- acquisitionResponse{err: causeErr}
```
wbxyz: Shouldn't this be returning err instead of causeErr?

pkg/registry/cratesio/index/manager.go
```diff
@@ -0,0 +1,452 @@
...
+			}
+		}
+	}
+	// Fetch the new repositories and get read locks
+	var causeErr error
+	for err := range pipe.ParInto(len(missing), pipe.FromSlice(missing), func(in *managedRepository, out chan<- error) {
+		in.rwMutex.Lock()
+		var err error
+		if foo, exists := m.repositories.LoadOrStore(in.key, in); exists {
+			in = foo
```
wbxyz: nit: Add a comment that foo is expected to be already initialized, and doesn't need a write lock? I don't think it can be in-the-process of being initialized because the acquisition requests are serialized and this handler won't return until all initializations complete or fail (I think).

pkg/registry/cratesio/index/manager.go
```diff
@@ -0,0 +1,452 @@
...
+		} else {
+			err = m.doInitialFetch(ctx, in)
+			if err != nil {
+				m.repositories.Delete(in.key)
+			}
+			in.rwMutex.Unlock()
+		}
+		in.rwMutex.RLock()
```
wbxyz, line 286: This is the Read lock that will be "held" by the handle until release() is called, right? Might be worth a comment.

pkg/registry/cratesio/index/manager.go
```diff
@@ -0,0 +1,452 @@
...
+			default:
+				req.response <- acquisitionResponse{err: errors.Errorf("unknown repository type: %v", key.Type)}
+				return
+			}
+			repo.lastAccess.Store(time.Now().UnixMilli())
+			missing = append(missing, repo)
+		}
+	}
+	// Ensure we have enough space to fetch the new repositories
+	if err := m.evictSnapshotsIfNeeded(ctx, missing, existing); err != nil {
```
wbxyz, line 234: Maybe add context timeout on eviction to detect and report deadlock?

internal/safememfs/safememfs.go
msuozzo: Yeah this is really only valid for the semantics of the Memory filesystem. memfs.New returns the type-erased version, though, so I've just added a comment to the fs type to document that it's intended to only be a memfs

internal/safememfs/safememfs.go
msuozzo: changed to go through s.OpenFile instead

internal/safememfs/safememfs.go
msuozzo: #738

pkg/registry/cratesio/index/manager.go
msuozzo: i think this one is fine. there is no existing repo so we can safely skip it. it's an internal method so that separation of concerns is tolerable.

pkg/registry/cratesio/index/manager.go
msuozzo: yeah this should be an error. done

pkg/registry/cratesio/index/manager.go
msuozzo: i think this is actually WAI since the caller should be the one determining what their tolerance for blocking waits will be. we don't know a priori what behavior the user would want here and passing through ctx is us allowing them to specify whatever policy they want.

pkg/registry/cratesio/index/manager.go
msuozzo: Yeah this is designed to be compatible with removing the serialized processing constraint. The design still applies, though, so I've added comments.

pkg/registry/cratesio/index/manager.go
msuozzo: yep good call. added locking and unlocking for this path.

pkg/registry/cratesio/index/manager.go
wbxyz: Sounds good. In that case I'd weakly suggest we add the timeout logic in registryscan as the example usage but not a strong belief

pkg/registry/cratesio/index/manager.go
wbxyz: Yeah definitely safe. I think was just just hung up on "loadExistingRepo" returning `nil` when it didn't actually load a repo, which seems like unexpected behavior. I guess the right way to check for success though is to look for the key in `m.repositories`.

wbxyz (approved):

PR #690, opened 2025-08-14
Fix race condition in GCB execution

msuozzo:
* Executor.activeBuilds is used to keep track of which builds need to be
  cleaned up on Close.
* handle.setResult is used to signal that Wait can safely return.

The existing logic allowed code like the following to reach Close with
the build still considered active and, thus, necessary to clean up:

  handle.setResult()
  handle.Wait()
  executor.Close() // observes and tries to cancel the completed build from activeBuild
  executor.activeBuilds.Delete()

### DIFF
--- a/pkg/build/gcb/executor.go
+++ b/pkg/build/gcb/executor.go
@@ -166,11 +166,7 @@ func (e *Executor) Close(ctx context.Context) error {
 
 // executeBuild runs the actual build process using Cloud Build
 func (e *Executor) executeBuild(ctx context.Context, handle *gcbHandle, cloudBuild *cloudbuild.Build, input rebuild.Input, opts build.Options, plan *Plan) {
-	defer func() {
-		e.activeBuilds.Delete(handle.id)
-		handle.output.Close()
-		close(handle.outputChan)
-	}()
+	defer close(handle.outputChan)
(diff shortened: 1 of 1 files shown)


pkg/build/gcb/executor.go
```diff
@@ -166,11 +166,7 @@ func (e *Executor) Close(ctx context.Context) error {
 
 // executeBuild runs the actual build process using Cloud Build
 func (e *Executor) executeBuild(ctx context.Context, handle *gcbHandle, cloudBuild *cloudbuild.Build, input rebuild.Input, opts build.Options, plan *Plan) {
-	defer func() {
-		e.activeBuilds.Delete(handle.id)
-		handle.output.Close()
-		close(handle.outputChan)
-	}()
+	defer close(handle.outputChan)
```
wbxyz, line 169: This looks unused, and it's odd that the state is managed by executor. Given that it's unused, can we delete it? Seems unnecessary when we also have handle.output which is a buffered pipe.

Separate PR is fine too.

wbxyz (approved):

pkg/build/gcb/executor.go
msuozzo: yep it was a hold-over from a prior implementation of output buffering that got left behind. removed.

PR #692, opened 2025-08-14
Infer Maven build command

algomaster99:
These changes will aim to infer the build command to require to rebuild a particular project.

### DIFF
--- a/pkg/rebuild/maven/strategy.go
+++ b/pkg/rebuild/maven/strategy.go
@@ -43,7 +43,8 @@ func (b *MavenBuild) ToWorkflow() (*rebuild.WorkflowStrategy, error) {
 			{
 				// TODO: Java 9 needs additional certificate installed in /etc/ssl/certs/java/cacerts
 				// It can be passed to maven command via -Djavax.net.ssl.trustStore=/etc/ssl/certs/java/cacerts
-				Runs:  "mvn clean package -DskipTests",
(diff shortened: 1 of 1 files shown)


pkg/rebuild/maven/strategy.go
```diff
@@ -59,25 +65,22 @@ func init() {
...
 	},
+	{
+		Name: "maven/export-java",
+		Steps: []flow.Step{{
+			Runs: textwrap.Dedent(`
+				export JAVA_HOME=/opt/jdk
+				export PATH=$JAVA_HOME/bin:$PATH
+			`[1:]),
+		}},
+	},
```
algomaster99: @msuozzo implementation detail but do we want to have this or `ENV` instruction in docker file?

msuozzo: no, we don't. the strategy doesn't even necessarily have the assumption that it will be used from a dockerfile. there is a bit of fuzziness around the contract between the two steps but, for now, env vars between src+deps and build are not retained.

algomaster99:
@wbxyz ping for review. Merge of  this PR would allow us to try running it on cloud. As of now, the strategy outputs the artifact and copies it in `/out` - ` mkdir /out && cp /src/target/ldapchai-0.8.6.jar /out/`. I believe this artifact should be compared with the reference on registry. Where does that happen? And is `/out` mounted to the host filesystem to that we can access it later?

PR #697, opened 2025-08-18
Add handling for Maven rebuilds

algomaster99:
The changes in this PR will configure scripts to perform rebuilds on multiple Maven packages.

### DIFF
--- a/internal/api/apiservice/rebuild_test.go
+++ b/internal/api/apiservice/rebuild_test.go
@@ -508,18 +508,18 @@ RLpmHHG1JOVdOA==
 					Response: &http.Response{
 						StatusCode: 200,
 						Body: io.NopCloser(must(archivetest.ZipFile([]archive.ZipEntry{
-							{FileHeader: &zip.FileHeader{Name: "com/example/Foo.class"}, Body: []byte{0xCA, 0xFE, 0xBA, 0xBE, 0x00, 0x00, 0x00, 0x38}},
-							{FileHeader: &zip.FileHeader{Name: "META-INF/MANIFEST.MF"}, Body: []byte("Manifest-Version: 1.0\nCreated-By: 11.0.2 (Oracle Corporation)\n\n")}},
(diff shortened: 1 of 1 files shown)


internal/api/apiservice/rebuild.go
msuozzo, line 1: Would you be able to also add a test for a Maven rebuild here? The ones in this package should be fairly exhaustive and test things end-to-end.

internal/api/apiservice/rebuild_test.go
```diff
@@ -498,6 +499,31 @@ RLpmHHG1JOVdOA==
...
+		{
+			name:   "maven jar success",
+			target: rebuild.Target{Ecosystem: rebuild.Maven, Package: "com.google.guava:guava", Version: "33.4.8-jre", Artifact: "guava-33.4.8-jre.jar"},
+			calls: []httpxtest.Call{
+				{
+					URL: "https://repo1.maven.org/maven2/com/google/guava/guava/33.4.8-jre/guava-33.4.8-jre.jar",
+					Response: &http.Response{
+						StatusCode: 200,
+						Body: io.NopCloser(must(archivetest.ZipFile([]archive.ZipEntry{
+							{FileHeader: &zip.FileHeader{Name: "com/example/Foo.class"}, Body: []byte{0xCA, 0xFE, 0xBA, 0xBE, 0x00, 0x00, 0x00, 0x38}},
```
msuozzo: Nit: I might use the most obvious version inference option so the test is a bit more self-describing

algomaster99: Sure. The most obvious for us is in the MANIFEST. I will update that.

algomaster99: Apparently, the inference is also being spoofed here. So it only matters we give it a JDK that we have a URL for.

msuozzo: Ah unfortunate. We could consider changing that at some point but this looks good then.

PR #701, opened 2025-08-19
Resolve source repository URL from parent

### LINKED ISSUE #700: Infer source URL for parent POM
Source URLs for some Maven packages are in the parent pom. For example, in GSON, the source repository URL is not mentioned in its POM. It is [mentioned in `gson-parent`](https://github.com/google/gson/blob/main/pom.xml#L27) which is the parent.

algomaster99:
Fixes #700 

I prioritize looking for `scm.url` in the `pom` or `parent pom` as that URL points to the source repository. `url`, by convention, is the project home page. For OSS-Rebuild, `url` is https://oss-rebuild.dev/ and not `https://github.com/google/oss-rebuild`.

We also need to resolve the parent pom for to scan for `scm.url` from there.

### DIFF
--- a/pkg/rebuild/maven/infer.go
+++ b/pkg/rebuild/maven/infer.go
@@ -32,7 +32,45 @@ func (Rebuilder) InferRepo(ctx context.Context, t rebuild.Target, mux rebuild.Re
 	if err != nil {
 		return "", err
 	}
+	for pom.Repo() == "" && pom.Parent.ArtifactID != "" {
+		pom, err = resolveParentPom(ctx, pom, t, mux)
+		if err != nil {
+			return "", errors.Errorf("failed to resolve parent POM for %s: %v", t.Package, err)
+		}
+	}
 	return uri.CanonicalizeRepoURI(pom.Repo())
+}
+
+func resolveParentPom(ctx context.Context, pom PomXML, t rebuild.Target, mux rebuild.RegistryMux) (PomXML, error) {
+	if pom.Parent.ArtifactID == "" {
+		return pom, nil
+	}
+	parent, err := NewPomXML(ctx, rebuild.Target{
+		Ecosystem: t.Ecosystem,
+		Package:   fmt.Sprintf("%s:%s", getGroupID(pom, pom.Parent), pom.Parent.ArtifactID),
+		Version:   getVersionID(pom, pom.Parent),
+		Artifact:  t.Artifact,
+	}, mux)
+	if err != nil {
+		return PomXML{}, err
+	}
+	return parent, nil
+}
+
+func getGroupID(pom PomXML, parent Parent) string {
+	// prioritize the groupID of parent
+	if parent.GroupID != "" {
+		return parent.GroupID
+	}
+	return pom.GroupID
+}
+
+func getVersionID(pom PomXML, parent Parent) string {
+	// prioritize the versionID of parent
+	if parent.VersionID != "" {
+		return parent.VersionID
+	}
+	return pom.VersionID
 }
 
 func (Rebuilder) CloneRepo(ctx context.Context, t rebuild.Target, repoURI string, fs billy.Filesystem, s storage.Storer) (r rebuild.RepoConfig, err error) {
--- a/pkg/rebuild/maven/infer.go
+++ b/pkg/rebuild/maven/infer.go
@@ -32,45 +32,13 @@ func (Rebuilder) InferRepo(ctx context.Context, t rebuild.Target, mux rebuild.Re
 	if err != nil {
 		return "", err
 	}
-	for pom.Repo() == "" && pom.Parent.ArtifactID != "" {
-		pom, err = resolveParentPom(ctx, pom, t, mux)
+	for pom.SCMURL == "" && pom.Parent.ArtifactID != "" {
+		pom, err = ResolveParentPom(ctx, pom, mux)
 		if err != nil {
 			return "", errors.Errorf("failed to resolve parent POM for %s: %v", t.Package, err)
 		}
 	}
 	return uri.CanonicalizeRepoURI(pom.Repo())
-}
-
-func resolveParentPom(ctx context.Context, pom PomXML, t rebuild.Target, mux rebuild.RegistryMux) (PomXML, error) {
-	if pom.Parent.ArtifactID == "" {
-		return pom, nil
-	}
-	parent, err := NewPomXML(ctx, rebuild.Target{
-		Ecosystem: t.Ecosystem,
-		Package:   fmt.Sprintf("%s:%s", getGroupID(pom, pom.Parent), pom.Parent.ArtifactID),
-		Version:   getVersionID(pom, pom.Parent),
-		Artifact:  t.Artifact,
-	}, mux)
-	if err != nil {
-		return PomXML{}, err
-	}
-	return parent, nil
-}
-
-func getGroupID(pom PomXML, parent Parent) string {
-	// prioritize the groupID of parent
-	if parent.GroupID != "" {
-		return parent.GroupID
-	}
-	return pom.GroupID
-}
-
-func getVersionID(pom PomXML, parent Parent) string {
-	// prioritize the versionID of parent
-	if parent.VersionID != "" {
-		return parent.VersionID
-	}
-	return pom.VersionID
 }
 
 func (Rebuilder) CloneRepo(ctx context.Context, t rebuild.Target, repoURI string, fs billy.Filesystem, s storage.Storer) (r rebuild.RepoConfig, err error) {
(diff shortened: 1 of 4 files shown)


algomaster99:
The data model is now forced to resolve for parent even when it is not needed. It would be better to have an API `.parent` which would resolve the parent pom to get the source URL.

algomaster99:
The changes at least eliminate (or there is a new cloning error) 16 of the inference failures due to source repo not found.

```
go_infer_logs/ch.qos.logback_logback-classic_1.2.3.log:
go_infer_logs/ch.qos.logback_logback-core_1.2.3.log:
go_infer_logs/com.fasterxml.jackson.datatype_jackson-datatype-jsr310_2.10.0.log:
go_infer_logs/com.google.code.gson_gson_2.10.1.log:
go_infer_logs/com.google.guava_failureaccess_1.0.1.log:
go_infer_logs/com.google.guava_guava_18.0.log:
go_infer_logs/com.sun.xml.bind_jaxb-core_2.3.0.log:
go_infer_logs/com.sun.xml.bind_jaxb-impl_2.3.0.log:
go_infer_logs/jakarta.xml.bind_jakarta.xml.bind-api_4.0.2.log:
go_infer_logs/javax.servlet_servlet-api_2.5.log:
go_infer_logs/javax.xml.bind_jaxb-api_2.3.0.log:
go_infer_logs/org.eclipse.collections_eclipse-collections_10.2.0.log:
go_infer_logs/org.eclipse.collections_eclipse-collections-api_10.2.0.log:
go_infer_logs/org.glassfish.main.javaee-api_javax.jws_3.1.2.2.log:
go_infer_logs/org.glassfish.web_el-impl_2.2.log:
go_infer_logs/org.quartz-scheduler_quartz_2.3.2.log:
```

pkg/rebuild/maven/infer.go
```diff
@@ -32,9 +32,47 @@ func (Rebuilder) InferRepo(ctx context.Context, t rebuild.Target, mux rebuild.Re
 	if err != nil {
 		return "", err
 	}
+	for pom.Repo() == "" && pom.Parent.ArtifactID != "" {
```
wbxyz: Should we remove the condition on pom.Repo()? Or maybe that should be something like pom.SCMURL == ""?

IIUC if the pom.URL != "" then the current implementation would prefer that over the parent's SCMURL.

pkg/rebuild/maven/infer.go
```diff
@@ -32,9 +32,47 @@ func (Rebuilder) InferRepo(ctx context.Context, t rebuild.Target, mux rebuild.Re
...
+		if err != nil {
+			return "", errors.Errorf("failed to resolve parent POM for %s: %v", t.Package, err)
+		}
+	}
 	return uri.CanonicalizeRepoURI(pom.Repo())
 }
 
+func resolveParentPom(ctx context.Context, pom PomXML, t rebuild.Target, mux rebuild.RegistryMux) (PomXML, error) {
+	if pom.Parent.ArtifactID == "" {
+		return pom, nil
```
wbxyz: Does this behavior lead to an infinite loop in some cases? It would make us think we moved to a new node but it's actually the same node. I think this should probably return an error, wdyt?

pkg/rebuild/maven/infer.go
```diff
@@ -32,9 +32,47 @@ func (Rebuilder) InferRepo(ctx context.Context, t rebuild.Target, mux rebuild.Re
...
+	for pom.Repo() == "" && pom.Parent.ArtifactID != "" {
+		pom, err = resolveParentPom(ctx, pom, t, mux)
+		if err != nil {
+			return "", errors.Errorf("failed to resolve parent POM for %s: %v", t.Package, err)
+		}
+	}
 	return uri.CanonicalizeRepoURI(pom.Repo())
 }
 
+func resolveParentPom(ctx context.Context, pom PomXML, t rebuild.Target, mux rebuild.RegistryMux) (PomXML, error) {
```
wbxyz: Let's move this function into pom.go, I think conceptually "get the parent" should be mechanical, rather than a heuristic which is my conceptual divide between pom.go and infer.go respectively.

Also just looking at the implementation, it might make sense for resolveParentPom to be implemented similar to NewPomXML

pkg/rebuild/maven/infer.go
```diff
@@ -32,9 +32,47 @@ func (Rebuilder) InferRepo(ctx context.Context, t rebuild.Target, mux rebuild.Re
...
+	return pom.GroupID
+}
+
+func getVersionID(pom PomXML, parent Parent) string {
+	// prioritize the versionID of parent
+	if parent.VersionID != "" {
+		return parent.VersionID
+	}
+	return pom.VersionID
+}
```
wbxyz: Are these needed outside resolveParentPom? They don't seem likely to be re-usable and inlining them might make that function easier to read.

pkg/rebuild/maven/infer.go
algomaster99: We were accidentally rescued. `resolveParentPom` is not called because under `InferRepo` we already do a check `pom.Parent.ArtifactID != ""`. However, I agree to keep the method self contained and have this check and also avoid accidental failures if we use this API anywhere else.

pkg/rebuild/maven/infer_test.go
```diff
@@ -158,3 +161,156 @@ func TestGetClassFileMajorVersion(t *testing.T) {
...
+						ArtifactID: "parent",
+						VersionID:  "1.0.0",
+					},
+				},
+				{
+					GroupID:    "com.example.parent",
+					ArtifactID: "parent",
+					VersionID:  "1.0.0",
+				},
+			},
```
algomaster99, line 281: It feels redundant code here because I have these structs and then I am writing them to XML in the mock Maven Registry so that they can be decoded into `PomXML` again by `NewPomXML`. But I believe that should be fixed when `NewPomXML` is added to `maven.Registry` interface. See todos in [`pom.go`](https://github.com/google/oss-rebuild/pull/701/files#diff-a0c6b68ec312e85677821d474acb4c16630c02dd28f79b824e5a77251f02ef9bR64-R65). Let me know if understood and added the todo correctly.

pkg/rebuild/maven/infer_test.go
```diff
@@ -17,15 +18,15 @@ import (
 
 type mockMavenRegistry struct {
 	maven.Registry
-	releaseFileContent io.ReadCloser
-	releaseFileError   error
+	artifactCoordinates map[struct{ PackageName, VersionID, FileType string }][]byte
```
algomaster99, line 21: The reason for `[]byte` here and not `string` (as we store pom XML content) is due to `TestJDKVersionInference` which requires jar bytes as release content.

pkg/rebuild/maven/infer_test.go
wbxyz: The TODO lgtm, we can leave it as a todo for now and revisit later if necessary.

pkg/rebuild/maven/infer_test.go
```diff
@@ -158,3 +161,156 @@ func TestGetClassFileMajorVersion(t *testing.T) {
...
+				addPomArtifact(mockRegistry, &pom)
+			}
+			mockMux := rebuild.RegistryMux{
+				Maven: mockRegistry,
+			}
+			targetPom := tc.pom[0]
+			got, err := Rebuilder{}.InferRepo(context.Background(), rebuild.Target{
+				Ecosystem: rebuild.Maven,
+				Package:   fmt.Sprintf("%s:%s", targetPom.GroupID, targetPom.ArtifactID),
+				Version:   targetPom.VersionID,
```
wbxyz: I think the rebuild.Target should be part of the test case, otherwise LGTM

pkg/rebuild/maven/infer_test.go
wbxyz: Yup that's fine.

pkg/rebuild/maven/pom.go
```diff
@@ -60,12 +60,38 @@ func (p *PomXML) Version() string {
 }
 
 // NewPomXML returns the POM file for a Maven package version.
-func NewPomXML(ctx context.Context, t rebuild.Target, mux rebuild.RegistryMux) (p PomXML, err error) {
-	var r io.ReadCloser
-	r, err = mux.Maven.ReleaseFile(ctx, t.Package, t.Version, maven.TypePOM)
+// TODO: NewPomXML should be added to the maven.Registry interface.
+// TODO: This will ensure that we can simply resolve parent POM by only changing the target, thus eliminating the need for `ResolveParentPom`.
```
algomaster99, line 64: @wbxyz I am having second thoughts about the todo. `pom.xml`, although being the build manifest, is also released on the Maven central thus qualifying it as a release artifact. There are also [multiple type of artifacts](https://github.com/google/oss-rebuild/pull/418#discussion_r2033936303) (plus anything you would like). It is better if all of them can be queried by ReleaseFile.

PR #703, opened 2025-08-19
Add the MixedAssetStore type

wbxyz:


pkg/rebuild/rebuild/storage.go
msuozzo, line 1: From the other PR: What about not (always?) supporting a fallback? Just a flat map that routes between asset stores.

msuozzo: Oh shoot you just commented about that at the top level.

wbxyz: See comment above (might not have been there when you reviewed): https://github.com/google/oss-rebuild/pull/703#issuecomment-3201987129

msuozzo:
> I also drafted a version that has no fallback, but I don't love the behavior of URL().

Ugh yeah that's as good a reason as any. Sgtm

msuozzo (approved):

wbxyz:
> > I also drafted a version that has no fallback, but I don't love the behavior of URL().
> 
> Ugh yeah that's as good a reason as any. I suppose panicking is an option. Basically the same behavior as a map. But keeping the default isn't the worst.

This is less of a concern after #704 so I'll do the no-default version. I think having a default was probably just a different kind of error-eating behavior. Long term we should maybe add an error arg on URL() but not important currently.

PR #705, opened 2025-08-19
Include error messages in http errors

wbxyz:
When we convert error objects into grpc status objects, then into http
codes, we lose the message. This means the receiver (usually a stub
caller) doesn't have any details of the failure besides the code.

For services like inference, this means a lot of rebuilds have a failure
explanatation (`message`) that looks something like:

```
getting strategy: fetching inference: 500 Internal Server Error: non-OK
response
```

When really, that failure message should include the error contents
generated by the inference service.

Note that we still preserve the correct status code in the response, so that information is not lost.

### DIFF
--- a/internal/api/rpc.go
+++ b/internal/api/rpc.go
@@ -53,7 +53,8 @@ func Stub[I Message, O any](client httpx.BasicClient, u *url.URL) StubT[I, O] {
 		}
 		defer resp.Body.Close()
 		if resp.StatusCode != http.StatusOK {
-			return nil, errors.Wrap(ErrNotOK, resp.Status)
+			b, _ := io.ReadAll(resp.Body)
+			return nil, errors.Wrap(errors.Wrap(ErrNotOK, resp.Status), string(b))
 		}
 		var o O
 		if err := json.NewDecoder(resp.Body).Decode(&o); err != nil {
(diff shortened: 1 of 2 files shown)


internal/api/rpc.go
```diff
@@ -53,7 +53,8 @@ func Stub[I Message, O any](client httpx.BasicClient, u *url.URL) StubT[I, O] {
 		}
 		defer resp.Body.Close()
 		if resp.StatusCode != http.StatusOK {
-			return nil, errors.Wrap(ErrNotOK, resp.Status)
+			b, _ := io.ReadAll(resp.Body)
+			return nil, errors.Wrap(errors.Wrap(ErrNotOK, resp.Status), string(b))
```
msuozzo, line 57: Can we add an XXX or FIXME here? We should be clear that this is not proper API behavior and this is expedience to address a quality-of-life concern. This leaves us at least some chance of revisiting this decision if/when we need to.

msuozzo (approved):

internal/api/rpc.go
wbxyz: What type of api behavior would be proper? Do you mean we should be using real grpc status objects that separate the status code from the message?

msuozzo: Yeah that's not a bad option. Mainly, I'd like to avoid this interface leaking potentially sensitive information through our normal error handling path. It's less acute if the user needs to explicitly opt to include this info (e.g. when the risk is low, as with inference)

wbxyz: Oh, yes definitely. That was something I thought about but without re-writing all the errors I don't have a good solution atm. One long-term option would be something similar to #623 but for error messages more broadly (including inference errors) where we can return the constant value and no dynamic details.

Regardless, it's not pressing because we don't have any public interfaces on our services.

The comment should probably go on the http writer suggesting that including the entire error message is potentially dangerous.

msuozzo (approved):
Would be nice to revisit this structure, regardless, so notes in both places would be nice.

PR #706, opened 2025-08-20
Reword errors to group JDK version and shorten them

algomaster99:
Does three things:
1. Groups failures due to absent JDK.
2. Groups failures due to cloning repositories.
3. Shortens the error prefix `getting strategy: fetching inference: failed to infer strategy:` to `inference failure:`. This helps with viewing error messages.

### DIFF
--- a/tools/ctl/rundex/rundex.go
+++ b/tools/ctl/rundex/rundex.go
@@ -132,8 +132,12 @@ func DoQuery[T any](ctx context.Context, q firestore.Query, fn func(*firestore.D
 }
 
 func cleanVerdict(m string) string {
+	// strip the leading error message of an inference failure to fit the latter, finer-grained error messages, on screen.
+	m = strings.Replace(m, "getting strategy: fetching inference: failed to infer strategy:", "inference failure:", -1)
 	switch {
 	// Generic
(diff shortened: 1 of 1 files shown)


tools/ctl/rundex/rundex.go
```diff
@@ -132,8 +132,11 @@ func DoQuery[T any](ctx context.Context, q firestore.Query, fn func(*firestore.D
 }
 
 func cleanVerdict(m string) string {
+	m = strings.Replace(m, "getting strategy: fetching inference: failed to infer strategy:", "inference failure:", -1)
```
msuozzo, line 136: can we add a short note that we expect this to be the inference error prefix for rebuilds so we strip it to ensure inference errors are properly sanitized?

msuozzo (approved):

tools/ctl/rundex/rundex.go
algomaster99: the purpose of this is to fit the more important error message on the screen. I don't think so this would fall under sanitation. I have given it a go so please take a look feel free to reword. :)

algomaster99:
wait what, let me fix this

PR #707, opened 2025-08-20
Store build version of oss rebuild in firestore

algomaster99:


wbxyz:
For context, this change moves the firestore version away from the cloud-run generated version number, and instead uses the golang version number. This is something like: `v0.0.0-20250507220843-d9ec242bc8ca`

Another option might be to combine K_REVISION and deps.ServiceRepo.Ref, giving us the best of both worlds.

PR #708, opened 2025-08-20
Remove rebuilds with OOM from benchmark

algomaster99:
It is a bit strange that some specific artifacts are returning out of memory and not the entire package with all versions.


algomaster99:
Will reevaluate this after running my script.

PR #709, opened 2025-08-20
Migrate TUI to new build API

wbxyz:


tools/ctl/ide/commands/commands.go
```diff
@@ -51,21 +52,66 @@ const (
...
 )
 
-func NewRebuildCmds(app *tview.Application, rb *rebuilder.Rebuilder, modalFn modal.Fn, butler localfiles.Butler, aiClient *genai.Client, buildDefs rebuild.LocatableAssetStore, dex rundex.Reader, benches benchmark.Repository, cmdReg *commandreg.Registry) []commandreg.RebuildCmd {
+func runLocalInput(ctx context.Context, executor *local.DockerRunExecutor, dex rundex.Reader, inp rebuild.Input) error {
+	assets := rebuild.NewFilesystemAssetStore(memfs.New())
+	h, err := executor.Start(ctx, inp, build.Options{
+		// Use a constant BuildID to make sure we overwrite the container each run
+		BuildID: containerName,
+		Resources: build.Resources{
+			AssetStore: assets,
```
algomaster99, line 71: The `DefaultBaseImageConfig` needs to be added here. Otherwise by default the alpine image is selected because of `DetectOS`.

Despite this, I am unable to run the docker file locally. It throws a format error which is hard to debug. I tried to rebuild `maven!ch.qos.logback:logback-classic!1.4.14!logback-classic-1.4.14.jar` but should be applicable to any Maven build.

tools/ctl/ide/commands/commands.go
```diff
@@ -51,21 +52,66 @@ const (
...
+}
+
+func runLocal(ctx context.Context, executor *local.DockerRunExecutor, dex rundex.Reader, target rebuild.Target) error {
+	cmd := exec.CommandContext(ctx, "go", "run", "tools/ctl/ctl.go", "infer", "--ecosystem", string(target.Ecosystem), "--package", target.Package, "--version", target.Version, "--artifact", target.Artifact)
+	cmd.Stderr = log.Default().Writer()
+	out, err := cmd.Output()
+	if err != nil {
+		return errors.Wrap(err, "running inference")
+	}
+	var stratOneOf schema.StrategyOneOf
```
msuozzo: nit: when we use "strat" as an abbreviation, the linter complains that we've misspelled "start". (I know it's dumb but I've decided to not rage against the machine on this point)

tools/ctl/ide/commands/commands.go
```diff
@@ -51,21 +52,66 @@ const (
...
+	} else if res.Error != nil {
+		return errors.Wrap(res.Error, "build failed")
+	}
+	// TODO: How should we the artifacts? Maybe butler?
+	// TODO: Log results to dex if it's writable
+	return nil
+}
+
+func runLocal(ctx context.Context, executor *local.DockerRunExecutor, dex rundex.Reader, target rebuild.Target) error {
+	cmd := exec.CommandContext(ctx, "go", "run", "tools/ctl/ctl.go", "infer", "--ecosystem", string(target.Ecosystem), "--package", target.Package, "--version", target.Version, "--artifact", target.Artifact)
```
msuozzo: I think this needs to be `./tools/ctl/ctl.go` (and might be nice to abbreviate to `./tools/ctl`).

msuozzo: for execution command, you could add an option to pull from upstream (i.e. github.com/google/oss-rebuild). you might even detect when the local case will fail and fall back to the upstream path (with log feedback, of course).

tools/ctl/ide/commands/commands.go
```diff
@@ -51,21 +52,66 @@ const (
...
+	cmd.Stderr = log.Default().Writer()
+	out, err := cmd.Output()
```
msuozzo, line 96: alternatively, you could construct the `Cmd` obj, get a stdout pipe, and pass that to the JSON decoder.

tools/ctl/ide/commands/commands.go
```diff
@@ -51,21 +52,66 @@ const (
...
-				rb.RunLocal(ctx, example, rebuilder.RunLocalOpts{})
-			},
-		},
-		{
-			Short: "restart && run local",
-			Func: func(ctx context.Context, example rundex.Rebuild) {
-				rb.Restart(ctx)
-				rb.RunLocal(ctx, example, rebuilder.RunLocalOpts{})
+				if err := runLocal(ctx, executor, dex, example.Target()); err != nil {
+					log.Println(err)
```
msuozzo, line 124: idle thought: might be worth changing the signature to support returning an error and having the printing handled centrally. Could introduce it gradually as FuncE comparably to the cli lib pattern

tools/ctl/ide/commands/commands.go
```diff
@@ -51,21 +52,66 @@ const (
 	RundexReadParallelism = 10
 	LLMRequestParallelism = 50
 	expertPrompt          = `You are an expert in diagnosing build issues in multiple open source ecosystems. You will help diagnose why builds failed, or why the builds might have produced an artifact that differs from the upstream open source package. Provide clear and concise explantions of why the rebuild failed, and suggest changes that could fix the rebuild`
+	containerName         = "oss-rebuild-debug-temp"
 )
 
-func NewRebuildCmds(app *tview.Application, rb *rebuilder.Rebuilder, modalFn modal.Fn, butler localfiles.Butler, aiClient *genai.Client, buildDefs rebuild.LocatableAssetStore, dex rundex.Reader, benches benchmark.Repository, cmdReg *commandreg.Registry) []commandreg.RebuildCmd {
+func runLocalInput(ctx context.Context, executor *local.DockerRunExecutor, dex rundex.Reader, inp rebuild.Input) error {
```
msuozzo: can we just make this runLocal and have the other be inferLocal?

if you think it necessary, you could then write inferAndRunLocal which composes the two.

msuozzo: do these need to take DockerRunExecutor instances? can they be Executors?

tools/ctl/ide/commands/commands.go
```diff
@@ -341,7 +395,7 @@ func NewRebuildCmds(app *tview.Application, rb *rebuilder.Rebuilder, modalFn mod
 	}
 }
 
-func NewRebuildGroupCmds(app *tview.Application, rb *rebuilder.Rebuilder, modalFn modal.Fn, butler localfiles.Butler, aiClient *genai.Client, buildDefs rebuild.LocatableAssetStore, dex rundex.Reader, benches benchmark.Repository, cmdReg *commandreg.Registry) []commandreg.RebuildGroupCmd {
+func NewRebuildGroupCmds(app *tview.Application, executor *local.DockerRunExecutor, modalFn modal.Fn, butler localfiles.Butler, aiClient *genai.Client, buildDefs rebuild.LocatableAssetStore, dex rundex.Reader, benches benchmark.Repository, cmdReg *commandreg.Registry) []commandreg.RebuildGroupCmd {
```
msuozzo: same here. possible to make this a generic Executor?

tools/ctl/ide/commands/commands.go
wbxyz: Added a DefaultBaseImageConfig, we can followup on the template failures in another thread.

tools/ctl/ide/commands/commands.go
wbxyz: I think for now I'll leave it as running the local version to keep this PR and logic minimal, but we can definitely revisit in a followup.

tools/ctl/ide/commands/commands.go
wbxyz: True, although I like the debugging capability from printing the strategy on failures.

tools/ctl/ide/commands/commands.go
wbxyz: Yeah I've thought about that too, but I'll leave it as a future item.

tools/ctl/ide/ui.go
```diff
@@ -48,31 +48,37 @@ func NewTuiApp(dex rundex.Reader, watcher rundex.Watcher, rundexOpts rundex.Fetc
 		log.Default().SetFlags(0)
 		logs.SetBorder(true).SetTitle("Logs")
 		logs.ScrollToEnd()
-		rb := &rebuilder.Rebuilder{}
+		executor, err := local.NewDockerRunExecutor(local.DockerRunExecutorConfig{
+			// Set this to true to make exploration possible. Requires state management.
+			RetainContainer: true,
```
msuozzo, line 57: can you give a bit more here? the strategy is to recycle the same container name each time and this setting allows us to do that.

PR #710, opened 2025-08-21
Fallback to JDK 11 in case of no URL found

algomaster99:
If we don't have a URL for particular JDK distribution, we fallback to 11.

### DIFF
--- a/pkg/rebuild/maven/infer.go
+++ b/pkg/rebuild/maven/infer.go
@@ -27,6 +27,8 @@ import (
 	"github.com/pkg/errors"
 )
 
+const fallbackJDK = "11"
+
 func (Rebuilder) InferRepo(ctx context.Context, t rebuild.Target, mux rebuild.RegistryMux) (string, error) {
 	pom, err := NewPomXML(ctx, t, mux)
 	if err != nil {
@@ -121,12 +123,9 @@ func (Rebuilder) InferStrategy(ctx context.Context, t rebuild.Target, mux rebuil
 		}
 		return cfg, errors.Errorf("no valid git ref")
 	}
-	jdk, err := getJarJDK(ctx, name, version, mux)
+	jdk, err := inferOrFallbackToDefaultJDK(ctx, name, version, mux)
 	if err != nil {
 		return cfg, errors.Wrap(err, "fetching JDK")
-	}
-	if jdk == "" {
-		return cfg, errors.New("no JDK found")
 	}
 	return &MavenBuild{
 		Location: rebuild.Location{
@@ -150,8 +149,24 @@ func getPomXML(tree *object.Tree, path string) (pomXML PomXML, err error) {
 	return pomXML, xml.Unmarshal([]byte(p), &pomXML)
 }
 
-// getJarJDK gets the JDK version that is used to compile the original artifact on registry.
-func getJarJDK(ctx context.Context, name, version string, mux rebuild.RegistryMux) (string, error) {
+// inferOrFallbackToDefaultJDK tries to infer the JDK version from the artifact's metadata, falling back to a default if necessary.
+func inferOrFallbackToDefaultJDK(ctx context.Context, name, version string, mux rebuild.RegistryMux) (string, error) {
+	jdk, err := inferJDKVersion(ctx, name, version, mux)
+	if err != nil {
+		return "", errors.Wrap(err, "fetching JDK")
+	}
+	if jdk == "" {
+		log.Printf("no JDK version inferred, falling back to JDK %s", fallbackJDK)
+		jdk = fallbackJDK
+	} else if JDKDownloadURLs[jdk] == "" {
+		log.Printf("%s has no associated JDK URL, falling back to JDK %s", jdk, fallbackJDK)
+		jdk = fallbackJDK
+	}
+	return jdk, nil
+}
+
+// inferJDKVersion gets the JDK version from the MANIFEST or Java bytecode.
+func inferJDKVersion(ctx context.Context, name, version string, mux rebuild.RegistryMux) (string, error) {
 	releaseFile, err := mux.Maven.ReleaseFile(ctx, name, version, maven.TypeJar)
 	if err != nil {
 		return "", errors.Wrap(err, "fetching jar file")
--- a/pkg/rebuild/maven/infer_test.go
+++ b/pkg/rebuild/maven/infer_test.go
@@ -83,6 +83,16 @@ func TestJDKVersionInference(t *testing.T) {
 			},
 			wantVersion: "11",
 		},
+		{
+			name: "fallback to default JDK version",
+			input: []*archive.ZipEntry{
+				{
+					FileHeader: &zip.FileHeader{Name: "META-INF/MANIFEST.MF"},
+					Body:       []byte("Manifest-Version: 1.0\r\nBuild-Jdk-Spec: 1.8.0_121\r\n\r\n"),
+				},
+			},
+			wantVersion: "11",
+		},
 	}
 
 	for _, tc := range testCases {
@@ -105,7 +115,7 @@ func TestJDKVersionInference(t *testing.T) {
 					},
 				},
 			}
-			got, err := getJarJDK(context.Background(), "dummy", "dummy", mockMux)
+			got, err := inferOrFallbackToDefaultJDK(context.Background(), "dummy", "dummy", mockMux)
 			if err != nil {
 				t.Fatalf("getJarJDK() error = %v", err)
 			}


pkg/rebuild/maven/strategy.go
```diff
@@ -23,7 +24,9 @@ var _ rebuild.Strategy = &MavenBuild{}
 func (b *MavenBuild) ToWorkflow() (*rebuild.WorkflowStrategy, error) {
 	jdkVersionURL, exists := JDKDownloadURLs[b.JDKVersion]
 	if !exists {
-		return nil, errors.Errorf("no download URL for JDK version %s", b.JDKVersion)
+		// TODO: instead of falling back to JDK 11, we should precisely detect the JDK version
+		log.Print("Falling back to JDK 11")
+		jdkVersionURL = JDKDownloadURLs["11"]
```
msuozzo: Hmm is this the right place for a default? I think we might be better off hard-coding this at inference time as a default so that it's easier to change the behavior over time.

algomaster99: Our inference does return a valid JDK version (based on the MANIFEST). It is just that we don't have the JDK URL. In the run, the map look up fails so we should ideally return the default there and this what the change is as of now.

msuozzo: Right but we own both sides of the system and, in other ecosystems, we apply these sorts of "practical" defaults at inference time. Let's move that logic earlier so the materialization remains simple.

wbxyz: I think we could do the default at the version-level rather than the URL level maybe? So adding a default case in here if jdk is empty or doesn't exist in JDKDownloadURLS?

wbxyz: Whoops this is the link https://github.com/google/oss-rebuild/blob/622d3b5d70e1e6874aa3dd30e6c2fbb00b141751/pkg/rebuild/maven/infer.go#L128

algomaster99: > we apply these sorts of "practical" defaults at inference time

okay, we can handle it at inference time, but we would need to store the URL in `MavenBuild` struct instead otherwise we do two lookups - one in infer and the other in strategy.

EDIT: I let the struct store version only not URL.

pkg/rebuild/maven/infer.go
```diff
@@ -125,8 +125,9 @@ func (Rebuilder) InferStrategy(ctx context.Context, t rebuild.Target, mux rebuil
 	if err != nil {
 		return cfg, errors.Wrap(err, "fetching JDK")
 	}
-	if jdk == "" {
-		return cfg, errors.New("no JDK found")
+	if jdk == "" || JDKDownloadURLs[jdk] == "" {
+		log.Print("falling back to JDK 11")
+		jdk = "11"
```
algomaster99: @msuozzo @wbxyz Don't you think this is too "hidden" for a default to be configured?

msuozzo: I might turn the log message into a comment. (if we were to keep the log message, I'd probably want to split the cases so you can give more specific feedback to differentiate the cases 1. the jdk was empty 2. the jdk had no associated url).

msuozzo (approved):

pkg/rebuild/maven/infer.go
algomaster99: >  you can give more specific feedback to differentiate the cases 1. the jdk was empty 2. the jdk had no associated url

I agree. Let me do that and then we can proceed to merge.

pkg/rebuild/maven/infer.go
```diff
@@ -125,8 +125,13 @@ func (Rebuilder) InferStrategy(ctx context.Context, t rebuild.Target, mux rebuil
 	if err != nil {
 		return cfg, errors.Wrap(err, "fetching JDK")
 	}
+	fallbackJDK := "11"
 	if jdk == "" {
-		return cfg, errors.New("no JDK found")
+		log.Print("no JDK version inferred, falling back to JDK 11")
+		jdk = fallbackJDK
+	} else if JDKDownloadURLs[jdk] == "" {
+		log.Printf("%s has no associated JDK URL, falling back to JDK 11", jdk)
```
wbxyz: Rather than hardcoding the fallback version in the error string, you could use the value of fallbackJDK in the Printf

pkg/rebuild/maven/strategy_test.go
```diff
@@ -62,6 +62,9 @@ wget -q -O - "https://download.java.net/java/GA/jdk11/13/GPL/openjdk-11.0.1_linu
 	}
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
+			if tc.name == "throw an error if JDK installation candidate is not found" {
+				t.Skip("Skipping this test until we have a better way to handle unknown JDK versions")
+			}
```
wbxyz: I think we should adjust the test case's expectation to match the new fallback behavior. Even if we plan to change the behavior in the future, this temporary behavior deserves to be covered. Updating the test case also makes it clear to readers what current (if temporary) the behavior is.

If you like, we can add comments on the test case explaining why the system does that behavior.

algomaster99: > this temporary behavior deserves to be covered.

We can add a test for this in `infer_test.go` because infer is where we set the default. In strategy, it still throws an error if an unknown JDK so this behaviour is preserved, hence no skip is needed.

wbxyz: Ah, right yes let's add a test for it in infer_test.go. Does the existing test in strategy_test not pass anymore?

algomaster99: I might have done too much of change to resolve this but I think it makes sense. So I am of the opinion we need to change `infer_test.go` to test the fallback behavior. However, the changes were in `InferStrategy` which does not have a test. For JDK version inference `getJarJDK` is tested and thus, I decided to include the logic of falling back there. Now, it is called `inferOrFallbackToDefaultJDK` which basically encapsulates inference and falling back to JDK 11.

algomaster99: > Does the existing test in strategy_test not pass anymore?

It does. The test case still throws the error because its input is an already inferred `MavenBuild` which has a value of `30`. We only changed the inference logic - by adding fallback to it.

pkg/rebuild/maven/infer.go
```diff
@@ -150,8 +147,25 @@ func getPomXML(tree *object.Tree, path string) (pomXML PomXML, err error) {
...
 
-// getJarJDK gets the JDK version that is used to compile the original artifact on registry.
-func getJarJDK(ctx context.Context, name, version string, mux rebuild.RegistryMux) (string, error) {
+// inferOrFallbackToDefaultJDK tries to infer the JDK version from the artifact's metadata, falling back to a default if necessary.
+func inferOrFallbackToDefaultJDK(ctx context.Context, name, version string, mux rebuild.RegistryMux) (string, error) {
+	jdk, err := inferJDKVersion(ctx, name, version, mux)
+	if err != nil {
+		return "", errors.Wrap(err, "fetching JDK")
+	}
+	fallbackJDK := "11"
```
wbxyz: nit: Let's make this a top-level constant

pkg/rebuild/maven/infer.go
```diff
@@ -150,8 +147,25 @@ func getPomXML(tree *object.Tree, path string) (pomXML PomXML, err error) {
...
+	if jdk == "" {
+		log.Printf("no JDK version inferred, falling back to JDK %s", fallbackJDK)
+		jdk = fallbackJDK
+	} else if JDKDownloadURLs[jdk] == "" {
+		log.Printf("%s has no associated JDK URL, falling back to JDK 11", jdk)
```
wbxyz: Similar to above, use the fallbackJDK in the printf?

wbxyz (approved):

pkg/rebuild/maven/strategy_test.go
wbxyz: Cool, changes look good to me besides those two nitpicks

pkg/rebuild/maven/infer.go
algomaster99: I recall changing it. Somehow, it did not make it through 😅

pkg/rebuild/maven/infer.go
algomaster99: Is there some convention you would like to follow? I see a mix of UpperCamelCase and lowerCamelCase in the codebase.

pkg/rebuild/maven/strategy_test.go
algomaster99: Resolved them too.

pkg/rebuild/maven/infer.go
wbxyz: Capitalizaion in golang determines whether a symbol is public or private, and in general the guidance is to keep things private until it's necessary to share them.

https://google.github.io/styleguide/go/decisions.html#constant-names

wbxyz (approved):

pkg/rebuild/maven/infer.go
algomaster99: > Capitalizaion in golang determines whether a symbol is public or private

right. I did not extend this knowledge to constants. thanks for reminder. :)

PR #712, opened 2025-08-22
Display oss-rebuild version in metadata

algomaster99:
I called it `OSSRebuildVersion` instead of `ExecutorVersion` just to be clearer.


msuozzo:
> I called it `OSSRebuildVersion` instead of `ExecutorVersion` just to be clearer.

Can we stick to executor? Or "system"? I don't really like recapitulation the project name in the data model.

algomaster99:
@msuozzo changed it to `ExecutorVersion`.

PR #714, opened 2025-08-25
Filter on pom.xml using base name and non test directory

algomaster99:
During inference, we first infer the the path to the pom.xml in the `HEAD` revision. We should ensure that this pom file's basename should exactly be `pom.xml` (and not suffix) and should not be a test resource. This hinder incorrect inference later when we do version heuristic search.

Note, that this only affects 5 artifacts in the current benchmark.

### DIFF
--- a/pkg/rebuild/maven/infer.go
+++ b/pkg/rebuild/maven/infer.go
@@ -283,7 +283,10 @@ func findPomXML(commit *object.Commit, pkg string) (*PomXML, string, error) {
 	var names []string
 	var pomXMLs []PomXML
 	commitTree.Files().ForEach(func(f *object.File) error {
-		if !strings.HasSuffix(f.Name, "pom.xml") {
+		// Per Maven conventions, skip non-"pom.xml" files and those inside a `src` directory (unlikely to contain metadata).
+		// Reference: https://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html
+		// Note: This will miss pom files, with name other than "pom.xml", whose paths are explicitly passed during build via "-f".
+		if path.Base(f.Name) != "pom.xml" || strings.HasPrefix(f.Name, "src/") || strings.Contains(f.Name, "/src/") {
 			return nil
 		}
 		pomXML, err := getPomXML(commitTree, f.Name)
--- a/pkg/rebuild/maven/infer_test.go
+++ b/pkg/rebuild/maven/infer_test.go
@@ -9,8 +9,14 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"path"
 	"testing"
+	"time"
 
+	"github.com/go-git/go-billy/v5/memfs"
+	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing/object"
+	"github.com/go-git/go-git/v5/storage/memory"
 	"github.com/google/oss-rebuild/pkg/archive"
 	"github.com/google/oss-rebuild/pkg/rebuild/rebuild"
 	"github.com/google/oss-rebuild/pkg/registry/maven"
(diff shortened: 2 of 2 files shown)


pkg/rebuild/maven/infer.go
```diff
@@ -283,7 +283,7 @@ func findPomXML(commit *object.Commit, pkg string) (*PomXML, string, error) {
 	var names []string
 	var pomXMLs []PomXML
 	commitTree.Files().ForEach(func(f *object.File) error {
-		if !strings.HasSuffix(f.Name, "pom.xml") {
+		if filepath.Base(f.Name) != "pom.xml" || strings.Contains(f.Name, "test/") {
```
msuozzo: Hmm I don't know about the test constraint. Can test appear in a genuine path? Is this the differentiator that unlocks the 4 entries in the benchmark?

msuozzo: I think we may want to use path instead of filepath (I think git metadata will always use forward slash?)

msuozzo (approved):

pkg/rebuild/maven/infer.go
algomaster99: maven-core 3.3.9, 3.6.3, 3.9.9
```
compat/maven-compat/src/test/resources/pom.xml
impl/maven-core/pom.xml
impl/maven-core/src/test/resources/pom.xml`
```
we want to select `impl/maven-core/pom.xml`

jedis 2.9.0 and 3.3.0

```
formatter-pom.xml
pom.xml
```

we want to select `pom.xml`

Correction: it is actually 5 and not 4. But really just two Maven releases.

> Can test appear in a genuine path?

Good point. We should probably filter on `src/test`.

algomaster99: > I think git metadata will always use forward slash?

I was not sure about this.[ `File` struct](https://github.com/go-git/go-git/blob/39261590f7f396873e83e7fc8ca8453f71225f10/plumbing/object/file.go#L15) did not specify this.

msuozzo: Uh maybe we could pick the shortest then? Either by character length or path component length? I might favor the latter

msuozzo: Yeah off the top of my head I'm not sure. You could dig into the implementation to see how it's set since the git object itself will only have the last level file path (the remainder of the path elements are constructed from the parent tree objects).

algomaster99: Or a better heuristic would be that `pom.xml` is at the root of the project and thus is always outside the `src` folder. 
I guess shortest path component length would work too but it seems less obvious.

algomaster99: It seems that[ `FileIter` simply joins the parent file and child with `/`](https://github.com/go-git/go-git/blob/39261590f7f396873e83e7fc8ca8453f71225f10/plumbing/object/tree.go#L553). Let's use `path` here.

On a related note, how about we configure our CI to also run on Windows? We would officially support Windows too. 
```
- runs-on: ubuntu-latest
+ runs-on: ${{matrix.os}}
+ os: ["ubuntu-latest", "windows-latest"]
```

msuozzo: It's an option. Maybe open an issue where we could discuss?

msuozzo: I guess I'm confused about the use of Contains and I'm not sure it's clear why this constraint is included. I guess I'm also unclear on what we expect about the structure(s) of these jar files which might be useful to mention here.

algomaster99: I have added a comment about the structure that we expect with some [related documentation](https://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html). Please have a look. I have also described the edge cases, but they are too rare in my opinion. I don't [see them in the reproducible-central dataset](https://github.com/search?q=repo%3Ajvm-repo-rebuild%2Freproducible-central+-f+path%3A*.buildspec&type=code&p=1).

pkg/rebuild/maven/infer.go
```diff
@@ -283,7 +283,13 @@ func findPomXML(commit *object.Commit, pkg string) (*PomXML, string, error) {
 	var names []string
 	var pomXMLs []PomXML
 	commitTree.Files().ForEach(func(f *object.File) error {
-		if !strings.HasSuffix(f.Name, "pom.xml") {
+		// The standard layout of a Maven project recommends to place "pom.xml" file at the root of the project.
```
msuozzo: What about this?

```
// Per Maven conventions, skip non-"pom.xml" files and those inside a `src` directory (unlikely to contain metadata).
// Reference: https://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html
// Note: This will miss pom files whose paths are explicitly passed during build via "-f".
if path.Base(f.Name) != "pom.xml" || (strings.HasPrefix(f.Name, "src/") || strings.Contains(f.Name, "/src/")) {
```

msuozzo (approved):

pkg/rebuild/maven/infer.go
algomaster99: > (strings.HasPrefix(f.Name, "src/") || strings.Contains(f.Name, "/src/"))

These two enforce precise `src` directory and not `foosrc/` so this is good.

> This will miss pom files whose paths are explicitly passed during build via "-f"

We will still catch them if they are called "pom.xml", but not if "test-pom.xml".

algomaster99: Done! Reworded it a bit:
```diff
- // Note: This will miss pom files whose paths are explicitly passed during build via "-f".
+ // Note: This will miss pom files, with name other than "pom.xml", whose paths are explicitly passed during build via "-f".
```

PR #720, opened 2025-08-26
Add tags to gcb executor

### LINKED ISSUE #719: Add tags on GCB builds
These are surfaced in the UI quite nicely for tracking specific builds you might be interested in

We should probably tag by ecosystem, package, version and maybe runid. It might also be nice to have a tag for automated rebuilds, but this could be the run-id tag

wbxyz:
Fixes #719

### DIFF
--- a/pkg/build/gcb/executor.go
+++ b/pkg/build/gcb/executor.go
@@ -6,7 +6,9 @@ package gcb
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"log"
+	"regexp"
 	"time"
 
 	"github.com/google/oss-rebuild/internal/bufiox"
@@ -19,6 +21,8 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 )
+
+var buildTagDisallowedChar = regexp.MustCompile(`[^\w.-]`)
 
 // Executor implements build.Executor for Google Cloud Build execution using a planner
 type Executor struct {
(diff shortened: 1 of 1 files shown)


pkg/build/gcb/executor.go
```diff
@@ -274,6 +275,11 @@ func (e *Executor) makeBuild(plan *Plan) *cloudbuild.Build {
 		Options:        buildOptions,
 		LogsBucket:     e.logsBucket,
 		ServiceAccount: e.serviceAccount,
+		Tags: []string{
+			fmt.Sprintf("ecosystem-%s", t.Ecosystem),
+			fmt.Sprintf("package-%s", t.Package),
```
msuozzo: Do we know if there are any troublesome characters here that we might want to sanitize?

msuozzo (approved):

pkg/build/gcb/executor.go
wbxyz: I didn't see anything in the docs: https://cloud.google.com/build/docs/build-config-file-schema#tags

wbxyz: Added some sanitization anyway

pkg/build/gcb/executor.go
```diff
@@ -259,8 +261,14 @@ func (e *Executor) doBuild(ctx context.Context, handle *gcbHandle, cloudBuild *c
 	return waitMetadata.Build, nil
 }
 
+func buildTag(desc, val string) string {
+	combined := fmt.Sprintf("%s-%s", desc, val)
+	combined = regexp.MustCompile(`[^\w.-]`).ReplaceAllString(combined, "")
```
msuozzo: nit: we could move the regex to a top-level constant and only ReplaceAll on the `val` to avoid re-assigning combined.

wbxyz: I didn't realize go doesn't cache regex patterns the way python does. Moved to the top of the file.

I decided to keep the ReplaceAll covering the description, just to be defensive. I did inline the Sprintf though to avoid re-assigning.

PR #724, opened 2025-08-26
Add an export command to upload data from a run to GCS

wbxyz:


tools/ctl/ctl.go
```diff
@@ -357,52 +352,140 @@ var getResults = &cobra.Command{
...
+			}
+			log.Printf("Querying results for [run=%v,pattern=%s]", req.Runs, req.Opts.Pattern)
+			rebuilds, err = fireDex.FetchRebuilds(ctx, req)
+			if err != nil {
+				log.Fatal(err)
+			}
+		}
+		log.Printf("Fetched %d rebuilds", len(rebuilds))
+		// Export all the run objects.
+		RundexReadParallelism := 50
```
msuozzo: Why is this local and capitalized?

tools/ctl/ctl.go
```diff
@@ -357,52 +352,140 @@ var getResults = &cobra.Command{
...
+				destDex = rundex.NewLocalClient(rundexFS)
+			default:
+				log.Fatal("destination must be a gs:// or file:// URL")
 			}
-		case "rundex":
-			dex := rundex.NewLocalClient(localfiles.Rundex())
-			ctx := cmd.Context()
-			runs, err := fireClient.FetchRuns(ctx, rundex.FetchRunsOpts{IDs: req.Runs})
+		}
+		aTypes := strings.Split(*assetTypes, ",")
```
msuozzo: Can this variable be assets or assetTypes instead? Had a bit of trouble parsing the use below with this abbreviation.

tools/ctl/ctl.go
```diff
@@ -357,52 +352,140 @@ var getResults = &cobra.Command{
...
+			if *exportRundex {
+				if err := destDex.WriteRebuild(ctx, in); err != nil {
+					res.errs = append(res.errs, errors.Wrap(err, "writing rebuild to rundex"))
 				}
-				cmd.OutOrStdout().Write([]byte(r.ID() + "\n"))
 			}
-		default:
-			log.Fatalf("Unknown --format type: %s", *format)
+			for _, at := range aTypes {
+				atype := rebuild.AssetType(at)
```
msuozzo: This type coercion feels like it should be done once centrally.

tools/ctl/ctl.go
```diff
@@ -357,52 +352,140 @@ var getResults = &cobra.Command{
...
+			}
+		})
+		bar := pb.New(len(rebuilds))
+		bar.Output = cmd.OutOrStderr()
+		bar.ShowTimeLeft = true
+		bar.Start()
+		for re := range p.Out() {
+			if len(re.errs) > 0 {
+				for _, err := range re.errs {
+					log.Printf("%s: %v", re.rebuild.ID(), err)
```
msuozzo: If we're just going to sink it to output, you could produce a stream of errors with the rebuild ID already encoded instead.

tools/ctl/ctl.go
wbxyz: Originally I used assetTypes but that collides with the flag var. If it was confusing though I can rename the flag to assetTypesFlag and keep this assetTypes.

tools/ctl/ctl.go
wbxyz: Copypasta, shouldn't be capitalized. Doesn't necessarily deserve to be top-level yet.

tools/ctl/ctl.go
wbxyz: I wanted to include the rebuild.ID() with each error line, and I also wanted to increment the progress bar exactly once for each processed output, both of which would have been hard with basic error objects.

An alternative would be to do all that inside the doFn, but idk if those processes are threadsafe so it seems slightly safer to do it this way though not hugely.

wbxyz: Oh wait yeah you suggested encoding the rebuildID in the error. Yeah I could do that, would need to add that encoding in each place we output the errors though (only 2 places, so wouldn't be a big deal).

wbxyz: I'll leave it for now but I'm open to tweaking it if needed.

PR #725, opened 2025-08-27
Fix object path parsing for scoped packages

msuozzo:


neuracr (approved):

pkg/analyzer/enqueue_test.go
```diff
@@ -0,0 +1,233 @@
...
+		objectName  string
+		want        *schema.TargetEvent
+		wantErr     bool
+		errContains string
+	}{
+		{
+			name:       "regular package - valid",
+			objectName: "npm/lodash/4.17.21/lodash-4.17.21.tgz/rebuild.intoto.jsonl",
+			want: &schema.TargetEvent{
+				Ecosystem: rebuild.Ecosystem("npm"),
```
wbxyz: nit: Use the package-defined constant `rebuild.NPM`?

pkg/analyzer/enqueue_test.go
```diff
@@ -0,0 +1,233 @@
...
+			},
+		},
+		{
+			name:        "regular package - too few segments",
+			objectName:  "npm/lodash/4.17.21/rebuild.intoto.jsonl",
+			wantErr:     true,
+			errContains: "unexpected object path length",
+		},
+		{
+			name:        "regular package - scoped package segments",
```
wbxyz: nit: Name should mention missing "@"

pkg/analyzer/enqueue.go
```diff
@@ -18,13 +17,21 @@ import (
...
-	// Expected form: ecosystem/package/version/artifact/rebuild.intoto.jsonl
-	// TODO: Use logic from AssetStore.
-	parts := strings.Split(filepath.Clean(event.Name), "/")
-	if len(parts) != 5 {
+	parts := strings.Split(event.Name, "/")
+	var ecosystem, pkg, version, artifact, obj string
+	if len(parts) == 6 {
+		// Format: ecosystem/@scope/package/version/artifact/rebuild.intoto.jsonl
+		scope := parts[1]
+		if !strings.HasPrefix(scope, "@") {
```
wbxyz: How do we want to handle debian packages, which have a similar pattern of "package" containing "/"? Those cases don't include "@" but they need a similar behavior of re-combining parts 1 and 2 into the `pkg` value.

PR #728, opened 2025-08-27
Fix output path for artifact in a submodule

algomaster99:
Initially, we would configure the output path as the directory `target/`, but if the artifact is in a submodule, the artifact would be inside the `target` directory which in turn is inside the submodule. For example, the artifact path of `maven!ch.qos.logback:logback-classic!1.5.18!logback-classic-1.5.18.jar ` is `logback-classic/target/logback-classic-1.5.18.jar` and not `target/logback-classic-1.5.18.jar`. We use `Dir` attribute in the MavenBuild struct. This `Dir` is the directory of the pom.xml that we infer and by default the target directory is at the same level as pom.xml.

### DIFF
--- a/pkg/rebuild/maven/strategy.go
+++ b/pkg/rebuild/maven/strategy.go
@@ -4,6 +4,8 @@
 package maven
 
 import (
+	"path"
+
 	"github.com/google/oss-rebuild/internal/textwrap"
 	"github.com/pkg/errors"
 
@@ -48,7 +50,7 @@ func (b *MavenBuild) ToWorkflow() (*rebuild.WorkflowStrategy, error) {
 				Needs: []string{"maven"},
 			},
 		},
-		OutputDir: "target/",
+		OutputDir: path.Join(b.Dir, "target"),
 	}, nil
 }
 
--- a/pkg/rebuild/maven/strategy_test.go
+++ b/pkg/rebuild/maven/strategy_test.go
@@ -42,7 +42,7 @@ wget -q -O - "https://download.java.net/java/GA/jdk11/13/GPL/openjdk-11.0.1_linu
 					export JAVA_HOME=/opt/jdk
 					export PATH=$JAVA_HOME/bin:$PATH
 					mvn clean package -DskipTests`[1:]),
-				OutputPath: "target/ldapchai-0.8.6.jar",
+				OutputPath: "dir/target/ldapchai-0.8.6.jar",
 			},
 			false,
 		},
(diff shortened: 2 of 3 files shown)


pkg/rebuild/maven/strategy.go
```diff
@@ -48,7 +50,7 @@ func (b *MavenBuild) ToWorkflow() (*rebuild.WorkflowStrategy, error) {
 				Needs: []string{"maven"},
 			},
 		},
-		OutputDir: "target/",
+		OutputDir: path.Join(b.Dir, "target"),
```
algomaster99, line 53: I have a naming concern here. `Dir` actually means where the rebuild instructions should be executed. However, here we are using it to find the where the artifact output is expected. The directory for executing rebuilds instruction is still the directory where the cloned repository is copied.

msuozzo: Yeah if I understand you, I believe that's working as intended.

msuozzo (approved):

pkg/rebuild/maven/strategy.go
algomaster99: Really? Based on docstrings, we should run the Maven build command at `Dir` but we don't (and we should not). We instead use `Dir` as to get directory of output artifact.

algomaster99: `crates`, `debian`, and `pypi` (`target/package`, `out`, and `dist`) seems to have fixed output paths and they seem to be using `Location.Dir` as the directory where the build command is run.

`npm` seems to be using it for both - running build cmd and output path.

msuozzo: Are you concerned about the naming? Location.Dir is meant to point to the place in the repo associated with that package (most of the time, where the package metadata definition is). The builds in different ecosystems may choose to use that data differently but I do think it's coherent.

algomaster99: > Are you concerned about the naming?

Yes.

> The builds in different ecosystems may choose to use that data differently

Should I reword:
```diff
- // Location is where a set of rebuild instruction should be executed.
+ // Location is where a set of rebuild instructions should be executed and rebuild output should be expected.
```

msuozzo: Ah that's not ideal. I'd remove the execution comment and present it as a higher-level construct e.g. "Location describes the source code location of a package" or something like that.

algomaster99: I revised it. Let me know what you think.

PR #732, opened 2025-08-28
Fix fetching of upstream artifact for diff

algomaster99:
Todos:
- [ ] Figure out why are there "success" builds for some cases even though fetching of upstream artifact of Maven was not in place.
- [ ] Even when press <kdb>d</kbd>, I don't see any diff in the tmux window.


msuozzo (approved):
I think this interface isn't used in the attestation-creating rebuild path (although it probably should be).

wbxyz:
To clarify a bit: the diff view in TUI is implemented separately from the artifact comparison done during attestation.

This is because attestation api uses a low-dependency content summary comparison (it's comparing the hash digest of the files among other things) whereas the TUI is using diffoscope for a rich human-readable diff view. The attestation flow I think is already working fine, but the locally-executed diffoscope version is that one that was broken and you're now fixing. That's  why "success" results were having errors when attempting to view the diff locally.

I suspect the unresponsive diff command behavior is a separate bug, but we can investigate that further.

algomaster99:
>  suspect the unresponsive diff command behavior is a separate bug, but we can investigate that further.

Yep, I am running diffoscope locally and investigating it.

pkg/rebuild/rebuild/compare.go
```diff
@@ -29,6 +30,8 @@ func UpstreamArtifactReader(ctx context.Context, t Target, mux RegistryMux) (io.
 			return nil, errors.Errorf("failed to parse debian component: %s", t.Package)
 		}
 		return mux.Debian.Artifact(ctx, component, name, t.Artifact)
+	case Maven:
+		return mux.Maven.ReleaseFile(ctx, t.Package, t.Version, maven.TypeJar)
```
algomaster99: @msuozzo @wbxyz do you think we should have another API `Artifact` so that we can make this generic and not just for `.jar` files? We could also extract artifact type from target here if we don't want to create another API.

wbxyz: Yeah we should probably have an Artifact interface that accepts a Target. For now it can just check the artifact suffix for .jar, then call ReleaseFile(... maven.TypeJar) or fail.

msuozzo:
We should definitely do a change like this in a context where we expect it not to change behavior. Let's move the bug fix to a follow on pr.

algomaster99:
>>  suspect the unresponsive diff command behavior is a separate bug, but we can investigate that further.

> Yep, I am running diffoscope locally and investigating it.

I briefly investigated it the same day as the comment. I remember it took around 20 minutes to generate the diffoscope file but then displaying that output crashed. When I have more details than this, I will create a ticket or propose a PR if possible.

pkg/registry/maven/maven.go
```diff
@@ -173,3 +174,22 @@ func (r HTTPRegistry) ReleaseURL(ctx context.Context, pkg, version, typ string)
...
+
+// Artifact returns file that is part of the Maven release.
+func (r HTTPRegistry) Artifact(ctx context.Context, pkg, version, artifact string) (io.ReadCloser, error) {
+	g, a, found := strings.Cut(pkg, ":")
+	if !found {
+		return nil, errors.New("package identifier not of form 'group:artifact'")
+	}
+	artifactPath := path.Join(strings.ReplaceAll(g, ".", "/"), a, version, artifact)
+	artifactURL := releaseURL.JoinPath(artifactPath)
+	req, _ := http.NewRequestWithContext(ctx, http.MethodGet, artifactURL.String(), nil)
```
msuozzo: hmm we might want to check this error, just to be sure. it's unlikely to be a concern but there are some characters in `g` that could cause the URL to be invalid (however unlikely that might be).

pkg/registry/maven/maven.go
```diff
@@ -173,3 +174,22 @@ func (r HTTPRegistry) ReleaseURL(ctx context.Context, pkg, version, typ string)
...
+	resp, err := r.Client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	if resp.StatusCode != 200 {
```
msuozzo: i think we've been using the StatusOK constants for the http repsonses.

pkg/rebuild/rebuild/compare.go
algomaster99: Resolved it here https://github.com/google/oss-rebuild/pull/735.

PR #733, opened 2025-08-28
Add meta.AllRebuilders

wbxyz:
It's common that we want to switch using target and get access to the
ecosystem-specific behavior. This package makes that relatively
straightforward, and makes adding new ecosystems easier.


pkg/rebuild/rebuilders/rebuilders.go
```diff
@@ -0,0 +1,21 @@
+// Copyright 2025 Google LLC
+// SPDX-License-Identifier: Apache-2.0
+
+package rebuilders
```
msuozzo: let's find a singular name to use for this package.

msuozzo (approved):

pkg/rebuild/rebuilders/rebuilders.go
wbxyz: Changed to meta.AllRebuilders

PR #734, opened 2025-08-28
Change ctl infer shell-script to use local/run_planner logic

wbxyz:


wbxyz:
Actually, I should probably change all the inference modes to create their respective planner objects, and just extract these formats from the plans

tools/ctl/ctl.go
```diff
@@ -925,6 +929,47 @@ var infer = &cobra.Command{
...
+		var buildScript string
+		{
+			plan, err := local.NewDockerRunPlanner().GeneratePlan(cmd.Context(), inp, build.PlanOptions{
+				UseTimewarp: rebuilders.All[inp.Target.Ecosystem].UsesTimewarp(inp),
+				Resources:   resources,
+			})
+			if err != nil {
+				log.Fatal(errors.Wrap(err, "generating plan"))
+			}
+			buildScript = strings.Join(plan.Command, " ")
```
wbxyz: Hmm actually this isn't right, the command is calling bash -c and the last arg needs to be escaped if it's going in a script. I should probably go back to makeShellScript.

tools/ctl/ctl.go
```diff
@@ -929,6 +889,47 @@ var infer = &cobra.Command{
...
+		inp := rebuild.Input{Target: rebuild.Target{
+			Ecosystem: rebuild.Ecosystem(*ecosystem),
+			Package:   *pkg,
+			Version:   *version,
+			Artifact:  *artifact,
+		}, Strategy: s}
+		resources := build.Resources{
+			ToolURLs: map[build.ToolType]string{
+				// TODO: Find a way to keep this up-to-date
+				build.TimewarpTool: "https://storage.googleapis.com/google-rebuild-bootstrap-tools/v0.0.0-20250428204534-b35098b3c7b7/timewarp",
```
msuozzo: really not a fan of hard-coding this. maybe flag(s) for the bucket and version?

tools/ctl/ctl.go
wbxyz: I changed the docker run planner to produce just the script, that can then be re-used here.

tools/ctl/ctl.go
wbxyz: I was being lazy, added a flag for bucket and version.

PR #736, opened 2025-08-29
Add inference for `GAV` given a Gradle project

algomaster99:
I have written a small Gradle script `printCoordinates.gradle` which works if you just go to a Gradle project root and run:
```
./gradlew printCoordinates --init-script printCoordinates.gradle
```
It outputs a file that contains all the GAVs of all submodules and the path of build.gradle.

However, I am struggling to get it execute on `git.Repository`. It complains that `Failed to run printCoordinates: failed to run Gradle command: exit status 127` which is strange. I know that `gradlew` exists with correct permissions and on the correct path `wt.Filesystem.Root()` (I did a programmatic `ls`).

PR #737, opened 2025-08-30
Add `get-gradle-gav` command to infer the GAVs in Gradle project

algomaster99:
Continuation of https://github.com/google/oss-rebuild/pull/736

Run: `go run tools/ctl/ctl.go get-gradle-gav --repository https://github.com/chains-project/maven-module-graph --ref 32245d0a433e1a36d36c9fffa16b5936243e9c6b`

```
{
  "groupId": "io.algomaster99.maven_module_graph",
  "artifactId": "maven-module-graph",
  "version": "1.0.0-SNAPSHOT",
  "buildManifest": "build.gradle"
}
```

### DIFF
--- a/tools/ctl/gradle/gradle.go
+++ b/tools/ctl/gradle/gradle.go
@@ -1,0 +1,85 @@
+// Copyright 2025 Google LLC
+// SPDX-License-Identifier: Apache-2.0
+
+package gradle
+
+import (
+	"encoding/json"
+	"os"
+	"os/exec"
+	"path"
+
+	"github.com/go-git/go-git/v5"
+	"github.com/pkg/errors"
+)
+
+const gradleScript = `import groovy.json.JsonOutput
+
+gradle.projectsLoaded {
+    def outputJsonPath = gradle.startParameter.projectProperties['outputJson']
+
+    gradle.rootProject.tasks.register("printCoordinates") {
+        doLast {
+            def jsonFile = new File(outputJsonPath ?: "output.json")
+
+            def rootProjectPath = gradle.rootProject.getProjectDir().toPath()
+
+            def toJsonMap
+            toJsonMap = { Project project ->
+                def map = [
+                        groupId   : project.getGroup().toString(),
+                        artifactId: project.getName(),
+                        version   : project.getVersion().toString(),
+                        buildManifest: rootProjectPath.relativize(project.getBuildFile().toPath()).toString(),
+                ]
+                if (!project.getChildProjects().isEmpty()) {
+                    map.submodules = project.getChildProjects().values().collect {
+                        toJsonMap(it)
+                    }
+                }
+                return map
+            }
+
+            def jsonModel = toJsonMap(gradle.getRootProject())
+            def jsonString = JsonOutput.prettyPrint(JsonOutput.toJson(jsonModel))
+            jsonFile.text = jsonString
+        }
+    }
+}`
+
+type GradleProject struct {
+	GroupId       string          `json:"groupId"`
+	ArtifactId    string          `json:"artifactId"`
+	Version       string          `json:"version"`
+	BuildManifest string          `json:"buildManifest"`
+	Submodules    []GradleProject `json:"submodules,omitempty"`
+}
+
+func RunPrintCoordinates(sourceRepo git.Repository) (GradleProject, error) {
+	cmd := exec.Command("./gradlew", "printCoordinates", "--init-script", "printCoordinates.gradle")
+	wt, err := sourceRepo.Worktree()
+	if err != nil {
+		return GradleProject{}, errors.Wrap(err, "getting worktree of Gradle project")
+	}
+	rootDir := wt.Filesystem.Root()
+	cmd.Dir = rootDir
+	if err := os.WriteFile(path.Join(rootDir, "printCoordinates.gradle"), []byte(gradleScript), 0644); err != nil {
+		return GradleProject{}, errors.Wrap(err, "failed to write print_coordinates.gradle file")
+	}
+	err = cmd.Run()
+	if err != nil {
+		return GradleProject{}, errors.Wrap(err, "failed to run Gradle command")
+	}
+	pathToGeneratedJson := path.Join(wt.Filesystem.Root(), "output.json")
+	f, err := os.Open(pathToGeneratedJson)
+	if err != nil {
+		return GradleProject{}, errors.Wrap(err, "failed to open generated JSON file")
+	}
+	defer f.Close()
+	var gradleProject GradleProject
+	err = json.NewDecoder(f).Decode(&gradleProject)
+	if err != nil {
+		return GradleProject{}, errors.Wrap(err, "failed to decode generated JSON file")
+	}
+	return gradleProject, nil
+}
--- a/tools/ctl/gradle/gradle_test.go
+++ b/tools/ctl/gradle/gradle_test.go
@@ -1,0 +1,93 @@
+// Copyright 2025 Google LLC
+// SPDX-License-Identifier: Apache-2.0
+
+package gradle
+
+import (
+	"fmt"
+	"os"
+	"reflect"
+	"slices"
+	"testing"
+
+	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing"
+)
+
+func TestRunPrintCoordinates(t *testing.T) {
+	testCases := []struct {
+		name         string
+		repoURL      string
+		sha          string
+		expectedGAVs []string
+	}{
+		{
+			name:         "GAV coordinates for a single module Gradle project",
+			repoURL:      "https://github.com/chains-project/maven-module-graph.git",
+			sha:          "32245d0a433e1a36d36c9fffa16b5936243e9c6b",
+			expectedGAVs: []string{"io.algomaster99.maven_module_graph:maven-module-graph:1.0.0-SNAPSHOT"},
+		},
+		{
+			name:    "GAV coordinates for a multi-module Gradle project",
+			repoURL: "https://github.com/perfmark/perfmark",
+			sha:     "9d9893c037949ec73ed9d018f6b9217b70d4bba6",
+			expectedGAVs: []string{
+				// "unspecified" version seems to be default value where no version is specified
+				":perfmark:unspecified",
+				"io.perfmark:perfmark-tracewriter:0.27.0",
+				"io.perfmark:perfmark-java7:0.27.0",
+				"io.perfmark:perfmark-api:0.27.0",
+				"io.perfmark:perfmark-java6:0.27.0",
+				"io.perfmark:perfmark-traceviewer:0.27.0",
+				"io.perfmark:perfmark-java15:0.27.0",
+				"io.perfmark:perfmark-java19:0.27.0",
+				"io.perfmark:perfmark-api-testing:0.27.0",
+				"io.perfmark:perfmark-impl:0.27.0",
+				"io.perfmark:perfmark-java9:0.27.0",
+				"io.perfmark:perfmark-agent:0.27.0",
+				"io.perfmark:perfmark-examples:0.27.0",
+				"io.perfmark:perfmark-testing:0.27.0"},
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			tempDir, err := os.MkdirTemp("", "gradle-test-")
+			if err != nil {
+				t.Fatalf("Failed to create temp directory: %v", err)
+			}
+			defer os.RemoveAll(tempDir)
+
+			r, err := git.PlainClone(tempDir, false, &git.CloneOptions{URL: tc.repoURL})
+			if err != nil {
+				t.Fatalf("Failed to clone repository: %v", err)
+			}
+			wt, err := r.Worktree()
+			if err != nil {
+				t.Fatalf("Failed to get worktree: %v", err)
+			}
+			err = wt.Checkout(&git.CheckoutOptions{Hash: plumbing.NewHash(tc.sha)})
+			if err != nil {
+				t.Fatalf("Failed to clone repository: %v", err)
+			}
+			gp, err := RunPrintCoordinates(*r)
+			if err != nil {
+				t.Fatalf("Failed to run printCoordinates: %v", err)
+			}
+			actualGAVs := getAllGAVs(gp)
+			slices.Sort(actualGAVs)
+			slices.Sort(tc.expectedGAVs)
+			if !reflect.DeepEqual(actualGAVs, tc.expectedGAVs) {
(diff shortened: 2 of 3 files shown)


tools/ctl/gradle/gradle_test.go
msuozzo, line 1: So this test accesses GitHub and starts up a JVM. Compared to the rest of the test suite, that would represent a pretty substantial outlier in terms of runtime, not to mention adding a heavyweight dependency to the CI environment. We could skip the test with t.Skip if there's no Gradle available but I might lean towards just leaving this test out. Yes, it's not ideal but we'd avoid those complexities for now.

algomaster99: We could reduce the complexity by mocking a git repository like we have done in other parts of the code. But starting up a JVM would be there. However, we run the [gradle wrapper](https://docs.gradle.org/current/userguide/gradle_wrapper.html) which is present in the souce code we clone. This is ideal for CI environment because we don't need to specify Gradle harness explicitly. The Gradle wrapper script takes care of that.

I also agree it adds complexity in terms of maintenance but maybe you are convinced by replacing cloning with a mock git repository and the usage of Gradle wrapper? I am inclined to keep the test just because the codebase is pretty big which tends to break random test mostly because we use a mixture of mock and real objects.

tools/ctl/gradle/gradle.go
```diff
@@ -0,0 +1,87 @@
...
+	ArtifactId    string          `json:"artifactId"`
+	Version       string          `json:"version"`
+	BuildManifest string          `json:"buildManifest"`
+	Submodules    []GradleProject `json:"submodules,omitempty"`
+}
+
+// RunPrintCoordinates runs the gradle script which print all possible GAV coordinates of a Gradle project and its submodules.
+// We run the script without a daemon as this is a one-off task and we do not want to keep a background process around.
+func RunPrintCoordinates(sourceRepo git.Repository) (GradleProject, error) {
+	cmd := exec.Command("./gradlew", "printCoordinates", "--init-script", "printCoordinates.gradle", "--no-daemon")
```
msuozzo: while it's not really in a good position to be used generically, we might want to abstract this with ./pkg/build/local.CommandExecutor so we can e.g. test it properly with the rest of the code.

tools/ctl/gradle/gradle.go
```diff
@@ -0,0 +1,87 @@
...
+	wt, err := sourceRepo.Worktree()
+	if err != nil {
+		return GradleProject{}, errors.Wrap(err, "getting worktree of Gradle project")
+	}
+	rootDir := wt.Filesystem.Root()
+	cmd.Dir = rootDir
+	if err := os.WriteFile(path.Join(rootDir, "printCoordinates.gradle"), []byte(gradleScript), 0644); err != nil {
```
msuozzo: these should use a billy.Filesystem provided as input to do the file i/o operations

tools/ctl/gradle/gradle_test.go
msuozzo: the changes i've suggested above should mean you'd be able to at least cover all the lines of code in the new behavior, although it wouldn't actually exercise the gradle logic.

algomaster99: >  although it wouldn't actually exercise the gradle logic.

I agree. We can leave that.

pkg/build/local/command.go
```diff
@@ -15,6 +15,8 @@ type CommandOptions struct {
 	Input io.Reader
 	// Output streams stdout/stderr to the writer (if nil, output is discarded)
 	Output io.Writer
+	// Directory in which to run the command
+	Directory string
```
msuozzo: nit: I would keep it to the abbreviated dir

pkg/build/local/command.go
```diff
@@ -45,6 +47,9 @@ func (r *realCommandExecutor) Execute(ctx context.Context, opts CommandOptions,
 		cmd.Stdout = opts.Output
 		cmd.Stderr = opts.Output
 	}
+	if opts.Directory != "" {
```
msuozzo: I believe you should be able to simplify this to unconditionally set this property

tools/ctl/gradle/gradle.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+	gradleScriptFile, err := wt.Filesystem.Create("printCoordinates.gradle")
+	if err != nil {
+		return GradleProject{}, errors.Wrap(err, "failed to create printCoordinates.gradle")
+	}
+	defer gradleScriptFile.Close()
+	if _, err := gradleScriptFile.Write([]byte(gradleScript)); err != nil {
+		return GradleProject{}, errors.Wrap(err, "failed to write printCoordinates.gradle file")
+	}
+	err = local.NewRealCommandExecutor().Execute(ctx, local.CommandOptions{Directory: wt.Filesystem.Root()}, "./gradlew", "printCoordinates", "--init-script", "printCoordinates.gradle", "--no-daemon")
```
msuozzo: In my original suggestion, I had meant for the command executor to be added as a parameter to this function. That way you can configure it to bypass actual execution during testing

algomaster99: The test verify if the gradle script is working. If we mock the execution, we are not testing anything. I thought you said so to reuse the executor harness.

algomaster99:
@msuozzo also notice that I have added `Ouptu: os.Output` to command line options to ensure that failure from Gradle is displayed and not discared.

tools/ctl/gradle/gradle.go
msuozzo: Testing is layered. Validating certain properties of a given piece of software can be done at many levels of complexity with commensurate costs, especially in added test code and new environment dependencies.

The test you've added here has a hard dependency on Gradle being present in the testing environment. This means adding it to all development environments, all CI ones, as well as paying the cost of its execution on each test run. In exchange for this cost, we get a high fidelity test that validates many of the properties of the code under test.

Assessing the trade-off of the costs and the benefits of such a test, I would say that given we are not reliant on this logic for production, we stand to gain less but bear the same costs. While not a test, the command you also add here is a manual way of validating the same behavioral properties without imposing any of the costs to the broader development process.

algomaster99: > has a hard dependency on Gradle being present in the testing environment

It does not have a dependency. We invoke the Gradle process using the [gradle wrapper] (https://docs.gradle.org/current/userguide/gradle_wrapper.html) (`./gradlew`) which downloads the toolchain if needed. One of the use cases of this wrapper is to not worry about the harness to setup Gradle toolchain. The tests also pass without any change in CI configuration.

> I would say that given we are not reliant on this logic for production, we stand to gain less but bear the same costs. While not a test, the command you also add here is a manual way of validating the same behavioral properties without imposing any of the costs to the broader development process.

I agree with this too. But then instead of mocking the output of gradle build, it is better to not have tests at all like you said before.

algomaster99:
@msuozzo I mocked clone and running Gradle script in the new tests. We are not _really_ testing that the output of the gradle script can be decoded. Drop your comments.

tools/ctl/gradle/gradle_test.go
```diff
@@ -0,0 +1,96 @@
...
+				return nil
+			})
+			gp, err := RunPrintCoordinates(context.Background(), *r, executor)
+			if err != nil {
+				t.Fatalf("Failed to run printCoordinates: %v", err)
+			}
+			actualGAVs := getAllGAVs(gp)
+			slices.Sort(actualGAVs)
+			slices.Sort(tc.expectedGAVs)
+			if !reflect.DeepEqual(actualGAVs, tc.expectedGAVs) {
```
msuozzo: Use cmp.Diff

pkg/build/local/command.go
```diff
@@ -15,6 +15,8 @@ type CommandOptions struct {
 	Input io.Reader
 	// Output streams stdout/stderr to the writer (if nil, output is discarded)
 	Output io.Writer
+	// Directory in which to run the command
```
msuozzo: nit: The first word needs to be Dir

tools/ctl/gradle/gradle.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+	gradleScriptFile, err := wt.Filesystem.Create("printCoordinates.gradle")
+	if err != nil {
+		return GradleProject{}, errors.Wrap(err, "failed to create printCoordinates.gradle")
+	}
+	defer gradleScriptFile.Close()
+	if _, err := gradleScriptFile.Write([]byte(gradleScript)); err != nil {
+		return GradleProject{}, errors.Wrap(err, "failed to write printCoordinates.gradle file")
+	}
+	err = gradleCmdExecutor.Execute(ctx, local.CommandOptions{Dir: wt.Filesystem.Root(), Output: os.Stdout}, "./gradlew", "printCoordinates", "--init-script", "printCoordinates.gradle", "--no-daemon")
```
msuozzo, line 74: nit: Either add a TODO or in this PR -- add CommandOptions as a param to this function since the command should actually want this to be `cmd.OutOrStdout`.

tools/ctl/gradle/gradle.go
```diff
@@ -0,0 +1,88 @@
...
+func RunPrintCoordinates(ctx context.Context, sourceRepo git.Repository, gradleCmdExecutor local.CommandExecutor) (GradleProject, error) {
+	wt, err := sourceRepo.Worktree()
+	if err != nil {
+		return GradleProject{}, errors.Wrap(err, "getting worktree of Gradle project")
+	}
+	gradleScriptFile, err := wt.Filesystem.Create("printCoordinates.gradle")
+	if err != nil {
+		return GradleProject{}, errors.Wrap(err, "failed to create printCoordinates.gradle")
+	}
+	defer gradleScriptFile.Close()
```
msuozzo: nit: you'll actually want to close this _before_ the command execution to be sure the write is flushed. as written, the file won't be closed until after the function returns.

tools/ctl/gradle/gradle.go
```diff
@@ -0,0 +1,88 @@
...
+	GroupId       string          `json:"groupId"`
+	ArtifactId    string          `json:"artifactId"`
+	Version       string          `json:"version"`
+	BuildManifest string          `json:"buildManifest"`
+	Submodules    []GradleProject `json:"submodules,omitempty"`
+}
+
+// RunPrintCoordinates runs the gradle script which print all possible GAV coordinates of a Gradle project and its submodules.
+// We run the script without a daemon as this is a one-off task and we do not want to keep a background process around.
+func RunPrintCoordinates(ctx context.Context, sourceRepo git.Repository, gradleCmdExecutor local.CommandExecutor) (GradleProject, error) {
```
msuozzo: nit: you could also make this return value a pointer which would mean you could `return nil, err` instead of needing to restate the return type.

tools/ctl/gradle/gradle.go
```diff
@@ -0,0 +1,88 @@
...
+        }
+    }
+}`
+
+type GradleProject struct {
+	GroupId       string          `json:"groupId"`
+	ArtifactId    string          `json:"artifactId"`
+	Version       string          `json:"version"`
+	BuildManifest string          `json:"buildManifest"`
+	Submodules    []GradleProject `json:"submodules,omitempty"`
```
msuozzo, line 55: Out of curiosity, can sub-modules be nested more one layer? If not, you could split this out into two types where only the top level could recurse i.e. GradleProject { InnerProject; Submodules []InnerProject }; InnerProject { Group...}

tools/ctl/gradle/gradle.go
```diff
@@ -0,0 +1,88 @@
...
+
+            def jsonModel = toJsonMap(gradle.getRootProject())
+            def jsonString = JsonOutput.prettyPrint(JsonOutput.toJson(jsonModel))
+            jsonFile.text = jsonString
+        }
+    }
+}`
+
+type GradleProject struct {
+	GroupId       string          `json:"groupId"`
```
msuozzo: nit: You could abbreviate this to "Group" and "Artifact" since the ID is somewhat superfluous (and linters may complain if they're not capitalized e.g. ID)

tools/ctl/gradle/gradle.go
algomaster99: Yes, they can be nested. Checkout: https://github.com/Waffle/waffle/tree/waffle-3.5.0

The graph of this is as follows. Each indentation is another nesting. This is the output of https://github.com/chains-project/maven-module-graph, but our script is similar.
```
com.github.waffle:waffle-launcher:3.4.1-SNAPSHOT
  com.github.waffle:waffle:3.5.0
    com.github.waffle:waffle-tomcat11:3.5.0
    com.github.waffle:waffle-tomcat10:3.5.0
    com.github.waffle:waffle-tomcat9:3.5.0
    com.github.waffle:waffle-tests-jakarta:3.5.0
    com.github.waffle:waffle-tests:3.5.0
    com.github.waffle:waffle-spring-security6:3.5.0
    com.github.waffle:waffle-spring-security5:3.5.0
    com.github.waffle:waffle-spring-boot3:3.5.0
      com.github.waffle:waffle-spring-boot-starter3:3.5.0
      com.github.waffle:waffle-spring-boot-autoconfigure3:3.5.0
    com.github.waffle:waffle-spring-boot2:3.5.0
      com.github.waffle:waffle-spring-boot-starter2:3.5.0
      com.github.waffle:waffle-spring-boot-autoconfigure2:3.5.0
    com.github.waffle:waffle-shiro:3.5.0
    com.github.waffle:waffle-jna-jakarta:3.5.0
    com.github.waffle:waffle-jna:3.5.0
    com.github.waffle:waffle-jetty-jakarta:3.5.0
    com.github.waffle:waffle-jetty:3.5.0
    com.github.waffle:waffle-distro:3.5.0
    com.github.waffle.demo:waffle-demo-parent:3.5.0
      com.github.waffle.demo:waffle-spring-boot-filter3:3.5.0
      com.github.waffle.demo:waffle-spring-boot-filter2:3.5.0
      com.github.waffle.demo:waffle-spring-form:3.5.0
      com.github.waffle.demo:waffle-spring-filter:3.5.0
      com.github.waffle.demo:waffle-negotiate:3.5.0
      com.github.waffle.demo:waffle-mixed-post:3.5.0
      com.github.waffle.demo:waffle-mixed:3.5.0
      com.github.waffle.demo:waffle-jaas:3.5.0
      com.github.waffle.demo:waffle-form:3.5.0
      com.github.waffle.demo:waffle-filter:3.5.0
    com.github.waffle:waffle-bom:3.5.0
```

tools/ctl/gradle/gradle.go
algomaster99: I added a Todo because it would involve adding another flag to the command. Thought it would be neater in another PR.

tools/ctl/gradle/gradle_test.go
algomaster99: I just remove it 😅 I think I did it minutes after I requested your review. Sorry for that.

PR #738, opened 2025-09-02
Add threadsafe variant of memfs

msuozzo:


internal/safememfs/safememfs.go
```diff
@@ -0,0 +1,155 @@
...
+
+func (s *SafeMemory) Root() string {
+	return "/"
+}
+
+// --- Write Operations (use exclusive Lock) ---
+
+func (s *SafeMemory) OpenFile(filename string, flag int, perm os.FileMode) (billy.File, error) {
+	// Use the read-lock if read-only, otherwise write-lock.
+	locker := s.mu.RLocker()
```
wbxyz: Is a whole-fs-lock the right level to have an RLocker? Maybe there should be a lock per file/dir? Idk.

wbxyz (approved):

internal/safememfs/safememfs.go
msuozzo: yep this was the original design and i shouldn't have messed with it. the locker should only be protecting the map and at that point, it turns out we could just use a normal mutex.

PR #739, opened 2025-09-02
Add logs viewer role for builder-agent

wbxyz:


msuozzo:
Refresh my memory: why do we need to read logs from this role? This is the GCB execution role?

msuozzo (approved):

PR #740, opened 2025-09-02
Allow agent builder to read from bootstrap tools

wbxyz:


deployments/main.tf
```diff
@@ -1175,6 +1176,11 @@ resource "google_storage_bucket_iam_binding" "builder-agent-uses-logs" {
   role   = "roles/storage.objectUser"
   members = ["serviceAccount:${google_service_account.builder-agent.email}"]
 }
+resource "google_storage_bucket_iam_binding" "builder-agent-view-logs" {
```
msuozzo: This should be conditional on public being true, right?

wbxyz: These are the GCB logs, which I don't think are ever public?

I copied this from the builders-view-logs rule above, which has the regular logs as the subject and the regular builder as the member.

wbxyz: Ah wait sorry, I didn't realize which PR we're in. This is stacked on #739, and the highlighted code is from that other PR. This PR only adds one line to the bootstrap tools binding which *is* conditional on it being public.

msuozzo: Ugh right. I thought I'd put this in the other PR.

PR #741, opened 2025-09-02
Add polling to the run-agent command

wbxyz:


tools/ctl/ctl.go
```diff
@@ -1157,6 +1166,34 @@ var runAgent = &cobra.Command{
 			return errors.Wrap(err, "running attest")
 		}
 		log.Printf("Successfully started session %s", resp.SessionID)
+		sessionDoc := fire.Collection("agent_sessions").Doc(resp.SessionID)
+		var session schema.AgentSession
+		for {
+			// Polling at 1 qps
```
msuozzo: Not sure this comment is correct/necessary

tools/ctl/ctl.go
```diff
@@ -1157,6 +1166,34 @@ var runAgent = &cobra.Command{
...
+			time.Sleep(5 * time.Second)
+			if session.Status != schema.AgentSessionStatusCompleted {
+				sessionSnap, err := sessionDoc.Get(ctx)
+				if err != nil && status.Code(err) == codes.NotFound {
+					continue
+				} else if err != nil {
+					log.Fatal(errors.Wrap(err, "getting session document"))
+				} else if !sessionSnap.Exists() {
```
msuozzo: Are we sure that this isn't redundant given the logic here?

tools/ctl/ctl.go
```diff
@@ -1157,6 +1166,34 @@ var runAgent = &cobra.Command{
...
+					log.Fatal("session not found")
+				}
+				var newSession schema.AgentSession
+				if err := sessionSnap.DataTo(&newSession); err != nil {
+					log.Fatal("reading session data")
```
msuozzo: Nit: I'd probably say parsing or deserializing

tools/ctl/ctl.go
```diff
@@ -1157,6 +1166,34 @@ var runAgent = &cobra.Command{
...
+				}
+				if newSession.Updated != session.Updated {
```
msuozzo, line 1186: Why do we want this here?

wbxyz: Current behavior is to print the entire session document whenever it's updated, ideally to show state changes. I'd like to poll the iteration documents as well to log those when iterations complete but I haven't written that part yet.

msuozzo: Ah so this is currently a proxy for an iteration having been run. Makes sense. Want to add a quick note here to that effect?

tools/ctl/ctl.go
wbxyz: Hmm reading the impl for Get() it looks like that already checks Exists() and returns NotFound if !Exists(). I figured it couldn't hurt to be defensive but I can remove it.

tools/ctl/ctl.go
wbxyz: Agreed, removed.

PR #742, opened 2025-09-02
Add support for Gradle inference

algomaster99:
This PR adds support for inferring the build tool Gradle. While going through the `HEAD` commit, if it finds `build.gradle` or  `build.gradle.kts`, it assumes we are working with a project built with Gradle. If we find both `pom.xml` and `build.gradle`/`build.gradle.kts`, we throw an error for now (we may default to pom.xml).

Todos:
- [ ] test for GradleBuild strategy
- [ ] neatly refactor inference. 1) Gradle has these steps: search in HEAD, find tag, and then it finalizes the build.gradle file. 2) Maven has these steps: search in HEAD, find the precise pom from the history, validate pom via tag guess and pomXML guess, and then it finalizes the pom.xml. I would like to reuse as much as possible.


pkg/rebuild/maven/tools.go
```diff
@@ -0,0 +1,59 @@
...
+		}},
+	},
+	{
+		Name: "maven/gradle-build",
+		Steps: []flow.Step{{
+			Runs: "./gradlew build -x test --no-daemon",
+		}},
+	},
+	{
+		Name: "maven/move-gradle-build-output",
```
wbxyz: Add a TODO that we should find a better way to determine gradle build output path?

pkg/rebuild/maven/strategy_test.go
```diff
@@ -46,6 +46,36 @@ wget -q -O - "https://download.java.net/java/GA/jdk11/13/GPL/openjdk-11.0.1_linu
...
+			rebuild.Instructions{
+				Location: rebuild.Location{
+					Repo: "https://foo.bar",
+					Ref:  "ref",
+					Dir:  "dir",
+				},
+				SystemDeps: []string{"git", "wget"},
+				Source:     "git clone https://foo.bar .\ngit checkout --force 'ref'",
+				Deps: `mkdir -p /opt/jdk
+wget -q -O - "https://download.java.net/java/GA/jdk11/13/GPL/openjdk-11.0.1_linux-x64_bin.tar.gz" | tar -xzf - --strip-components=1 -C /opt/jdk`,
```
wbxyz: nit: Dedent?

algomaster99:
Squashing the two commits so that the other one is not marked `stacked`.

pkg/rebuild/rebuild/strategy.go
```diff
@@ -60,6 +62,10 @@ type Strategy interface {
 	GenerateFor(Target, BuildEnv) (Instructions, error)
 }
 
+type StrategyInferer interface {
```
wbxyz: Is it important to make a public interface for this type? If I understand correctly, the StrategyInferer implementations are internal to the ecosystem specific inference implementation.

I think the direction of moving maven and gradle inference into their own files is a good way to reduce the increasing size of the infer.go file. I also like that both maven and gradle are behind dedicated methods to make it clear what shared data they each depend on (the arguments in the function).

However, I don't think it's necessary to make this abstract interface to get the above benefits. To the contrary - making this interface adds a layer of abstraction that I don't think is really reducing the complexity in maven/infer.go, and I'm not sure I see a place where we need generic StrategyInfer types. The ecosystem specific infer.go will always make an informed decision on which type of strategy it would like to generate, which means it won't be treating them as generic StrategyInferer types anyway.

pkg/rebuild/maven/infer.go
```diff
@@ -57,96 +54,54 @@ func (Rebuilder) CloneRepo(ctx context.Context, t rebuild.Target, repoURI string
...
+	minDepth := -1
+
+	fileIter, _ := commit.Files()
+	fileIter.ForEach(func(f *object.File) error {
+		var currentInferer rebuild.StrategyInferer
+		var currentDepth int
+
+		// Check for Maven's build file
+		if path.Base(f.Name) == "pom.xml" && !strings.HasPrefix(f.Name, "src/") && !strings.Contains(f.Name, "/src/") {
+			currentInferer = &MavenBuildInferer{}
```
wbxyz: If the inferer were eventually a more complicated type with an expensive setup cost, initializing the inferer here might be too much (especially if the search changes its mind and throws away the inferer in a later step). Initializing here also might force any Inferer dependencies to be available to this code doing the search.

WDYT of changing inferBuildSystem to return a enum type that describes it's choice of build system, then the calling code can do any setup that's necessary for that build system?

pkg/rebuild/maven/infer.go
```diff
@@ -57,96 +54,54 @@ func (Rebuilder) CloneRepo(ctx context.Context, t rebuild.Target, repoURI string
...
 
-	name, version := t.Package, t.Version
-	cfg := &MavenBuild{}
-	dir := path.Dir(pkgPath)
-	pomXMLGuess := refMap[version]
-	tagGuess, err := rebuild.FindTagMatch(name, version, repoConfig.Repository)
-	if err != nil {
-		return cfg, errors.Wrapf(err, "[INTERNAL] tag heuristic error")
-	}
+	return buildSystem.Infer(ctx, t, mux, repoConfig, commitObject)
```
wbxyz: If the tool-specific inferences are each going to search for an appropriate git ref, then why do we pass a commitObject into that inference?

If I understand correctly, commitObject here is just the Head commit, which the tool-specific code could still get from the repoConfig if they need.

pkg/rebuild/maven/infer.go
```diff
@@ -57,96 +54,54 @@ func (Rebuilder) CloneRepo(ctx context.Context, t rebuild.Target, repoURI string
...
 	head, _ := repoConfig.Repository.Head()
 	commitObject, _ := repoConfig.Repository.CommitObject(head.Hash())
-	_, pkgPath, err := findPomXML(commitObject, t.Package)
-	if err != nil {
-		log.Printf("pom.xml path heuristic failed [pkg=%s,repo=%s]: %s\n", t.Package, repoConfig.URI, err.Error())
-	}
 
-	// Do version heuristic search.
-	refMap, err := pomXMLSearch(t.Package, pkgPath, repoConfig.Repository)
+	buildSystem, err := inferBuildSystem(commitObject)
```
wbxyz: I think this means we're choosing the buildSystem based on the head commit, even if the build we're targeting is from long ago? In some cases if the project changed build tools since the target artifact was released, that might mean we chose the wrong build tool.

I think that's probably ok for now, but might be worth adding a comment here mentioning it.

pkg/rebuild/maven/gradleinfer.go
```diff
@@ -0,0 +1,150 @@
...
+		return nil, errors.Wrapf(err, "[INTERNAL] tag heuristic error")
+	}
+
+	var ref string
+	if tagGuess != "" {
+		ref = tagGuess
+		log.Printf("using tag heuristic ref: %s", tagGuess[:9])
+	} else {
+		return nil, errors.Errorf("no valid git ref")
+	}
```
wbxyz, line 42: If we're always using the tagGuess, should we do the ref inference first, then get the ref's commitObject to use when running the findBuildGradle() method? I think that would be more resilient to any restructuring in the repo.

pkg/rebuild/maven/gradleinfer.go
```diff
@@ -0,0 +1,150 @@
...
+		// Skip files in gradle/ directory
+		// This directory often contains wrapper scripts and other configuration files
+		// https://docs.gradle.org/current/userguide/gradle_directories.html
+		if strings.HasPrefix(f.Name, "gradle/") {
+			return nil
+		}
+		if (strings.HasSuffix(f.Name, ".gradle") || strings.HasSuffix(f.Name, ".gradle.kts")) && !strings.HasPrefix(f.Name, "src/") && !strings.Contains(f.Name, "/src/") {
+			candidateArtifactNames = append(candidateArtifactNames, path.Dir(f.Name))
+
+			parseAndSetGroupID(f, &topLevelGroupID)
```
wbxyz: Why is this passing topLevelGroupID by reference rather than returning a string that can be assigned to topLevelGroupID?

pkg/rebuild/maven/gradleinfer.go
```diff
@@ -0,0 +1,150 @@
...
+	var candidateArtifactNames []string
+	var topLevelGroupID string
+	commitTree.Files().ForEach(func(f *object.File) error {
+		// Skip files in gradle/ directory
+		// This directory often contains wrapper scripts and other configuration files
+		// https://docs.gradle.org/current/userguide/gradle_directories.html
+		if strings.HasPrefix(f.Name, "gradle/") {
+			return nil
+		}
+		if (strings.HasSuffix(f.Name, ".gradle") || strings.HasSuffix(f.Name, ".gradle.kts")) && !strings.HasPrefix(f.Name, "src/") && !strings.Contains(f.Name, "/src/") {
```
wbxyz: Should the check of src/ and /src/ be in the same block as the "gradle/" check?

pkg/rebuild/maven/gradleinfer.go
```diff
@@ -0,0 +1,150 @@
...
+			candidateArtifactNames = append(candidateArtifactNames, path.Dir(f.Name))
+
+			parseAndSetGroupID(f, &topLevelGroupID)
+		}
+		if f.Name == "gradle.properties" {
+			parseAndSetGroupID(f, &topLevelGroupID)
```
wbxyz: If either of these set the topLevelGroupID, should the search terminate? I think you can do that by returning object.ErrStop

pkg/rebuild/maven/strategy_test.go
wbxyz, line 1: Now that the strategies are in separate files, should the tests be in matching separate files? Or maybe the strategies should be in the same strategy.go file, leaving inference to be separate?

pkg/rebuild/maven/tools.go
algomaster99: I believe I had removed the code related to this much before your reviews. Anyway, instead of using `find`, we do a string similarity search using minimum levenshtein distance + checking if directory name is a substring in `pkg`. It seems to give good outputs right now. But I will validate this later today by running a benchmark of Gradle builds and look for failures due to find the artifact.

cc @msuozzo

pkg/rebuild/maven/infer.go
algomaster99: I agree with this and [the comment above](https://github.com/google/oss-rebuild/pull/742#discussion_r2330617132). I have used enums for the same purpose for the performance reason. The specific inferences are now renamed to `MavenInfer` and `GradleInfer`.

pkg/rebuild/maven/infer.go
algomaster99: I think it is the remnants of separating maven and Gradle inference. Both of them start with using head commit so its logic just stayed there and I reused it. I have removed this and computed it individually in the inferences.

pkg/rebuild/maven/infer.go
algomaster99: I also thought about this that we should prioritise checking for version first, preferably using tag match. There is just one downside to this as there are cases where the release is done from a detached commit and not even tagged. I left a todo.

pkg/rebuild/maven/gradleinfer.go
algomaster99: This comment is on the same lines as [the one above](https://github.com/google/oss-rebuild/pull/742#discussion_r2331020158). I have added a todo. Let's take this up in a new PR.

pkg/rebuild/maven/gradleinfer.go
algomaster99: I wanted to make sure that once a group ID is found, this variable is not set again. I did not want to add this check twice. Is it a bad idea to pass string as references? 
```go
func parseAndSetGroupID(f *object.File, topLevelGroupID *string) {
	if *topLevelGroupID != "" {
		return
	}
	content, err := f.Contents()
	if err != nil {
		return
	}
	regex := regexp.MustCompile(`group\s*=?\s*['"]?([^'"\s]+)['"]?`)
	matcher := regex.FindStringSubmatch(content)
	if len(matcher) > 1 {
		*topLevelGroupID = matcher[1]
	}
}
```

pkg/rebuild/maven/gradleinfer.go
algomaster99: Yep sounds like what I need.

pkg/rebuild/maven/strategy_test.go
algomaster99: Test for strategies can be in the same file as they are still very similar. The only difference is the build command.
Test for inference may be in the same file too as it has tests which are common to both build systems (and more in future) like JDK verision inference.

Also, I don't have Gradle specific inference tests yet. I will add it.

pkg/rebuild/maven/gradleinfer.go
algomaster99: I did not find a built-in error for this so I just introduced a var, although the code looks a bit redundant now.

pkg/rebuild/maven/strategy_test.go
algomaster99: On a second thought, it is nice to refactor them.
```
infer.go -> infer_test.go
maveninfer.go -> maveninfer_test.go
gradleinfer.go -> gradleinfer_test.go
```

pkg/rebuild/maven/gradleinfer.go
```diff
@@ -0,0 +1,180 @@
...
+	err := commitTree.Files().ForEach(func(f *object.File) error {
+		// Skip files in gradle/ directory
+		// This directory often contains wrapper scripts and other configuration files
+		// https://docs.gradle.org/current/userguide/gradle_directories.html
+		if strings.HasPrefix(f.Name, "gradle/") || strings.HasPrefix(f.Name, "src/") || strings.Contains(f.Name, "/src/") {
+			return nil
+		}
+		if strings.HasSuffix(f.Name, ".gradle") || strings.HasSuffix(f.Name, ".gradle.kts") {
+			candidateArtifactNames = append(candidateArtifactNames, path.Dir(f.Name))
+			content, err := readFileContents(f)
```
algomaster99: I separated out reading because it made it easier to write test for it in `TestGetGroupIDFromFile`.

pkg/rebuild/maven/gradleinfer.go
```diff
@@ -0,0 +1,180 @@
...
+			groupID, err := getGroupIDFromFile(content)
+			if err != nil {
+				return err
+			}
+			if groupID != "" {
+				topLevelGroupID = groupID
+				return stopIteration
+			}
+		}
+		return nil
```
algomaster99: This is the redundancy I get if I don't pass by reference.

pkg/rebuild/maven/gradleinfer.go
```diff
@@ -0,0 +1,180 @@
...
+
+func GradleInfer(ctx context.Context, t rebuild.Target, mux rebuild.RegistryMux, repoConfig *rebuild.RepoConfig) (rebuild.Strategy, error) {
+	head, _ := repoConfig.Repository.Head()
+	commitObject, _ := repoConfig.Repository.CommitObject(head.Hash())
+
+	buildGradleDir, err := findBuildGradleDir(commitObject, t.Package)
+	if err != nil {
+		return nil, errors.Wrapf(err, "build manifest heuristic failed")
+	}
+	name, version := t.Package, t.Version
```
msuozzo, line 30: inline these

pkg/rebuild/maven/gradleinfer.go
```diff
@@ -0,0 +1,180 @@
...
+
+func readFileContents(f *object.File) (string, error) {
+	content, err := f.Contents()
+	if err != nil {
+		return "", errors.Wrapf(err, "reading file %s", f.Name)
+	}
+	return content, nil
+}
+
+func minEditDistance(s1, s2 string) int {
```
msuozzo, line 142: might want to mention in a comment that this is levenshtein distance

pkg/rebuild/maven/gradleinfer.go
```diff
@@ -0,0 +1,180 @@
...
+	len1 := len(s1)
+	len2 := len(s2)
+	dp := make([][]int, len1+1)
+	for i := range dp {
+		dp[i] = make([]int, len2+1)
+	}
+
+	for i := 0; i <= len1; i++ {
```
msuozzo, line 150: nit: range len1+1

pkg/rebuild/maven/gradleinfer.go
```diff
@@ -0,0 +1,180 @@
...
+		dp[i][0] = i
+	}
+	for j := 0; j <= len2; j++ {
```
msuozzo, line 153: and here

pkg/rebuild/maven/gradleinfer.go
```diff
@@ -0,0 +1,180 @@
...
+func getGroupIDFromFile(content string) (string, error) {
+	regex := regexp.MustCompile(`group\s*=?\s*['"]?([^'"\s]+)['"]?`)
+	matcher := regex.FindStringSubmatch(content)
+	if len(matcher) > 1 {
+		return matcher[1], nil
+	}
+	return "", nil
+}
+
+func readFileContents(f *object.File) (string, error) {
```
msuozzo: i'd inline these. better to use the git infrastructure to do a more expansive test than to poke through interfaces like this for mocking out during testing.

pkg/rebuild/maven/gradleinfer.go
```diff
@@ -0,0 +1,180 @@
...
+			Dir:  buildGradleDir,
+			Ref:  ref,
+		},
+		JDKVersion: jdk,
+	}, nil
+}
+
+func findBuildGradleDir(commit *object.Commit, pkg string) (string, error) {
+	commitTree, _ := commit.Tree()
+	var candidateArtifactNames []string
```
msuozzo, line 62: i don't think this is an artifact name, right? these are directories with gradle build defns, right?

pkg/rebuild/maven/gradleinfer.go
```diff
@@ -0,0 +1,180 @@
...
+	var topLevelGroupID string
+
+	err := commitTree.Files().ForEach(func(f *object.File) error {
+		// Skip files in gradle/ directory
+		// This directory often contains wrapper scripts and other configuration files
+		// https://docs.gradle.org/current/userguide/gradle_directories.html
+		if strings.HasPrefix(f.Name, "gradle/") || strings.HasPrefix(f.Name, "src/") || strings.Contains(f.Name, "/src/") {
```
msuozzo, line 69: why is `src` excluded? that explanation doesn't appear in the comment.

pkg/rebuild/maven/gradleinfer.go
```diff
@@ -0,0 +1,180 @@
...
+			if err != nil {
+				return err
+			}
+			groupID, err := getGroupIDFromFile(content)
+			if err != nil {
+				return err
+			}
+			if groupID != "" {
+				topLevelGroupID = groupID
+				return stopIteration
```
msuozzo, line 90: is there an explanation for why we choose the first groupID we encounter? do we expect to see one in .properties AND .gradle ever? can they be different? given that logic below this relies on the latter being seen at least once, do we ever expect to see the former first and exit before seeing a .gradle file?

pkg/rebuild/maven/gradleinfer.go
```diff
@@ -0,0 +1,180 @@
...
+			minDist = dist
+			bestMatch = candidate
+		}
+	}
+	log.Printf("Found best match with minimum edit distance: %s (distance %d)", bestMatch, minDist)
+	return bestMatch, nil
+}
+
+func getGroupIDFromFile(content string) (string, error) {
+	regex := regexp.MustCompile(`group\s*=?\s*['"]?([^'"\s]+)['"]?`)
```
msuozzo, line 134: would be nice to move this to a package-level variable and reference it here.

pkg/rebuild/maven/strategy.go
```diff
@@ -22,6 +21,15 @@ type MavenBuild struct {
 
 var _ rebuild.Strategy = &MavenBuild{}
 
+type GradleBuild struct {
```
msuozzo, line 24: imo i'd define the type right above the methods i.e. move this down

pkg/rebuild/maven/maveninfer_test.go
```diff
@@ -0,0 +1,262 @@
...
+		PackageName: fmt.Sprintf("%s:%s", pom.GroupID, pom.ArtifactID),
+		VersionID:   pom.VersionID,
+		FileType:    maven.TypePOM,
+	}
+	xml := "<project xmlns=\"http://maven.apache.org/POM/4.0.0\" xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\" xsi:schemaLocation=\"http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd\"><groupId>" + pom.GroupID + "</groupId><artifactId>" + pom.ArtifactID + "</artifactId><version>" + pom.VersionID + "</version>"
+	// Add parent if present
+	if pom.Parent.GroupID != "" && pom.Parent.ArtifactID != "" && pom.Parent.VersionID != "" {
+		xml += "<parent><groupId>" + pom.Parent.GroupID + "</groupId><artifactId>" + pom.Parent.ArtifactID + "</artifactId><version>" + pom.Parent.VersionID + "</version></parent>"
+	}
+	xml += "<url>" + pom.URL + "</url><scm><url>" + pom.SCMURL + "</url></scm></project>"
```
msuozzo, line 260: can you construct and serialize a PomXML struct instead?

pkg/rebuild/maven/infer.go
```diff
@@ -57,96 +59,64 @@ func (Rebuilder) CloneRepo(ctx context.Context, t rebuild.Target, repoURI string
...
-				log.Printf("using tag heuristic ref: %s", tagGuess[:9])
-				ref = tagGuess
-				dir = filepath.Dir(newPath)
-				break
-			}
-		} else if err == plumbing.ErrObjectNotFound {
-			log.Printf("tag heuristic ref not found in repo")
-		} else {
-			return cfg, errors.Wrapf(err, "[INTERNAL] Failed ref resolve from tag [repo=%s,ref=%s]", repoConfig.URI, tagGuess)
+func inferBuildTool(commit *object.Commit) (string, error) {
```
msuozzo, line 83: it would be good to have some tests for this, as well. especially to document the precedence that these tools have.

pkg/rebuild/maven/gradleinfer.go
algomaster99: That's correct. I called them that because they are candidates of artifact names. I can change that to something more related to directory.

algomaster99: Changed it to `candidateDirs`

pkg/rebuild/maven/gradleinfer.go
algomaster99: I first thought that the files returned by `object.FileIter` would be in BFS order so we would get the group ID from, the `build.gradle` closest to root. We want it to be closest to root because then it can apply to all the subprojects. This was not the case. We probably need to ensure that the depth is minimum. 

I have done something similar in [`infer.go`](https://github.com/google/oss-rebuild/pull/742/files#diff-12392ac4d5a92c6f764d315ebd0bbce56a8c745cbfe6be00500fdce4d799a3fcR104-R109).

algomaster99: Let me take this up in another PR for Gradle.

pkg/rebuild/maven/infer.go
algomaster99: Resolved in #765

pkg/rebuild/maven/maveninfer_test.go
algomaster99: Much better idea.

PR #747, opened 2025-09-03
Add benchmark for grouping by GroupID for maven packages

algomaster99:
During inference, the errors are often duplicated in the same group ID. This will ensure that benchmark has unique group IDs and which roughly implies unique source repositories.

### DIFF
--- a/tools/benchmark/generate/main.go
+++ b/tools/benchmark/generate/main.go
@@ -48,6 +48,7 @@ var all = []RebuildBenchmark{
 	npmTop500,
 	npmTop2500,
 	mavenTop500,
+	mavenTop500UniqueGroups,
 }
 
 const (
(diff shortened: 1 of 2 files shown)


algomaster99:
@wbxyz should we do this in the SQL query instead? The tables don't seem to have individual group and artifact columns.

tools/benchmark/generate/main.go
```diff
@@ -837,6 +838,140 @@ LIMIT 2500
...
+					break
+				}
+				if err != nil {
+					log.Fatal(err.Error())
+				}
+				pkgs <- entry
+			}
+			close(pkgs)
+		}()
+		seenGroups := make(map[string]struct{})
```
wbxyz, line 926: If this were a map[string]bool then you can assign true as the value and just do something like `if seenGroups[g] { ...`

tools/benchmark/generate/main.go
```diff
@@ -837,6 +838,140 @@ LIMIT 2500
...
+					continue
+				}
+				ps.Packages = append(ps.Packages, benchmark.Package{Name: p.Package, Ecosystem: "maven"})
+				idx = len(ps.Packages) - 1
+			}
+			psp := &ps.Packages[idx]
+			if len(psp.Versions) >= 5 {
+				continue
+			}
+			nameParts := strings.SplitN(p.Package, ":", 2)
```
wbxyz, line 958: Can this be combined with the group split above?

tools/benchmark/generate/main.go
```diff
@@ -837,6 +838,140 @@ LIMIT 2500
...
+var mavenTop500UniqueGroups = RebuildBenchmark{
+	Filename: "maven_top_500_unique_groups.json",
+	Generator: func(ctx context.Context) (ps benchmark.PackageSet) {
+		now := time.Now()
+		client, err := bigquery.NewClient(ctx, *project, option.WithQuotaProject(*project))
+		if err != nil {
+			log.Fatal(err.Error())
+		}
+		query := client.Query(`
+SELECT
```
wbxyz, line 850: I think this SQL needs to be updated following some of the other PRs

PR #749, opened 2025-09-04
Show cases where regex matches in TUI

algomaster99:


tools/ctl/ide/commands/commands.go
```diff
@@ -431,6 +431,7 @@ func NewRebuildGroupCmds(app *tview.Application, rb *rebuilder.Rebuilder, modalF
 						if regex.MatchString(line) {
 							log.Printf("%s\n\t%s", in.ID(), line)
 							out <- in
+							found++
```
msuozzo: i think this is better to increment in the loop consuming p.Out(). might as well move the variable declaration down there, as well.

algomaster99:
@msuozzo ping for merge

PR #754, opened 2025-09-04
Implement agent job logic

wbxyz:


internal/agent/logic.go
```diff
@@ -0,0 +1,51 @@
...
+		}
+		resp, err := inferenceservice.Infer(ctx, req, deps)
+		if err != nil {
+			return nil, errors.Wrap(err, "inferring initial strategy")
+		}
+		return resp, nil
+	}
+	// TODO: Implement actual agent logic here
+	return nil, errors.New("agent logic not yet implemented")
+	// return &schema.StrategyOneOf{
```
msuozzo: .

cmd/agent/main.go
```diff
@@ -65,27 +67,21 @@ func main() {
...
-		SessionID:  *sessionID,
-		StopReason: "NO_OP",
-		Summary:    "Agent completed as no-op stub",
+	// Create agent API client stubs
+	iterationStub := api.Stub[schema.AgentCreateIterationRequest, schema.AgentCreateIterationResponse](client, baseURL.JoinPath("agent/session/iteration"))
+	completeStub := api.Stub[schema.AgentCompleteRequest, schema.AgentCompleteResponse](client, baseURL.JoinPath("agent/session/complete"))
+	deps := agent.RunSessionDeps{
+		IterationStub: iterationStub,
+		CompleteStub:  completeStub,
+		// TODO: Should we allow users to swap this out?
```
msuozzo: "Should we make this configurable?"

internal/agent/session.go
```diff
@@ -0,0 +1,121 @@
...
+		}
+		log.Println("Session cleanup completed")
+	}()
+	var iteration *schema.AgentIteration
+	var iterNum int
+	for range req.MaxIterations {
+		iterNum++
+		log.Printf("Iteration %d", iterNum)
+		log.Println("Allowing the agent to propose a strategy")
+		s, err := deps.Agent.ProposeStrategy(ctx, req.Target, iteration)
```
msuozzo: okay so i think we should be able to pull out this loop body into a per-iteration helper that returns response and an error. The Error cases here can be modeled as errors and the success can return the iteration.

max iteration failures would need to be handled outside this function.

what this gets us is more natural control flow and a bit more brevity. I don't think we sacrifice much clarity in exchange.

internal/agent/session.go
```diff
@@ -0,0 +1,121 @@
...
+
+type RunSessionReq struct {
+	SessionID     string
+	Target        rebuild.Target
+	MaxIterations int
+}
+
+type RunSessionDeps struct {
+	IterationStub func(context.Context, schema.AgentCreateIterationRequest) (*schema.AgentCreateIterationResponse, error)
+	CompleteStub  func(context.Context, schema.AgentCompleteRequest) (*schema.AgentCompleteResponse, error)
```
msuozzo: these types can be api.Stub types, right?

internal/agent/logic.go
```diff
@@ -0,0 +1,51 @@
...
+			GitCache:   nil,
+		}
+		resp, err := inferenceservice.Infer(ctx, req, deps)
+		if err != nil {
+			return nil, errors.Wrap(err, "inferring initial strategy")
+		}
+		return resp, nil
+	}
+	// TODO: Implement actual agent logic here
+	return nil, errors.New("agent logic not yet implemented")
```
algomaster99, line 45: I was wondering how we would parse the non-deterministic output of an LLM into a valid `rebuild.Strategy`, but apparently [there is a way to get structured output from an LLM](https://ai.google.dev/gemini-api/docs/structured-output#go).

We do have `GenerateTypedContent` in `llm.go` but it is based on `GenerateTextContent` which may generate unparseable text sometimes. Should I propose a PR for an API returning structured content?

internal/agent/session.go
wbxyz: I think I did what you were suggesting, but take a look. It turned out to be a bit more complicated than ~3 cases, but I think the control flow is easier to follow now.

internal/agent/logic.go
msuozzo: GenerateTypedContent uses ResponseSchema already which is the API in the example you linked.

internal/agent/session.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+		switch iteration.Status {
+		case schema.AgentIterationStatusSuccess:
+			return &schema.AgentCompleteRequest{
+				StopReason:         schema.AgentCompleteReasonSuccess,
+				Summary:            "Build successful",
+				SuccessIterationID: iteration.ID,
+			}
+		case schema.AgentIterationStatusFailed:
+			prev = iteration
```
msuozzo, line 91: nit: might want to add a comment noting this behavior that we retry failures.

internal/agent/session.go
msuozzo: yeah this looks great!

internal/agent/session.go
wbxyz: I think retrying failures is the "work" of the agent, so shouldn't be surprising right? Or maybe something else about this is surprising.

Unless you meant to highlight the section below where we throw away errors and attempt again?

msuozzo: eh I just figured it would be nice to see something like: "On unsuccessful rebuilds, try another rebuild using this failed one as input.". If I see "failure", I might think more 'unrecoverable' than 'use as input to the next iteration'. but i can also see the assumption that it's clear on the face of it. your choice.

wbxyz: Ah yeah I can see that being confusing. I'll add a note.

PR #755, opened 2025-09-04
Add local-agent cli command

wbxyz:


tools/ctl/ctl.go
```diff
@@ -1184,9 +1185,91 @@ var runAgent = &cobra.Command{
...
+			TimeoutSeconds: 60 * 60, // 1 hr
+			Context:        &schema.AgentContext{},
+			Status:         schema.AgentSessionStatusRunning,
+			JobName:        "",
+			Created:        sessionTime,
+			Updated:        sessionTime,
+		}
+		// Create session in Firestore
+		err = fire.RunTransaction(ctx, func(ctx context.Context, t *firestore.Transaction) error {
+			// NOTE: This would fail if the
```
msuozzo: if the

tools/ctl/ctl.go
```diff
@@ -1184,9 +1185,91 @@ var runAgent = &cobra.Command{
...
+		}
+		if *metadataBucket == "" {
+			return errors.New("metadata bucket not provided")
+		}
+		fire, err := firestore.NewClient(cmd.Context(), *project)
+		if err != nil {
+			return errors.Wrap(err, "creating firestore client")
+		}
+		sessionTime := time.Now().UTC()
+		sessionID := fmt.Sprintf("session-%s", sessionTime.Format(time.RFC3339))
```
msuozzo: i'd be fine with including something indicating this was local. there's no purpose in my mind to putting "session" in the session key, though.

tools/ctl/ctl.go
```diff
@@ -1184,9 +1185,91 @@ var runAgent = &cobra.Command{
...
+		// Set defaults for configuration
+		maxIterations := 10
+		session := schema.AgentSession{
+			ID:             sessionID,
+			Target:         t,
+			MaxIterations:  maxIterations,
+			TimeoutSeconds: 60 * 60, // 1 hr
+			Context:        &schema.AgentContext{},
+			Status:         schema.AgentSessionStatusRunning,
+			JobName:        "",
```
msuozzo: what's going on here?

tools/ctl/ctl.go
```diff
@@ -1184,9 +1185,91 @@ var runAgent = &cobra.Command{
 		return nil
 	},
 }
+
+var localAgent = &cobra.Command{
+	Use:   "local-agent --project <project> --agent-api <URI> --metadata-bucket <bucket> --ecosystem <ecosystem> --package <name> --version <version> --artifact <name>",
+	Short: "Run benchmark",
```
msuozzo: nit:

tools/ctl/ctl.go
```diff
@@ -1184,9 +1185,91 @@ var runAgent = &cobra.Command{
...
+		if err != nil {
+			log.Fatalf("Failed to parse agent API URL: %v", err)
+		}
+		// Create agent API client stubs
+		iterationStub := api.Stub[schema.AgentCreateIterationRequest, schema.AgentCreateIterationResponse](client, baseURL.JoinPath("agent/session/iteration"))
+		completeStub := api.Stub[schema.AgentCompleteRequest, schema.AgentCompleteResponse](client, baseURL.JoinPath("agent/session/complete"))
+		deps := agent.RunSessionDeps{
+			IterationStub:  iterationStub,
+			CompleteStub:   completeStub,
+			Agent:          agent.NewDefaultAgentLogic(),
```
msuozzo, line 1264: didn't see this in the previous PR but I'm not a fan of "logic" here. we could just call this "Agent", right?

msuozzo (commented):
generally looks good. might be nice to include a note on the assumptions of the impl.

tools/ctl/ctl.go
wbxyz: This as copied from the "regular" way of making sessions. I'm trying to keep them in-line with each other.

wbxyz: Given #756 making changes to sessionID, I rebased this PR onto that one.

The session record initialization here matches that code almost exactly. I briefly tried to find a way to re-use the AgentCreate method by just injecting a different stub for the job start but I couldn't immediately find a good interface for that so I decided to just copy and assume neither will change much once we've settled on something.

tools/ctl/ctl.go
wbxyz: Normally a session relates to a Cloud Run Job and this would be the job name, but we aren't doing cloud run so that's not applicable.

tools/ctl/ctl.go
wbxyz: It's not really the whole "Agent" as much as the internal logic of one loop of an agent. I added an AgentLogic interface which this is returning, but that might be unnecessary and it could maybe just be a function signature instead. I'll be touching that code in followup PRs so we can revisit then.

PR #756, opened 2025-09-04
Update sessionID and iterationID formats

wbxyz:
Changing sessionID to be time-sortable for debugging, and change iterationID to have more granularity to avoid collisions

### DIFF
--- a/internal/api/apiservice/agent.go
+++ b/internal/api/apiservice/agent.go
@@ -35,7 +35,12 @@ type AgentCreateDeps struct {
 }
 
 func AgentCreate(ctx context.Context, req schema.AgentCreateRequest, deps *AgentCreateDeps) (*schema.AgentCreateResponse, error) {
-	sessionID := uuid.New().String()
+	sessionUUID, err := uuid.NewV7()
+	if err != nil {
+		return nil, errors.Wrap(err, "making sessionID")
+	}
+	sessionID := sessionUUID.String()
+	sessionTime := time.Unix(sessionUUID.Time().UnixTime())
 	jobName := fmt.Sprintf("agent-%s", sessionID)
 	// Set defaults for configuration
 	maxIterations := req.MaxIterations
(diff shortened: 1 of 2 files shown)


internal/api/apiservice/agent.go
```diff
@@ -35,8 +35,9 @@ type AgentCreateDeps struct {
 }
 
 func AgentCreate(ctx context.Context, req schema.AgentCreateRequest, deps *AgentCreateDeps) (*schema.AgentCreateResponse, error) {
-	sessionID := uuid.New().String()
-	jobName := fmt.Sprintf("agent-%s", sessionID)
+	sessionTime := time.Now().UTC()
+	sessionID := fmt.Sprintf("session-%s-%d", sessionTime.Format(time.RFC3339), uuid.New().ID()) // TODO: Is there a more compact way to encode the UUID?
```
msuozzo: not a fan of this. what's the motivation?

wbxyz: Using just the UUID as a string makes the pantheon UI and any other cli-based listing method that sorts the documents both cluttered and pretty unusable for debugging LLM sessions. Including a timestamp prefix means they show up sorted, while keeping them un-guessable with a UUID-worth of randomness. Otherwise we would need to create indexes on fields and probably a new CLI tool to look at the recent session documents in firestore.

It's worth noting that representing the UUID as an int is much more information dense than .String() (10 chars instead of 36 chars) so if you're worried about length we could tweak the time format and still keep it at the same size.

The "session" prefix was to match the existing iteration format, and would help make it obvious if ID types were being misused somewhere. We could either drop it or shorten it to just "S" vs "I".

msuozzo: > cluttered and pretty unusable...we would need to create indexes on fields

the default UI being navigable feels like a marginal use-case. given the expected scale, this isn't going to be particularly usable in the common case, anyway. if we only care about sort order, we could use [uuidv7]( https://pkg.go.dev/github.com/google/uuid#NewV7) which has lexicographic chronological sort order at the cost of ~50 bits of entropy.

wbxyz: I had looked at uuidV7 before proposing this but I was under the impression that the timestamp was in the middle of the format. I think gemini might have mislead me :(

Using uuidV7 sounds good.

internal/api/apiservice/agent.go
```diff
@@ -35,7 +35,12 @@ type AgentCreateDeps struct {
 }
 
 func AgentCreate(ctx context.Context, req schema.AgentCreateRequest, deps *AgentCreateDeps) (*schema.AgentCreateResponse, error) {
-	sessionID := uuid.New().String()
+	sessionTime := time.Now().UTC()
+	sessionUUID, err := uuid.NewV7()
```
msuozzo, line 38: aha it seems there's a Time() method on uuid so you can invert these two statements to pull the time out for sessionTime

wbxyz: Yeah, unfortunately uuid.Time() doesn't return a time.Time object, so I had to do some conversion but got there in the end.

PR #761, opened 2025-09-08
Fix timewarp configuration

algomaster99:
The `timewarp` setting should be set by the `UseTimewarp` and not by URL. Also, there is a default URL set above in the same file. Don't we have parameters to set it? Why is there a default hardcoded here?


algomaster99:
@wbxyz ping for merge

PR #765, opened 2025-09-08
Split inference depending upon build tool

algomaster99:
Split 1/2 of #742 

- [x] https://github.com/google/oss-rebuild/pull/742/files#r2331396896

### DIFF
--- a/pkg/rebuild/maven/infer.go
+++ b/pkg/rebuild/maven/infer.go
@@ -7,17 +7,15 @@ import (
 	"archive/zip"
 	"bytes"
 	"context"
-	"encoding/xml"
 	"fmt"
 	"io"
 	"log"
+	"math"
 	"path"
-	"path/filepath"
 	"strings"
 
 	"github.com/go-git/go-billy/v5"
 	"github.com/go-git/go-git/v5"
-	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/go-git/go-git/v5/plumbing/object"
 	"github.com/go-git/go-git/v5/plumbing/transport"
 	"github.com/go-git/go-git/v5/storage"
@@ -25,6 +23,10 @@ import (
 	"github.com/google/oss-rebuild/pkg/rebuild/rebuild"
 	"github.com/google/oss-rebuild/pkg/registry/maven"
 	"github.com/pkg/errors"
+)
+
+const (
+	mavenBuildTool = "maven"
 )
 
 const fallbackJDK = "11"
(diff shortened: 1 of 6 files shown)


pkg/rebuild/maven/infer_test.go
```diff
@@ -178,245 +172,71 @@ func TestGetClassFileMajorVersion(t *testing.T) {
...
-			}
-			for _, pom := range tc.pom {
-				addPomArtifact(mockRegistry, &pom)
-			}
-			mockMux := rebuild.RegistryMux{
-				Maven: mockRegistry,
-			}
-			got, err := Rebuilder{}.InferRepo(context.Background(), tc.targetPom, mockMux)
+			repo := must(gitxtest.CreateRepoFromYAML(tc.repo, nil))
+			head, _ := repo.Head()
```
msuozzo: use must() here and below, just in case

pkg/rebuild/maven/maveninfer_test.go
```diff
@@ -0,0 +1,258 @@
...
+					t.Errorf("InferRepo() = %q, want %q", got, tc.expectedURL)
+				}
+			}
+
+		})
+	}
+}
+
+func addPomArtifact(mavenRegistry *mockMavenRegistry, pom *PomXML) {
+	key := struct{ PackageName, VersionID, FileType string }{
```
msuozzo, line 252: in a follow-on change, I think we should just name this type so we can stop using this verbose double-declare pattern

pkg/rebuild/maven/infer.go
```diff
@@ -57,96 +59,56 @@ func (Rebuilder) CloneRepo(ctx context.Context, t rebuild.Target, repoURI string
...
-				dir = filepath.Dir(newPath)
-				break
-			}
-		} else if err == plumbing.ErrObjectNotFound {
-			log.Printf("tag heuristic ref not found in repo")
-		} else {
-			return cfg, errors.Wrapf(err, "[INTERNAL] Failed ref resolve from tag [repo=%s,ref=%s]", repoConfig.URI, tagGuess)
+func inferBuildTool(commit *object.Commit) (string, error) {
+	var bestBuildTool string
+	minDepth := math.MaxInt
```
msuozzo, line 85: can you add a brief explanatory comment saying you prefer prefer a match that's the closest to the repo root?

pkg/rebuild/maven/infer.go
```diff
@@ -57,96 +59,56 @@ func (Rebuilder) CloneRepo(ctx context.Context, t rebuild.Target, repoURI string
...
+
+	fileIter, _ := commit.Files()
+	fileIter.ForEach(func(f *object.File) error {
+		var currentBuildTool string
+		var currentDepth int
+
+		// Check for Maven's build file
```
msuozzo, line 94: and can you note here why this src/ path exclusion exists?

pkg/rebuild/maven/infer.go
```diff
@@ -57,96 +59,56 @@ func (Rebuilder) CloneRepo(ctx context.Context, t rebuild.Target, repoURI string
...
+		if path.Base(f.Name) == "pom.xml" && !strings.HasPrefix(f.Name, "src/") && !strings.Contains(f.Name, "/src/") {
+			currentBuildTool = mavenBuildTool
+			currentDepth = strings.Count(f.Name, "/")
```
msuozzo: you should be able to unconditionally set this. in fact, you could eagerly skip all cases where it's already greater than the minDepth.

pkg/rebuild/maven/infer.go
```diff
@@ -57,96 +59,56 @@ func (Rebuilder) CloneRepo(ctx context.Context, t rebuild.Target, repoURI string
...
-		return pomXML, err
+		return nil
+	})
+
+	if bestBuildTool != "" {
+		return bestBuildTool, nil
 	}
-	return pomXML, xml.Unmarshal([]byte(p), &pomXML)
+
+	return "", errors.Errorf("only Maven is supported, but no pom.xml found")
```
msuozzo: nit: maybe just "no pom.xml found" for now.

msuozzo (approved):

pkg/rebuild/maven/infer.go
algomaster99: I did not do this to avoid repetition of if check in both Maven and Gradle. I can do this here.

pkg/rebuild/maven/infer.go
algomaster99: I just copied it from the `maveninfer`.

pkg/rebuild/maven/infer.go
```diff
@@ -57,96 +59,58 @@ func (Rebuilder) CloneRepo(ctx context.Context, t rebuild.Target, repoURI string
...
+// In a typical multi-module project, the root project's pom.xml (for Maven) or build.gradle (for Gradle) defines groupId for the entire project.
+// This parent groupId is then inherited by all sub-modules.
+// This structure ensures that all related modules are grouped under a common namespace, simplifying dependency management and versioning across the project.
+func inferBuildTool(commit *object.Commit) (string, error) {
+	var bestBuildTool string
+	minDepth := math.MaxInt
+
+	fileIter, _ := commit.Files()
+	fileIter.ForEach(func(f *object.File) error {
+
```
msuozzo: nit: remove this whitespace.

algomaster99:
@msuozzo ping for merge

PR #766, opened 2025-09-08
Add support for Gradle inference

algomaster99:
Split 2/2 of #742 

- [x] https://github.com/google/oss-rebuild/pull/742#discussion_r2331372879

### DIFF
--- a/pkg/rebuild/maven/gradleinfer.go
+++ b/pkg/rebuild/maven/gradleinfer.go
@@ -1,0 +1,162 @@
+// Copyright 2025 Google LLC
+// SPDX-License-Identifier: Apache-2.0
+
+package maven
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"math"
+	"path"
+	"regexp"
+	"strings"
+
+	"github.com/go-git/go-git/v5/plumbing/object"
+	"github.com/google/oss-rebuild/pkg/rebuild/rebuild"
+	"github.com/pkg/errors"
+)
+
+func GradleInfer(ctx context.Context, t rebuild.Target, mux rebuild.RegistryMux, repoConfig *rebuild.RepoConfig) (rebuild.Strategy, error) {
+	head, _ := repoConfig.Repository.Head()
+	commitObject, _ := repoConfig.Repository.CommitObject(head.Hash())
+	buildGradleDir, err := findBuildGradleDir(commitObject, t.Package)
+	if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/rebuild/maven/gradleinfer.go
```diff
@@ -0,0 +1,167 @@
...
+	"path"
+	"regexp"
+	"strings"
+
+	"github.com/go-git/go-git/v5/plumbing/object"
+	"github.com/google/oss-rebuild/pkg/rebuild/rebuild"
+	"github.com/pkg/errors"
+)
+
+var groupIDRegex = regexp.MustCompile(`(?m)^\s*<groupId>([^<]+)</groupId>`)
```
msuozzo: `[^<]+?`

pkg/rebuild/maven/infer.go
```diff
@@ -98,13 +101,20 @@ func inferBuildTool(commit *object.Commit) (string, error) {
 			bestBuildTool = mavenBuildTool
 			minDepth = currentDepth
 		}
+		// Check if Gradle wrapper is present
+		// It is common practice to include the Gradle wrapper script (`gradlew`) at the root of the project.
+		// Referenence: https://docs.gradle.org/current/userguide/gradle_wrapper_basics.html
+		if path.Base(f.Name) == "gradlew" {
```
msuozzo, line 107: might want a test for this

pkg/rebuild/maven/strategy.go
```diff
@@ -62,6 +62,56 @@ func (b *MavenBuild) GenerateFor(t rebuild.Target, be rebuild.BuildEnv) (rebuild
 	return workflow.GenerateFor(t, be)
 }
 
+func (b *GradleBuild) ToWorkflow(t rebuild.Target) (*rebuild.WorkflowStrategy, error) {
```
algomaster99: @msuozzo btw just want to bring this to your attention that we pass target struct to `ToWorkflow` which is not what we do for Maven build tool. We need this as Gradle reads this property and appends it the jar it creates. [This also helps overriding versions from `gradle.properties` file](https://stackoverflow.com/a/40239092).

msuozzo: Ah yeah this isn't great. We might want to consider if there are other options here (like baking it into the strategy).

msuozzo: Actually, in the strategy, you should be able to do avoid the parameter by using the following template: `{{.Target.Version}}`

PR #769, opened 2025-09-10
Pr pypi stabs

gtystahl:
Adding current iteration of python stabilizers for review!

### DIFF
--- a/pkg/archive/pypi.go
+++ b/pkg/archive/pypi.go
@@ -10,13 +10,13 @@ import (
 	"encoding/base64"
 	"fmt"
 	"io"
+	"net/mail"
 	"regexp"
-	"sort"
 	"strings"
 )
 
 var AllPypiStabilizers = []Stabilizer{
-	StablePypiMetadata,
+	RemoveMetadataJSON,
 	StableWheelBuildMetadata,
 	StablePypiDescription,
 	StableCommentsCollapse,
@@ -191,35 +191,6 @@ func RemovePythonComments(pythonCode []byte) ([]byte, error) {
 	return outputBuffer.Bytes(), nil
 }
 
-// __field_list = [
-//     _make_field("Metadata-Version", required=True),
-//     _make_field("Name", required=True),
-//     _make_field("Version", required=True),
-//     _make_field("Dynamic", multiple=True),
-//     _make_field("Platform", multiple=True),
-//     _make_field("Supported-Platform", multiple=True),
-//     _make_field("Summary"),
-//     _make_field("Summary"),
-//     _make_field("Description"),
-//     _make_field("Description-Content-Type"),
-//     _make_field("Keywords"),
-//     _make_field("Home-page"),
-//     _make_field("Download-URL"),
-//     _make_field("Author"),
-//     _make_field("Author-email"),
-//     _make_field("Maintainer"),
-//     _make_field("Maintainer-email"),
-//     _make_field("License"),
-//     _make_field("Classifier", multiple=True),
-//     _make_field("Requires-Dist", multiple=True),
-//     _make_field("Requires-Python"),
-//     _make_field("Requires-External", multiple=True),
-//     _make_field("Project-URL", multiple=True),
-//     _make_field("Provides-Extra", multiple=True),
-//     # rarely used fields
-//     _make_field("Provides-Dist", multiple=True),
-//     _make_field("Obsoletes-Dist", multiple=True),
-
 type MetadataDistInfo struct {
 	// Name of the package
 	MetadataVersion        string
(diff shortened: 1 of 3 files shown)


pkg/archive/zip.go
```diff
@@ -216,14 +225,22 @@ func StabilizeZip(zr *zip.Reader, zw *zip.Writer, opts StabilizeOpts) error {
...
 		case ZipArchiveStabilizer:
+			currentStabName = s.(ZipArchiveStabilizer).Name
 			s.(ZipArchiveStabilizer).Stabilize(&mr)
 		case ZipEntryStabilizer:
+			currentStabName = s.(ZipEntryStabilizer).Name
 			for _, mf := range mr.File {
 				s.(ZipEntryStabilizer).Stabilize(mf)
 			}
 		}
+		newArchiveHash := getArchiveHash(&mr)
```
msuozzo: neat! let's separate this out into a different PR. it's a nice feature to add but we can probably review it separately

tools/ctl/ctl.go
```diff
@@ -575,7 +576,14 @@ var runBenchmark = &cobra.Command{
 			}
 			return
 		}
-		bar := pb.New(set.Count)
+
+		// Added since the count was always 0
```
msuozzo, line 580: hmm I think this just means our other benchmark generation logic neglected to set Count and we should fix it there. this can also be moved to a separate change.

pkg/archive/pypi.go
```diff
@@ -0,0 +1,748 @@
...
+	Path   string
+	Size   int64
+	SHA256 string
+}
+
+type RecordDistInfo struct {
+	entries []RecordDistEntry
+}
+
+func ParseMetadataDistInfo(r io.Reader) (*MetadataDistInfo, error) {
```
msuozzo: So it's not perfect but `net/mail.ReadMessage` should be able to parse out the headers here. It will always produce []string so if we want to assert uniqueness, we'd need to do that separately. Looks like writing it out would also need to be custom (but shouldn't be too tricky).

pkg/archive/pypi.go
```diff
@@ -0,0 +1,748 @@
...
+	Name: "pypi-metadata",
+	Func: func(zr *MutableZipReader) {
+		for _, zf := range zr.File {
+			// Only process metadata.json files
+			if !strings.HasSuffix(zf.Name, "metadata.json") {
+				continue
+			}
+			println("Processing file:", zf.Name)
+
+			zf.SetContent([]byte("This needed to change (metadata)"))
```
msuozzo, line 366: what do we want here? I remember this was a source of skew (maybe tools aren't generating these anymore?) but we should reconcile the impl and the stabilizer name: either call this "RemoveMetadataJSON" or keep the file around and apply some stabilizer to it.

pkg/archive/pypi.go
```diff
@@ -0,0 +1,748 @@
...
+
+		// Open the existing METADATA file
+		r, err := zf.Open()
+		if err != nil {
+			println("Error opening METADATA file:", err)
+			return
+		}
+
+		// Parse the METADATA file
+		manifest, err := ParseMetadataDistInfo(r)
```
msuozzo, line 407: hmm so actually per my comment above, you may not even need the structured golang object: You could just extract the 3-4 keys below that we special-case, and then unpack the rest naively from the mail.Header map.

pkg/archive/pypi.go
```diff
@@ -0,0 +1,748 @@
...
+					SHA256: sha256sum,
+				})
+
+			}
+
+		}
+		// Replace the RECORD file with the new computed entries
+		for _, zf := range zr.File {
+			if strings.HasSuffix(zf.Name, "RECORD") {
+				var buf strings.Builder
```
msuozzo, line 489: hmm. it looks as though this will only add a record entry of the RECORD file. are these the desired semantics? did we want to go through the code to regenerate all the hashes?

pkg/archive/pypi.go
```diff
@@ -0,0 +1,748 @@
...
+				if err != nil {
+					println("Error reading Python file content:", err)
+					continue
+				}
+
+				// Check if the content contains signs of auto-generated docstrings
+				if mayContainGeneratedDocstring(string(originalContent)) {
+					println("File likely contains auto-generated docstrings, applying stabilizer:", zf.Name)
+
+					cleanedContent, err := RemovePythonComments(originalContent)
```
msuozzo, line 540: alternatively, we could be a bit more conservative and just remove the aspects of the generated docstrings that we see create issues. if it's isolated to the args and returns sections, that would really reduce the blast radius of this stabilizer (and, consequently, its risk).

pkg/archive/pypi.go
```diff
@@ -0,0 +1,748 @@
...
+			println("Error opening file:", err)
+			return
+		}
+		data, err := io.ReadAll(r)
+		if err != nil {
+			println("Error reading file:", err)
+			return
+		}
+		// Replace all \r\n with \n
+		normalized := bytes.ReplaceAll(data, []byte("\r\n"), []byte("\n"))
```
msuozzo, line 578: I'd like to try the git version of this in the build script before we try to hack it in the stabilizer: `git config --global core.autocrlf true`

Let's move this to a separate change until we can see if we need it.

pkg/archive/pypi.go
```diff
@@ -0,0 +1,748 @@
...
+var StableVersionFile = ZipArchiveStabilizer{
+	Name: "version-file",
+	Func: func(zr *MutableZipReader) {
+
+		// Compute a hash of the entire archive content for debugging
+		// originalArchiveHash := getArchiveHash(zr)
+		// println("Archive SHA256 (base64):", originalArchiveHash)
+
+		// Define the pattern to find
+		patternToFind := regexp.MustCompile(`(?m)^TYPE_CHECKING = False\nif TYPE_CHECKING:\n(\s*)from typing import Tuple, Union\n(\s*)VERSION_TUPLE = Tuple\[Union\[int, str\], \.\.\.\]$`)
```
msuozzo, line 588: I'd be curious as to where this is used and what set of issues this is meant to address. If there's a specific library generating this, we should document the version that produces these differing formats.

pkg/archive/pypi.go
```diff
@@ -0,0 +1,748 @@
...
+		normalized := bytes.ReplaceAll(data, []byte("\r\n"), []byte("\n"))
+		zf.SetContent(normalized)
+	},
+}
+
+var StableVersionFile = ZipArchiveStabilizer{
+	Name: "version-file",
+	Func: func(zr *MutableZipReader) {
+
+		// Compute a hash of the entire archive content for debugging
```
msuozzo: commented code. here and below

pkg/archive/pypi.go
```diff
@@ -0,0 +1,748 @@
...
+}
+
+var StableVersionFile2 = ZipArchiveStabilizer{
+	Name: "version-file-2",
+	Func: func(zr *MutableZipReader) {
+
+		// originalArchiveHash := getArchiveHash(zr)
+		for _, zf := range zr.File {
+			// Only process METADATA files
+			if !strings.HasSuffix(zf.Name, "_version.py") {
```
msuozzo, line 639: not sure this is a safe change. I'd be curious to know how effective/necessary this is to improve outcomes. in addition to being quite blunt, we also over-match paths here that might have text before the underscore e.g. "foo_version.py"

pkg/archive/zip.go
giacomobenedetti: Sounds good! I'm moved it to a separate PR.

pkg/archive/pypi.go
giacomobenedetti: I just renamed the stabilizer. Maybe we can think something to stabilize the file instead of removing. I can investigate on how to stabilize instead of removing it

pkg/archive/pypi.go
giacomobenedetti: You're right. This was very experimental. We definitely need a different approach here

PR #770, opened 2025-09-10
Add new recent maven benchmark

msuozzo:


wbxyz (approved):

algomaster99:
> `com.google.code.findbugs:jsr305:1.3.9`

This is from [2009](https://mvnrepository.com/artifact/com.google.code.findbugs/jsr305). Also, we need the latest release only because a lot of them have multiple versions.

PR #774, opened 2025-09-11
Add timewarp support for sparse cargo registries

msuozzo:


internal/timewarp/timewarp.go
```diff
@@ -335,3 +352,36 @@ func timeWarpPyPIProjectRequest(client httpx.BasicClient, obj map[string]any, at
...
+
+// isHexString checks if string contains only hex characters
+func isHexString(s string) bool {
+	for _, c := range s {
+		if !((c >= '0' && c <= '9') || (c >= 'a' && c <= 'f') || (c >= 'A' && c <= 'F')) {
+			return false
+		}
+	}
+	return true
+}
```
wbxyz: can this be a regex? Might be easier to read as [0-9a-fA-F]

wbxyz: You might be able to roll the length check into regex as well, maybe something like ^[0-9a-fA-F]{7-40}$?

PR #776, opened 2025-09-12
Tracing which stabilizer is taking effect

giacomobenedetti:
This version just prints out when an archive has a different hash after applying a stabilizer. Depending on how we want to deal with this info, we can think about a different output

### DIFF
--- a/pkg/archive/zip.go
+++ b/pkg/archive/zip.go
@@ -7,9 +7,9 @@ import (
 	"archive/zip"
 	"bytes"
 	"crypto/sha256"
-	"encoding/base64"
 	"encoding/hex"
 	"io"
+	"log"
 	"slices"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


pkg/archive/zip.go
```diff
@@ -216,14 +217,22 @@ func StabilizeZip(zr *zip.Reader, zw *zip.Writer, opts StabilizeOpts) error {
...
 			s.(ZipArchiveStabilizer).Stabilize(&mr)
 		case ZipEntryStabilizer:
+			currentStabName = s.(ZipEntryStabilizer).Name
 			for _, mf := range mr.File {
 				s.(ZipEntryStabilizer).Stabilize(mf)
 			}
 		}
+		newArchiveHash := getArchiveHash(&mr)
+		if originalArchiveHash != newArchiveHash {
+			println("Stabilizer taking effect:", currentStabName)
```
algomaster99: Maybe use a logger? This put this in STDERR. a good advantage of this is that during inference, we output a JSON to STDOUT which is then directly decoded to strategy struct so we want to keep the STDOUT clean of such statements. Although inference is irrelevant while stabilization,  but the point is to keep STDOUT clean and have everything in STDERR

pkg/archive/zip.go
```diff
@@ -216,14 +217,22 @@ func StabilizeZip(zr *zip.Reader, zw *zip.Writer, opts StabilizeOpts) error {
 	}
 	mr := NewMutableReader(zr)
 	for _, s := range opts.Stabilizers {
+		var currentStabName string
+		originalArchiveHash := getArchiveHash(&mr)
 		switch s.(type) {
```
wbxyz: You can assign the switch output to a var to avoid casting `s` inside each case.

For example, I think this would work:
```
switch s := s.(type) {
case ZipArchiveStabilizer:
    currentStabName = s.Name
    ...
```

https://go.dev/tour/methods/16

pkg/archive/zip.go
wbxyz: Rather than printing these types, what do you think of collecting them and returning somehow? Eventually we can include the stabilizers used during comparison as part of the attestation.

pkg/archive/zip.go
```diff
@@ -252,3 +261,27 @@ func toZipCompatibleReader(r io.Reader) (io.ReaderAt, int64, error) {
...
+			continue
+		}
+		data, err := io.ReadAll(content)
+		if err != nil {
+			continue
+		}
+		buf.Write(data)
+	}
+	return computeSHA256Base64(buf.Bytes())
+}
```
wbxyz: I think we could maybe re-use the mr.WriteTo func for this:

```
h1 := crypto.SHA256.New()
err := mr.WriteTo(zip.NewWriter(h))
if err != nil {
...
}
hashBytes := h1.Sum(nil)
```

Also, it's probably ok to leave the hash as bytes and do bytes.Equal() to compare.

pkg/archive/zip.go
giacomobenedetti: I agree, collecting them from stderr is not great. Do you have any suggestion?

wbxyz: I think they could be appended to a slice and that slice could be returned by this function

PR #778, opened 2025-09-12
Test Cargo.lock inference

msuozzo:


pkg/rebuild/cratesio/infer_test.go
```diff
@@ -294,6 +294,47 @@ func TestInferStrategy(t *testing.T) {
...
+			wantFn: func(repo *gitxtest.Repository) rebuild.Strategy {
+				return &CratesIOCargoPackage{
+					Location: rebuild.Location{
+						Repo: "https://github.com/serde-rs/serde",
+						Ref:  repo.Commits["version-bump"].String(),
+						Dir:  ".",
+					},
+					RustVersion: "1.67.0",
+					ExplicitLockfile: &ExplicitLockfile{
+						LockfileBase64: "IyBUaGlzIGZpbGUgaXMgYXV0b21hdGljYWxseSBAZ2VuZXJhdGVkIGJ5IENhcmdvLgojIEl0IGlzIG5vdCBpbnRlbmRlZCBmb3IgbWFudWFsIGVkaXRpbmcuCnZlcnNpb24gPSAzCg==",
```
wbxyz, line 333: nit: Can you call base64 encode here?

msuozzo: i think it's a tradeoff. i'm not sure it brings too much extra clarity in this case.

PR #779, opened 2025-09-12
Add service to host crates registry queries

msuozzo:


cmd/registry/main.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/go-git/go-billy/v5/osfs"
+	"github.com/google/oss-rebuild/internal/api"
+	"github.com/google/oss-rebuild/internal/api/cratesregistryservice"
+	"github.com/google/oss-rebuild/pkg/registry/cratesio/index"
+	"github.com/pkg/errors"
+)
+
+var (
+	cacheDir                  = flag.String("cache-dir", "/tmp/registry-cache", "Directory to cache registry indices")
+	maxSnapshots              = flag.Int("max-snapshots", 18, "Maximum number of snapshot indices to cache")
```
wbxyz: nit: Where did 18 come from as the default? Based on approximate size of each?

deployments/main.tf
```diff
@@ -752,6 +767,41 @@ resource "google_cloud_run_v2_service" "inference" {
...
+  name     = "crates-registry"
+  location = "us-central1"
+  template {
+    service_account = google_service_account.crates-registry.email
+    timeout         = "${10 * 60}s" // 10 minutes
+    containers {
+      image = data.google_artifact_registry_docker_image.crates-registry.self_link
+      args = [
+        "--cache-dir=/cache",
+        "--max-snapshots=16",
```
wbxyz, line 788: nit: Is this based on amount of memory and/or volume size? Probably worth a comment.

internal/api/cratesregistryservice/findcommit.go
```diff
@@ -0,0 +1,146 @@
...
+	"github.com/go-git/go-git/v5"
+	"github.com/google/oss-rebuild/internal/api"
+	"github.com/google/oss-rebuild/pkg/registry/cratesio/cargolock"
+	"github.com/google/oss-rebuild/pkg/registry/cratesio/index"
+	"github.com/pkg/errors"
+	"google.golang.org/grpc/codes"
+)
+
+// FindRegistryCommitRequest represents a request to find a registry commit
+type FindRegistryCommitRequest struct {
```
wbxyz, line 21: Is there a reason this request and response type is defined in the service implementation library rather than schema?

cmd/registry/main.go
msuozzo: i've lowered this to 4 and commented in the terraform why we use 16

deployments/main.tf
msuozzo: noted in the memory volume below

PR #780, opened 2025-09-12
Query crates index from inference

msuozzo:


internal/api/inferenceservice/infer.go
```diff
@@ -53,8 +54,9 @@ func doInfer(ctx context.Context, rebuilder rebuild.Rebuilder, t rebuild.Target,
 }
 
 type InferDeps struct {
-	HTTPClient httpx.BasicClient
-	GitCache   *gitx.Cache
+	HTTPClient         httpx.BasicClient
+	GitCache           *gitx.Cache
+	CratesRegistryStub api.StubT[cratesregistryservice.FindRegistryCommitRequest, cratesregistryservice.FindRegistryCommitResponse]
```
wbxyz, line 52: Can you add stubs for this elsewhere where Infer is used?

pkg/rebuild/cratesio/infer.go
```diff
@@ -231,13 +242,27 @@ func (Rebuilder) InferStrategy(ctx context.Context, t rebuild.Target, mux rebuil
...
+		stub, err := getRegistryStub(ctx)
 		if err != nil {
-			return nil, errors.New("rust version heuristic failed")
+			return nil, errors.Wrapf(err, "[INTERNAL] Failed to access registry query stub")
+		}
+		req := cratesregistryservice.FindRegistryCommitRequest{
+			LockfileBase64: base64.StdEncoding.EncodeToString(lockContent),
+			PublishedTime:  vmeta.Updated.Format(time.RFC3339),
+		}
+		resp, err := stub(ctx, req)
```
wbxyz: nit: Inline?

internal/api/inferenceservice/infer.go
msuozzo: will address the new agent call path in a follow-on since we need to add a dep to that job

PR #787, opened 2025-09-15
Move inference storage and fs to deps

wbxyz:
This allows the caller to optionally re-use or pre-populate the
inference's git repo.

### DIFF
--- a/internal/api/inferenceservice/infer.go
+++ b/internal/api/inferenceservice/infer.go
@@ -7,8 +7,8 @@ import (
 	"context"
 	"log"
 
-	"github.com/go-git/go-billy/v5/memfs"
-	"github.com/go-git/go-git/v5/storage/memory"
+	"github.com/go-git/go-billy/v5"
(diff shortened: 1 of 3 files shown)


internal/api/inferenceservice/infer.go
```diff
@@ -51,12 +49,20 @@ func doInfer(ctx context.Context, rebuilder rebuild.Rebuilder, t rebuild.Target,
 type InferDeps struct {
 	HTTPClient httpx.BasicClient
 	GitCache   *gitx.Cache
+	Storage    storage.Storer
```
msuozzo: nit: so i think we should model these Deps structs as per-server, not per-instance. in this case, i think we'd want this to be a callback that returns the repo options. that way, each request would receive a different instance of the pair (even though the current local stub impl could just hard-code a single storer and fs for each request).

wbxyz: Changed to factory functions.

PR #789, opened 2025-09-15
Infer different Java build tools

algomaster99:
Helps in segregation of inference failures due to build tool.

### DIFF
--- a/pkg/rebuild/maven/infer.go
+++ b/pkg/rebuild/maven/infer.go
@@ -26,8 +26,14 @@ import (
 )
 
 const (
-	mavenBuildTool  = "maven"
-	gradleBuildTool = "gradle"
+	mavenBuildTool     = "maven"
+	gradleBuildTool    = "gradle"
+	sbtBuildTool       = "sbt"
+	antBuildTool       = "ant"
+	ivyBuildTool       = "ivy"
+	leiningenBuildTool = "leiningen"
+	npmBuildTool       = "npm"
+	millBuildTool      = "mill"
 )
 
 const fallbackJDK = "11"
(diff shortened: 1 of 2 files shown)


pkg/rebuild/maven/infer.go
```diff
@@ -74,8 +80,10 @@ func (Rebuilder) InferStrategy(ctx context.Context, t rebuild.Target, mux rebuil
 		return MavenInfer(ctx, t, mux, repoConfig)
 	case gradleBuildTool:
 		return GradleInfer(ctx, t, mux, repoConfig)
+	case sbtBuildTool, antBuildTool, ivyBuildTool, leiningenBuildTool, npmBuildTool, millBuildTool:
+		return nil, errors.Errorf("build tool %s is recognized but not yet supported", buildTool)
 	default:
-		return nil, errors.Errorf("unsupported build tool: %s", buildTool)
+		return nil, errors.New("unsupported build tool")
```
msuozzo: i think this should be "no recognized build tool"

pkg/rebuild/maven/infer_test.go
```diff
@@ -238,6 +238,114 @@ func TestBuildToolInference(t *testing.T) {
...
+			repo: `
+            commits:
+              - id: initial-commit
+                files:
+                  project.clj: |
+                      (defproject example "0.1.0"
+                        :description "An example Clojure project"
+                        :dependencies [[org.clojure/clojure "1.10.3"]])`,
+			expectedBuildTool: leiningenBuildTool,
+			wantErr:           false,
```
msuozzo: omit these since it's implicitly false

algomaster99: It is `true` in two test cases above.

algomaster99: Done!

PR #790, opened 2025-09-16
Add support for installing system-wide gradle

algomaster99:
@msuozzo are the changes good or would you want to incorporate checking if we need to install gradle during inference?

### DIFF
--- a/pkg/rebuild/maven/gradleinfer.go
+++ b/pkg/rebuild/maven/gradleinfer.go
@@ -18,6 +18,10 @@ import (
 	"github.com/pkg/errors"
 )
 
+// We use Gradle 8.14.3 as it supports running JDK versions 8-24.
+// Reference: https://docs.gradle.org/current/userguide/compatibility.html
+const gradleVersion = "8.14.3"
+
 func GradleInfer(ctx context.Context, t rebuild.Target, mux rebuild.RegistryMux, repoConfig *rebuild.RepoConfig) (rebuild.Strategy, error) {
 	tagGuess, err := rebuild.FindTagMatch(t.Package, t.Version, repoConfig.Repository)
 	if err != nil {
@@ -51,15 +55,37 @@ func GradleInfer(ctx context.Context, t rebuild.Target, mux rebuild.RegistryMux,
 	if err != nil {
 		return nil, errors.Wrap(err, "fetching JDK")
 	}
-
+	// Check if gradlew is present in the commit
+	var systemGradle string
+	hasGradleWrapper, err := isGradleWrapperPresent(commitObject)
+	if err != nil {
+		return nil, errors.Wrap(err, "checking for gradle wrapper")
+	}
+	if !hasGradleWrapper {
+		systemGradle = gradleVersion
+		log.Printf("Gradle wrapper (gradlew) not found in the repository. Will use system-installed Gradle.")
+	}
 	return &GradleBuild{
 		Location: rebuild.Location{
 			Repo: repoConfig.URI,
 			Dir:  buildGradleDir,
 			Ref:  ref,
 		},
-		JDKVersion: jdk,
+		JDKVersion:   jdk,
+		SystemGradle: systemGradle,
 	}, nil
+}
+
+func isGradleWrapperPresent(commit *object.Commit) (bool, error) {
+	tree, err := commit.Tree()
+	if err != nil {
+		return false, errors.Wrap(err, "getting tree")
+	}
+	_, err = tree.FindEntry("gradlew")
+	if err != nil {
+		return false, nil
+	}
+	return true, nil
 }
 
 func findBuildGradleDir(commit *object.Commit, pkg string) (string, error) {
--- a/pkg/rebuild/maven/strategy.go
+++ b/pkg/rebuild/maven/strategy.go
@@ -67,38 +67,52 @@ func (b *GradleBuild) ToWorkflow() (*rebuild.WorkflowStrategy, error) {
 		return nil, errors.Errorf("no download URL for JDK version %s", b.JDKVersion)
 	}
 
+	deps := []flow.Step{{
+		Uses: "maven/setup-java",
+		With: map[string]string{
+			"versionURL": jdkVersionURL,
+		},
+	}}
+	if b.SystemGradle != "" {
+		deps = append(deps, flow.Step{
+			Uses: "maven/setup-gradle",
+			With: map[string]string{
+				"version": gradleVersion,
+			},
+		})
+	}
+	build := []flow.Step{{
+		Uses: "maven/export-java",
+	}}
+	if b.SystemGradle != "" {
+		build = append(build, flow.Step{
+			Uses: "maven/export-gradle",
+		})
+		build = append(build, flow.Step{
+			Uses: "maven/gradle-build",
+		})
+	} else {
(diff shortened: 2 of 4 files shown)


pkg/rebuild/maven/strategy.go
```diff
@@ -84,10 +84,18 @@ func (b *GradleBuild) ToWorkflow() (*rebuild.WorkflowStrategy, error) {
 				Uses: "maven/export-java",
 			},
 			{
-				// We assume the project uses the Gradle Wrapper (gradlew).
+				Uses: "maven/setup-gradle",
```
msuozzo: i think this should go in Deps, no?

pkg/rebuild/maven/strategy.go
```diff
@@ -137,4 +145,20 @@ var toolkit = []*flow.Tool{
...
+			{
+				Runs: textwrap.Dedent(`
+				if [ ! -f gradlew ]; then
+					wget -q -O gradle-8.14.3-bin.zip https://services.gradle.org/distributions/gradle-8.14.3-bin.zip
+					unzip -q gradle-8.14.3-bin.zip -d /opt/
+					rm gradle-8.14.3-bin.zip
+					export GRADLE_HOME=/opt/gradle-8.14.3
+					export PATH=$GRADLE_HOME/bin:$PATH
+				fi`[1:]),
+				Needs: []string{"zip", "wget"},
```
msuozzo: both of these are in busybox so you can omit them for now.

msuozzo (approved):

pkg/rebuild/maven/strategy.go
```diff
@@ -137,4 +145,20 @@ var toolkit = []*flow.Tool{
 				export PATH=$JAVA_HOME/bin:$PATH`[1:]),
 		}},
 	},
+	{
+		Name: "maven/setup-gradle",
+		Steps: []flow.Step{
+			{
+				Runs: textwrap.Dedent(`
+				if [ ! -f gradlew ]; then
+					wget -q -O gradle-8.14.3-bin.zip https://services.gradle.org/distributions/gradle-8.14.3-bin.zip
```
msuozzo: why do we use this fixed version of gradle in particular?

algomaster99: Forgot to specify this, but based on the [compatibility matrix](https://docs.gradle.org/current/userguide/compatibility.html), 8.14.* seems to accommodate most JDK version.

pkg/rebuild/maven/strategy.go
algomaster99: It fails to find `unzip` if I omit it. So we need `zip`.
Also, `wget` is not present in trixie-slim. It is there because it is a requirement of `setup-java`.

algomaster99:
ping for merge @msuozzo

algomaster99:
@msuozzo forgot to ping for merge after the changes.

pkg/rebuild/maven/strategy.go
```diff
@@ -137,4 +156,28 @@ var toolkit = []*flow.Tool{
...
 	},
+	{
+		Name: "maven/export-gradle",
+		Steps: []flow.Step{{
+			Runs: textwrap.Dedent(`
+				export GRADLE_HOME=/opt/gradle
+				export PATH=$GRADLE_HOME/bin:$PATH`[1:]),
+		}},
+	},
+	// We use Gradle 8.14.3 as it is valid for building with JDK versions 8-24.
```
msuozzo: want to move this comment up to the version definition?

pkg/rebuild/maven/strategy.go
```diff
@@ -61,6 +61,8 @@ func (b *MavenBuild) GenerateFor(t rebuild.Target, be rebuild.BuildEnv) (rebuild
 	return workflow.GenerateFor(t, be)
 }
 
+const gradleVersion = "8.14.3"
```
msuozzo: we could consider moving the version hard-coding into the inference itself. that might be nice since rerunning the inference from the same code location would produce the same strategy+script whereas if we bump the version here, the same strategy at the same code location would generate a different script.

In fact, since we can detect when the gradlew is absent in the inference itself, could we go one step further and move this to a property like "ManualGradleVersion" which would conditionally add the deps and build steps to do the build? Another benefit of that approach would be that we would avoid runtime conditionals e.g. "if [ -f gradlew ]"

algomaster99: I thought of this before but I did not want to have `GradleBuild` and `GradlewBuild` but perhaps I can have a property like you suggested.

algomaster99: I added a property `NeedsSystemGradle` as I wanted the default behaviour to always use `./gradlew`.

algomaster99: We have inference like this now
```
{
  "gradle_build": {
    "repo": "https://github.com/tink-crypto/tink-java",
    "ref": "52a599de062bb551c1bf2b6f3e44709e243618be",
    "dir": ".",
    "jdk_version": "11",
    "system_gradle": "8.14.3"
  }
}
```

pkg/rebuild/maven/gradleinfer.go
```diff
@@ -51,17 +55,48 @@ func GradleInfer(ctx context.Context, t rebuild.Target, mux rebuild.RegistryMux,
...
 			Dir:  buildGradleDir,
 			Ref:  ref,
 		},
-		JDKVersion: jdk,
+		JDKVersion:   jdk,
+		SystemGradle: systemGradle,
 	}, nil
 }
 
+func isGradleWrapperPresent(commit *object.Commit) (bool, error) {
```
msuozzo, line 79: We should be able to simplify this to just check the one path we're interested in with Tree.File(). No need to loop over the whole Tree.

pkg/rebuild/maven/strategy.go
```diff
@@ -67,38 +67,55 @@ func (b *GradleBuild) ToWorkflow() (*rebuild.WorkflowStrategy, error) {
...
 		OutputDir: path.Join(b.Dir, "build", "libs"),
 	}, nil
 }
 
 type GradleBuild struct {
 	rebuild.Location
-
 	// JDKVersion is the version of the JDK to use for the build.
 	JDKVersion string `json:"jdk_version" yaml:"jdk_version"`
+	// SystemGradle indicates the version of Gradle to install if the project does not include the Gradle wrapper (gradlew).
```
msuozzo: Eh I'd generalize this statement to be "to install instead of using the Gradle wrapper (gradlew)."

pkg/rebuild/maven/strategy.go
msuozzo: nice lg

pkg/rebuild/maven/gradleinfer.go
```diff
@@ -51,17 +55,38 @@ func GradleInfer(ctx context.Context, t rebuild.Target, mux rebuild.RegistryMux,
...
+	tree, err := commit.Tree()
+	if err != nil {
+		return false, errors.Wrap(err, "getting tree")
+	}
+	if _, err := tree.File("gradlew"); err == nil {
```
msuozzo: hmm i think there should be an explicit "file not found" error to check here instead of capturing all errors.

msuozzo (approved):

PR #794, opened 2025-09-17
Use source jar to approximate commit

algomaster99:
This new heuristic goes through the log history and compares `SHA256` hash of source files in upstream `-sources.jar` and the repository. We go through the log (ordered by time) and then compare hash of each file in a commit with the hash in sources jar. We select for the local maxima in the log; this is to avoid searching through the entire log history. For example, first commit returns 5 matching files; second commit 7; third commit 6. We return the second commit immediately. I have also added some test cases which reflect this behavior.

Todo:
- [x] Waiting for the experiments to see how this performs. See what strategy is the best out of `ExhaustiveTreeSearchStrategy` and `DynamicTreeSearchStrategy`. (I just went with `DynamicTreeSearchStrategy`)
- [x] In Gradle, the artifact path is always inferred based on the head commit (defaults to `.`). Maybe we should also use this heuristic for it. #797 
- [x] Separate PR for indexscan refactor #795

### DIFF
--- a/pkg/rebuild/maven/infer.go
+++ b/pkg/rebuild/maven/infer.go
@@ -15,12 +15,14 @@ import (
 	"strings"
 
 	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/go-git/go-git/v5/plumbing/object"
 	"github.com/go-git/go-git/v5/plumbing/transport"
 	"github.com/google/oss-rebuild/internal/gitx"
 	"github.com/google/oss-rebuild/internal/uri"
 	"github.com/google/oss-rebuild/pkg/rebuild/rebuild"
 	"github.com/google/oss-rebuild/pkg/registry/maven"
+	"github.com/google/oss-rebuild/pkg/vcs/gitscan"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/rebuild/maven/infer.go
msuozzo, line 1: A more efficient version of this is already implemented in tools/indexscan. Please use that.

algomaster99: Oh I had no idea. I will look into it.

pkg/rebuild/maven/infer.go
```diff
@@ -254,3 +255,46 @@ func getClassFileMajorVersion(classBytes []byte) (int, error) {
...
+	}
+	log.Printf("commits (%d): %v", len(closest), closest)
+	log.Printf("matched %d/%d files using git index scan", matched, total)
+	if len(closest) == 0 {
+		// No matches found so we will not use the source jar heuristic, but this is not an error.
+		// The caller should try other heuristics.
+		log.Printf("no matching commit found using source jar heuristic")
+		return nil, nil
+	}
+	// TODO: further resolve this by overlapping it with time heuristic
```
msuozzo: nit: "use a better heuristic here like using commit time"

pkg/rebuild/maven/infer_test.go
```diff
@@ -364,3 +364,150 @@ func must[T any](t T, err error) T {
...
+                files:
+                  src/main/java/com/example/App.java: |
+                    b`,
+			zipEntries: []*archive.ZipEntry{
+				{
+					FileHeader: &zip.FileHeader{Name: "com/example/App.java"},
+					Body:       []byte("b"),
+				},
+			},
+			expectedCommitMessage: "second-commit",
```
msuozzo: nit: you should be able to check against the commit ID directly (using the RepoMap) and remove it here.

msuozzo (approved):

pkg/rebuild/maven/infer_test.go
algomaster99: I resolved this, but if the test fails, we will not get good messages as commit hashes are compared. The ID in YAML is consumed internally to create keys of the map so I cannot retrieve it, but I can retrieve the commit hash based on it (which I have done).
```go
t.Errorf("sourceJarGuess() = %v, want %v", sourceCommit.Hash, repo.Commits[tc.expectedCommitID])
```

PR #795, opened 2025-09-18
Move git index search to internal

algomaster99:
To use it for implementing source heuristic #794 .

### DIFF
--- a/internal/indexsearch/indexsearch.go
+++ b/internal/indexsearch/indexsearch.go
@@ -1,0 +1,315 @@
+// Copyright 2025 Google LLC
+// SPDX-License-Identifier: Apache-2.0
+
+package indexsearch
+
+import (
+	"archive/zip"
+	"context"
+	"io"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/go-git/go-billy/v5/osfs"
+	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/go-git/go-git/v5/plumbing/cache"
+	"github.com/go-git/go-git/v5/plumbing/filemode"
+	"github.com/go-git/go-git/v5/plumbing/object"
+	"github.com/go-git/go-git/v5/storage"
+	"github.com/go-git/go-git/v5/storage/filesystem"
+	"github.com/google/oss-rebuild/internal/bitmap"
+	"github.com/pkg/errors"
+)
+
+func GetRepo(ctx context.Context, uri, path string) (*git.Repository, error) {
+	var s storage.Storer
+	if path != "" {
+		gitfs, err := osfs.New(path).Chroot(".git")
+		if err != nil {
+			return nil, errors.Wrap(err, "creating git chroot")
+		}
+		s = filesystem.NewStorageWithOptions(gitfs, cache.NewObjectLRUDefault(), filesystem.Options{ExclusiveAccess: true})
+	} else {
+		base, err := os.MkdirTemp("", "oss-rebuild")
+		if err != nil {
+			return nil, errors.Wrap(err, "creating temp dir")
+		}
+		if err := os.Mkdir(filepath.Join(base, ".git"), os.ModePerm); err != nil {
+			return nil, errors.Wrap(err, "creating temp dir")
+		}
+		gitfs, err := osfs.New(base).Chroot(".git")
+		if err != nil {
+			return nil, errors.Wrap(err, "creating git chroot")
+		}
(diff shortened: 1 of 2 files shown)


internal/indexsearch/indexsearch.go
```diff
@@ -0,0 +1,315 @@
...
+	"github.com/go-git/go-git/v5/plumbing/cache"
+	"github.com/go-git/go-git/v5/plumbing/filemode"
+	"github.com/go-git/go-git/v5/plumbing/object"
+	"github.com/go-git/go-git/v5/storage"
+	"github.com/go-git/go-git/v5/storage/filesystem"
+	"github.com/google/oss-rebuild/internal/bitmap"
+	"github.com/pkg/errors"
+)
+
+func GetRepo(ctx context.Context, uri, path string) (*git.Repository, error) {
```
msuozzo: this shouldn't be part of this library. let's keep this in indexscan

internal/indexsearch/indexsearch.go
```diff
@@ -0,0 +1,315 @@
...
+	}
+	if path != "" {
+		r, err := git.Open(s, nil)
+		return r, errors.Wrap(err, "opening repo")
+	}
+	r, err := git.CloneContext(ctx, s, nil, &git.CloneOptions{URL: uri, NoCheckout: true})
+	return r, errors.Wrap(err, "cloning repo")
+}
+
+func ZipHashes(zr *zip.Reader) (files []string, err error) {
```
msuozzo: "GitHashesFromZip" and you might also change the return type to []plumbing.Hash values (so before the String() call).

internal/indexsearch/indexsearch.go
```diff
@@ -0,0 +1,315 @@
...
+		}
+		if err = f.Close(); err != nil {
+			return nil, err
+		}
+		files = append(files, h.Sum().String())
+	}
+	return files, nil
+}
+
+// DynamicTreeSearchStrategy  searches all TreeObjects in the repository for
```
msuozzo: fix space

internal/indexsearch/indexsearch.go
```diff
@@ -0,0 +1,315 @@
...
+// the number of matches against the input files provided.
+type DynamicTreeSearchStrategy struct {
```
msuozzo: we should avoid the word Strategy since it's overloaded in the project.

internal/indexsearch/indexsearch.go
```diff
@@ -0,0 +1,315 @@
...
+			}
+		}
+	}
+	cache[t.Hash] = count
+	return
+}
+
+// CommitsNearPublishStrategy searches across repository for the input files
+// provided for the nearest matching commit(s).
+type CommitsNearPublishStrategy struct {
```
msuozzo: same here. let's try to use something besides "Strategy" in this name

internal/indexsearch/indexsearch.go
```diff
@@ -0,0 +1,315 @@
...
+	it, err := r.CommitObjects()
+	if err != nil {
+		return
+	}
+	found := make(map[string]bool)
+	notNearPublishDate := func(c *object.Commit) bool {
+		committed := c.Committer.When
+		return committed.Before(s.Published.Add(-1*s.Window)) || committed.After(s.Published.Add(s.Window))
+	}
+	err = it.ForEach(func(c *object.Commit) error {
```
msuozzo: not sure which would be best but it's possible this is slower than using Log with the date bounds. i suppose we could leave this as-is and maybe add a TODO to consider the other approach.

msuozzo (commented):
should we move this to like gitx? it isn't exactly a direct extension to git but "indexsearch" isn't a great package name.

internal/indexsearch/indexsearch.go
algomaster99: Came up with these names:
```
	case "dynamic":
		s = &gitx.GetCommitMatchesViaCache{}
	case "commits-near-publish":
		s = &gitx.CommitsNearPublish{Published: published, Window: 7 * 24 * time.Hour}
	case "dynamic-exhaustive":
		s = &gitx.GetCommitMatchesViaBlob{}
```

internal/gitx/indexsearch.go
msuozzo, line 1: sorry for the runaround. let's add this to pkg/. what about pkg/vcs/gitscan/?

then we get to expose the functionality, keep the "git" name at the callsite, and group this and related functionality into a new "vcs" top-level

msuozzo (approved):

PR #797, opened 2025-09-18
Reduce reliance on HEAD

algomaster99:
Maven: Use the pom path inferred from the commit guessed by heuristic not `HEAD`.
Gradle: The `GradleBuild.Dir` is now also computed based on commit guessed by heuristic.


msuozzo (approved):
with a few tests, this is lgtm

pkg/rebuild/maven/maveninfer_test.go
```diff
@@ -257,3 +263,187 @@ func addPomArtifact(mavenRegistry *mockMavenRegistry, pom *PomXML) {
...
+						FileHeader: &zip.FileHeader{Name: "META-INF/maven/foo/bar/pom.xml"},
+						Body:       []byte("<project><groupId>foo</groupId><artifactId>bar</artifactId><version>1.0.0</version></project>"),
+					},
+					{
+						FileHeader: &zip.FileHeader{Name: "META-INF/MANIFEST.MF"},
+						Body:       []byte("Manifest-Version: 1.0\nBuild-Jdk: 11.0.1\n"),
+					},
+				},
+			},
+			expectedHeuristic: "using source jar heuristic with mismatched version",
```
msuozzo, line 349: This isn't ideal imo since we really want to be checking the program outputs, not its incidental side-effects. that said, we can revisit later

algomaster99:
I added some test cases but ideally there should be 7
1. using git log heuristic with pkg match
2. using git log heuristic with pkg and version match
1. using source jar heuristic with pkg match
2. using source jar heuristic with pkg and version match
1. using tag heuristic with pkg match
2. using tag heuristic with pkg and version match
3. error

I thought for `source` and `tag` heuristic with pkg and version match I would need to mock a detached commit, but I found [a real world example](https://github.com/package-url/packageurl-java/commit/ed9be60d7) where the commit is inferred using tag heuristic even though the commit is there in the git repository. I wonder why refMap failed here.

pkg/rebuild/maven/maveninfer_test.go
algomaster99: I completely agree, but this seemed the best way to go forward.

PR #821, opened 2025-09-22
Prevent falling through if condition is false

algomaster99:
This won't throw a nil dereferencing error if there is a `fallthrough` from git log heuristic to source jar heuristic and sourceJar guess is `nil`.
I also compute the heuristics lazily. `refMap` is only computed when tag heurisitic does not give a valid match or fails.

### DIFF
--- a/pkg/rebuild/maven/maveninfer.go
+++ b/pkg/rebuild/maven/maveninfer.go
@@ -107,18 +107,20 @@ func MavenInfer(ctx context.Context, t rebuild.Target, mux rebuild.RegistryMux, 
 		fallthrough
 	case sourceJarGuess != nil:
 		commit = sourceJarGuess
-		pomXML, foundPkgPath, err := findPomXML(commit, t.Package)
-		if err != nil {
-			log.Printf("source jar heuristic failed: could not find a pom.xml for the package")
-		} else {
-			ref = sourceJarGuess.Hash.String()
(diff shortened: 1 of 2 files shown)


pkg/rebuild/maven/maveninfer.go
```diff
@@ -37,97 +66,34 @@ func MavenInfer(ctx context.Context, t rebuild.Target, mux rebuild.RegistryMux,
...
-
-		} else if err == plumbing.ErrObjectNotFound {
-			log.Printf("git log heuristic ref not found in repo")
-		} else {
+		} else if !errors.Is(err, plumbing.ErrObjectNotFound) {
 			return nil, errors.Wrapf(err, "[INTERNAL] Failed ref resolve from git log [repo=%s,ref=%s]", repoConfig.URI, pomXMLGuess)
 		}
-		fallthrough
-	case sourceJarGuess != nil:
-		commit = sourceJarGuess
```
msuozzo, line 109: so i think i might ask that we fix the bug in-place (with a test confirming it). then, we can move forward with this refactor (which we might want to rethink across all of our ecosystem inference implementations).

algomaster99: I added a test. If you run it before the changes in source, it would throw a dereferencing error.

msuozzo (approved):

PR #824, opened 2025-09-22
GCB planner only validate necessary asset URLs

wbxyz:
We should only throw errors for empty URLs if those assets are actually
going to be generated.


pkg/build/gcb/planner.go
```diff
@@ -542,9 +542,6 @@ func (p *Planner) generateAssetUploadScript(target rebuild.Target, opts build.Pl
 		}
 		for _, assetType := range assetTypes {
 			url := opts.Resources.AssetStore.URL(assetType.For(target))
-			if url == nil {
```
msuozzo, line 545: hmm we definitely want URL to be non-nil, right? but i do see the issue here. looking at it again, i think it'd be cleaner to conditionally handle the UseNetworkProxy and syscal cases above the loop and ensure they only appear when necessary in assetTypes for processing.

PR #826, opened 2025-09-22
Add support for location hint during inference

algomaster99:


internal/api/inferenceservice/infer.go
```diff
@@ -19,8 +19,8 @@ import (
 
 func doInfer(ctx context.Context, rebuilder rebuild.Rebuilder, t rebuild.Target, mux rebuild.RegistryMux, hint rebuild.Strategy, ropt *gitx.RepositoryOptions) (rebuild.Strategy, error) {
 	var repo string
-	if lh, ok := hint.(*rebuild.LocationHint); ok && lh != nil {
-		repo = lh.Location.Repo
+	if repoHint := hint.(*rebuild.LocationHint).Repo; repoHint != "" {
```
msuozzo: Is this safe? Wouldn't the receiver be nil and this access fail?

algomaster99: I always initialize it to the following. Then if repoHint is present, the object below is mutated.
```
&schema.StrategyOneOf{
			LocationHint: &rebuild.LocationHint{
				Location: rebuild.Location{},
			},
		}
```
EDIT: but this might not be safe for other places. I will fix it.

msuozzo: Yeah let's not assume the inputs

tools/ctl/ctl.go
```diff
@@ -836,16 +836,25 @@ var migrate = &cobra.Command{
 }
 
 var infer = &cobra.Command{
-	Use:   "infer --ecosystem <ecosystem> --package <name> --version <version> [--artifact <name>] [--api <URI>] [--format strategy|dockerfile|debug-steps]",
+	Use:   "infer --ecosystem <ecosystem> --package <name> --version <version> [--repo-hint <repo>] [--artifact <name>] [--api <URI>] [--format strategy|dockerfile|debug-steps]",
 	Short: "Run inference",
 	Args:  cobra.NoArgs,
 	Run: func(cmd *cobra.Command, args []string) {
+		var strategyHint *schema.StrategyOneOf = &schema.StrategyOneOf{
+			LocationHint: &rebuild.LocationHint{
```
msuozzo: move this to the if-case

internal/api/inferenceservice/infer.go
msuozzo, line 1: can you add the repo != "" check to line 49 and assume that the presence of a hint implies a repo?

internal/api/inferenceservice/infer.go
```diff
@@ -19,8 +19,8 @@ import (
 
 func doInfer(ctx context.Context, rebuilder rebuild.Rebuilder, t rebuild.Target, mux rebuild.RegistryMux, hint rebuild.Strategy, ropt *gitx.RepositoryOptions) (rebuild.Strategy, error) {
 	var repo string
-	if lh, ok := hint.(*rebuild.LocationHint); ok && lh != nil {
-		repo = lh.Location.Repo
+	if lh, ok := hint.(*rebuild.LocationHint); ok && lh != nil && lh.Repo != "" {
```
msuozzo: the `lh != nil` is superfluous when `ok` is true

tools/ctl/ctl.go
algomaster99: I instead added an anonymous function that will help us extend it with other location hints later.

internal/api/inferenceservice/infer.go
algomaster99: If `hint` is nil, ok is true so we have to add `lh!=nil`.

internal/api/inferenceservice/infer.go
algomaster99: The other way round made more sense and I have also updated the error message.

internal/api/inferenceservice/infer.go
```diff
@@ -49,6 +49,9 @@ func Infer(ctx context.Context, req schema.InferenceRequest, deps *InferDeps) (*
 	if req.LocationHint() != nil && req.LocationHint().Ref == "" && req.LocationHint().Dir != "" {
 		return nil, api.AsStatus(codes.Unimplemented, errors.New("location hint dir without ref not implemented"))
 	}
+	if req.LocationHint() != nil && req.LocationHint().Repo != "" {
+		return nil, api.AsStatus(codes.InvalidArgument, errors.New("location hint without repo is not allowed"))
```
msuozzo: s/is not allowed/is not supported/g

PR #827, opened 2025-09-23
Migrate the medic to agent-api

wbxyz:


tools/medic/main.go
```diff
@@ -431,26 +439,55 @@ func main() {
...
 		log.Printf("Running rebuild for %s: %s", rec.Rebuild.ID(), strings.Join(rec.NewScript.Commands, "; "))
-		t := rec.Rebuild.Target()
-		oneof := schema.NewStrategyOneOf(rec.NewStrategy)
-		resp, err := rebuildSmoketest(ctx, schema.SmoketestRequest{
-			Ecosystem: t.Ecosystem,
-			Package:   t.Package,
-			Versions:  []string{t.Version},
-			Strategy:  &oneof,
-			ID:        runID,
+		sessionUUID, err := uuid.NewV7()
```
msuozzo, line 452: let's add a quick comment here. might also want to point to the existing impl which this copies.

PR #835, opened 2025-09-24
Add rust version heuristics from Cargo.toml

msuozzo:


pkg/rebuild/cratesio/hints.go
```diff
@@ -0,0 +1,55 @@
...
+		hi = min("1.54.0", hi)
+	}
+	if resolverTwoPattern.MatchString(cargoTomlText) {
+		hi = min("1.63.0", hi) // After which resolver 2 became default and was omitted
+		lo = max("1.51.0", lo)
+	} else if lo > "1.51.0" {
+		lo = max("1.64.0", lo)
+	} else if hi < "1.63.0" {
+		hi = min("1.50.0", hi)
+	}
```
wbxyz: It took me a second to think through these, want to add a comment along the lines of "If no twoPattern found, then move the lo and hi bounds out of the twoPattern range"

pkg/rebuild/cratesio/hints.go
```diff
@@ -0,0 +1,55 @@
...
+	prettyArrayPattern = regexp.MustCompile(`(?s)\[\s*\n\s+.*\n\s*\]`)
+	// modernHeaderPattern detects a change in the Cargo.toml header (Rust 1.55+)
+	modernHeaderPattern = regexp.MustCompile(`#.*to registry \(e\.g\., crates\.io\) dependencies\.`)
+	// docExamplesRegex detects the addition of the scape indicator (Rust 1.67+)
+	docExamplesRegex = regexp.MustCompile(`(?m)^\s*doc-scrape-examples\s*=\s*(true|false)\s*$`)
+)
+
+// detectRustVersionBounds analyzes Cargo.toml for structural patterns that indicate
+// minimum Rust version requirements based on tooling behavior changes.
+func detectRustVersionBounds(cargoTomlText string) (lo, hi string) {
```
wbxyz, line 29: Mention that empty string indicates no constraint?

pkg/rebuild/cratesio/hints.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+	if modernHeaderPattern.MatchString(cargoTomlText) {
+		lo = max("1.55.0", lo)
+	} else {
+		hi = min("1.54.0", hi)
+	}
+	if resolverTwoPattern.MatchString(cargoTomlText) {
+		hi = min("1.63.0", hi) // After which resolver 2 became default and was omitted
+		lo = max("1.51.0", lo)
+	} else if lo > "1.51.0" {
```
wbxyz: The most robust way to do this might be to use semver.Cmp(). As is, the code works correctly, but it would probably break in non-obvious ways if someone added a 3-digit minor version heuristic in the future.

I can understand wanting to stick with string comparison for now though if you're prioritizing simplicity.

pkg/rebuild/cratesio/infer.go
```diff
@@ -234,7 +234,19 @@ func (Rebuilder) InferStrategy(ctx context.Context, t rebuild.Target, mux rebuil
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "[INTERNAL] Failed to extract upstream Cargo.toml")
+	}
+	minVer, maxVer := detectRustVersionBounds(string(cargoTomlText))
+	if minVer != "" && semver.Cmp(rustVersion, minVer) < 0 {
+		rustVersion = minVer
+	}
+	if maxVer != "" && semver.Cmp(rustVersion, maxVer) > 0 {
+		rustVersion = maxVer
+	}
```
wbxyz, line 250: Is there a reason to prefer the maxVer? I suspect it's generally improved, but if there's a reason for it, that might be worth adding as a comment.

wbxyz (approved):

pkg/rebuild/cratesio/hints.go
msuozzo: eh for now i think this is tolerable. min() and max() are nice to have that simplicity here and i have a draft of a better format for this logic that would better encapsulate a use of semver. i'll leave a TODO

pkg/rebuild/cratesio/hints.go
msuozzo: i don't think we use that behavior, right? it's also in the unit tests so it's at least accessible. i'll leave that as-is

pkg/rebuild/cratesio/infer.go
msuozzo: done

pkg/rebuild/cratesio/hints.go
msuozzo: done

PR #837, opened 2025-09-24
Compute heuristics in Maven inference lazily

algomaster99:
Related: https://github.com/google/oss-rebuild/pull/821

I refactored from switch-case to if-else as it was much more readable.

### DIFF
--- a/pkg/rebuild/maven/maveninfer.go
+++ b/pkg/rebuild/maven/maveninfer.go
@@ -107,18 +107,20 @@ func MavenInfer(ctx context.Context, t rebuild.Target, mux rebuild.RegistryMux, 
 		fallthrough
 	case sourceJarGuess != nil:
 		commit = sourceJarGuess
-		pomXML, foundPkgPath, err := findPomXML(commit, t.Package)
-		if err != nil {
-			log.Printf("source jar heuristic failed: could not find a pom.xml for the package")
-		} else {
-			ref = sourceJarGuess.Hash.String()
-			dir = filepath.Dir(foundPkgPath)
-			if pomXML.Version() != version {
-				log.Printf("using source jar heuristic with mismatched version [expected=%s,actual=%s,path=%s,ref=%s]", version, pomXML.Version(), path.Join(dir, "pom.xml"), ref[:9])
+		if err == nil {
+			pomXML, foundPkgPath, err := findPomXML(commit, t.Package)
+			if err != nil {
+				log.Printf("source jar heuristic failed: could not find a pom.xml for the package")
 			} else {
-				log.Printf("using source jar heuristic (pkg and version match) ref: %s", ref[:9])
+				ref = sourceJarGuess.Hash.String()
+				dir = filepath.Dir(foundPkgPath)
+				if pomXML.Version() != version {
+					log.Printf("using source jar heuristic with mismatched version [expected=%s,actual=%s,path=%s,ref=%s]", version, pomXML.Version(), path.Join(dir, "pom.xml"), ref[:9])
+				} else {
+					log.Printf("using source jar heuristic (pkg and version match) ref: %s", ref[:9])
+				}
+				break
 			}
-			break
 		}
 		fallthrough
 	default:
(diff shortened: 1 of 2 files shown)


pkg/rebuild/maven/maveninfer.go
```diff
@@ -19,117 +19,83 @@ import (
...
+	if !found {
+		sourceJarGuess, err = findClosestCommitToSource(ctx, t, mux, repoConfig.Repository)
+		if err != nil {
+			log.Printf("source jar heuristic failed: %s", err)
 		}
-		fallthrough
-	default:
+		if sourceJarGuess != nil {
+			dir, found = validateCommit(sourceJarGuess, t, "source jar")
+			ref = sourceJarGuess.Hash.String()
```
algomaster99:
```go
        sourceJarGuess, err = findClosestCommitToSource(ctx, t, mux, repoConfig.Repository)
		if err != nil {
			log.Printf("source jar heuristic failed: %s", err)
		}
		if sourceJarGuess != nil {
			dir, found = validateCommit(sourceJarGuess, t, "source jar")
			ref = sourceJarGuess.Hash.String()
		}
```

I don't like this pattern. Will consider returning an error from source jar heuristic if no matches are found.

pkg/rebuild/maven/maveninfer.go
```diff
@@ -143,6 +111,23 @@ func MavenInfer(ctx context.Context, t rebuild.Target, mux rebuild.RegistryMux,
...
+	pomXML, foundPkgPath, err := findPomXML(commit, t.Package)
+	if err != nil {
+		return "", fmt.Sprintf("could not find a pom.xml for the package in ref %s", commit.Hash.String()[:9])
+	}
+	dir = filepath.Dir(foundPkgPath)
+	ref := commit.Hash.String()
+	if pomXML.Version() != t.Version {
+		return dir, fmt.Sprintf("with mismatched version [expected=%s,actual=%s,path=%s,ref=%s]", t.Version, pomXML.Version(), path.Join(dir, "pom.xml"), ref[:9])
+	} else {
+		return dir, fmt.Sprintf("with pkg and version match ref [version=%s,path=%s,ref=%s]", t.Version, path.Join(dir, "pom.xml"), ref[:9])
```
algomaster99, line 124: @msuozzo I have come up with this pattern to make `validateCommit` be independent of `heuristicName`. I did not want to throw error since none of them are really errors. We just need to log in all three cases so returning an error and then wrapping above did not seem the best idea.

We would want to stick with this pattern "use <> heuristic: " because it just makes organizing errors easier.

algomaster99: I fixed this logging pattern to something like this
```
2025/10/16 19:40:34 using source jar heuristic:
2025/10/16 19:40:34 with mismatched version [expected=1.4.1,actual=${io.hotmoka.annotations.version},path=io-hotmoka-annotations/pom.xml,ref=143098491]
```
Looking at the tests now.

pkg/rebuild/maven/maveninfer.go
```diff
@@ -143,6 +111,23 @@ func MavenInfer(ctx context.Context, t rebuild.Target, mux rebuild.RegistryMux,
 	}, nil
 }
 
+// validateCommit is a helper that checks if a given commit contains a valid pom.xml for the package.
+// It returns the directory containing the pom.xml and a log message.
+// If no valid pom.xml is found, it returns an empty directory and a log message indicating the failure reason.
+func validateCommit(commit *object.Commit, t rebuild.Target) (dir string, logMessage string) {
```
msuozzo: okay i can't think of a better solution for this pattern so let's rename the function to "findBuildDir" and the logMessage output to "summary" for now.

pkg/rebuild/maven/maveninfer.go
```diff
@@ -19,117 +19,85 @@ import (
...
-	case sourceJarGuess != nil:
-		commit = sourceJarGuess
-		if err == nil {
-			pomXML, foundPkgPath, err := findPomXML(commit, t.Package)
-			if err != nil {
-				log.Printf("source jar heuristic failed: could not find a pom.xml for the package")
-			} else {
+	}
+	// 3. Source Jar Heuristic
+	// Generate guess by comparing source JAR contents to repo contents.
```
msuozzo: let's cut these secondary descriptions

msuozzo (approved):

PR #838, opened 2025-09-24
Test for gradle inference

algomaster99:
I also changed the log message for Gradle to `no git ref` as in Gradle we don't verify the build.gradle like we do for pom.xml in Maven. In Maven, the artifact ID is looked for in the commit tree. In Gradle, we don't do this. It is merely used for inferring build directory.

I can probably add validation to Gradle too.

### DIFF
--- a/pkg/rebuild/maven/gradleinfer.go
+++ b/pkg/rebuild/maven/gradleinfer.go
@@ -36,7 +36,7 @@ func GradleInfer(ctx context.Context, t rebuild.Target, mux rebuild.RegistryMux,
 		ref = sourceJarGuess.Hash.String()
 		log.Printf("using source jar heuristic ref: %s", ref[:9])
 	default:
-		return nil, errors.Errorf("no valid git ref")
+		return nil, errors.Errorf("no git ref")
 	}
 	commitObject, err := repoConfig.Repository.CommitObject(plumbing.NewHash(ref))
 	if err != nil {
--- a/pkg/rebuild/maven/gradleinfer_test.go
+++ b/pkg/rebuild/maven/gradleinfer_test.go
@@ -4,9 +4,13 @@
 package maven
 
 import (
+	"archive/zip"
 	"testing"
 
 	"github.com/google/oss-rebuild/internal/gitx/gitxtest"
+	"github.com/google/oss-rebuild/pkg/archive"
+	"github.com/google/oss-rebuild/pkg/rebuild/rebuild"
+	"github.com/google/oss-rebuild/pkg/registry/maven"
 )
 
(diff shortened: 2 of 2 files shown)


pkg/rebuild/maven/gradleinfer.go
```diff
@@ -36,7 +36,7 @@ func GradleInfer(ctx context.Context, t rebuild.Target, mux rebuild.RegistryMux,
 		ref = sourceJarGuess.Hash.String()
 		log.Printf("using source jar heuristic ref: %s", ref[:9])
 	default:
-		return nil, errors.Errorf("no valid git ref")
+		return nil, errors.Errorf("no git ref (gradle)")
```
msuozzo: don't really enjoy this style. at that point you'd need all of the errors to annotate their origin. instead, you could wrap the higher-level consumer of this error with "gradle infer" or "inferring gradle" and test that interface (or just test this one and rely on the higher-level to differentiate between the two).

pkg/rebuild/maven/gradleinfer_test.go
```diff
@@ -109,3 +116,165 @@ func TestFindBuildGradleDir(t *testing.T) {
...
+					},
+					{
+						FileHeader: &zip.FileHeader{Name: "META-INF/MANIFEST.MF"},
+						Body:       []byte("Manifest-Version: 1.0\nBuild-Jdk: 11.0.1\n"),
+					},
+				},
+			},
+			// select second-commit because it has the tag matching version 1.0.0
+			expectedCommit: "second-commit",
+			wantErr:        false,
```
msuozzo: remove the false entries

pkg/rebuild/maven/gradleinfer.go
algomaster99: A better idea indeed.

Implemented here: https://github.com/google/oss-rebuild/pull/839

msuozzo (approved):

pkg/rebuild/maven/gradleinfer_test.go
```diff
@@ -109,3 +114,159 @@ func TestFindBuildGradleDir(t *testing.T) {
...
+			}
+			repoConfig.Repository = repo.Repository
+			mockRegistry := &mockMavenRegistry{
+				artifactCoordinates: make(map[artifactCoordinates][]byte),
+			}
+			addArtifacts(mockRegistry, tc.zipEntries, tc.target)
+			mockMux := rebuild.RegistryMux{
+				Maven: mockRegistry,
+			}
+			got, err := GradleInfer(context.Background(), tc.target, mockMux, repoConfig)
```
msuozzo: t.Context()

PR #842, opened 2025-09-24
Use git assertions in maven inference tests

algomaster99:


pkg/rebuild/maven/maveninfer_test.go
```diff
@@ -533,7 +531,7 @@ func TestMavenInfer(t *testing.T) {
 					},
 				},
 			},
-			expectedHeuristic: "",
+			expectedCommit: "",
```
msuozzo: the err cases can omit this empty string

msuozzo (approved):

PR #843, opened 2025-09-24
Compute gradle inference heuristics lazily

algomaster99:
Analogous to #837


pkg/rebuild/maven/gradleinfer.go
```diff
@@ -27,20 +27,20 @@ func GradleInfer(ctx context.Context, t rebuild.Target, mux rebuild.RegistryMux,
...
 		return nil, errors.Wrapf(err, "[INTERNAL] tag heuristic error")
 	}
-	sourceJarGuess, err := findClosestCommitToSource(ctx, t, mux, repoConfig.Repository)
-	if err != nil {
-		log.Printf("source jar heuristic failed: %s", err)
-	}
 	var ref string
-	switch {
-	case tagGuess != "":
+	if tagGuess != "" {
```
msuozzo, line 32: can we use the same pattern as maven? numbered comments on top of blocks that are all at the same level? the current form was used specifically to avoid nesting like this.

PR #558, opened 2025-06-10
Port to new genai library

msuozzo:

PR #848, opened 2025-09-25
Restore function calling to chat

msuozzo:
#558 accidentally removed this and hasn't been caught since :(

### DIFF
--- a/internal/llm/chat.go
+++ b/internal/llm/chat.go
@@ -82,12 +82,17 @@ func NewChat(ctx context.Context, client *genai.Client, model string, config *ge
 // by the model using the configured tools, and sends results back to the model
 // until a final content response is received or an error occurs.
 func (cm *Chat) SendMessage(ctx context.Context, parts ...*genai.Part) (*genai.GenerateContentResponse, error) {
-	// Convert []*Part to []Part
-	partsValues := make([]genai.Part, len(parts))
-	for i, part := range parts {
-		partsValues[i] = *part
+	var last *genai.Content
+	for content, err := range cm.SendMessageStream(ctx, parts...) {
+		if err != nil {
+			return nil, err
+		}
+		last = content
 	}
-	return cm.session.SendMessage(ctx, partsValues...)
+	if last != nil {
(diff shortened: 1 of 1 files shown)


wbxyz (approved):

internal/llm/chat.go
```diff
@@ -82,12 +82,13 @@ func NewChat(ctx context.Context, client *genai.Client, model string, config *ge
 // by the model using the configured tools, and sends results back to the model
 // until a final content response is received or an error occurs.
 func (cm *Chat) SendMessage(ctx context.Context, parts ...*genai.Part) (*genai.GenerateContentResponse, error) {
-	// Convert []*Part to []Part
-	partsValues := make([]genai.Part, len(parts))
-	for i, part := range parts {
-		partsValues[i] = *part
+	for content, err := range cm.SendMessageStream(ctx, parts...) {
```
wbxyz, line 85: The first content from SendMessageStream will the the original user's request, maybe this should iterate on SendMessageStream until content.Role != User?

msuozzo: ah i fixed this right when you commented :)

internal/llm/chat.go
```diff
@@ -127,8 +132,26 @@ func (cm *Chat) SendMessageStream(ctx context.Context, parts ...*genai.Part) ite
...
+					if !found {
+						yield(nil, errors.Errorf("tool implementation not found for function call '%s'", call.Name))
+						return
+					}
+					funcResponse := implFunc(call.Args)
+					responsePart := genai.Part{FunctionResponse: &funcResponse}
+					currentParts = append(currentParts, &responsePart)
+				}
+				continue
+			} else if candidate.FinishReason == genai.FinishReasonStop {
```
wbxyz, line 154: Nit: Should the check for finished go before the function calling?

msuozzo: eh in that case trying to complete the interaction is maybe the best option. from a cursory glance, i don't see any reason the client library would fail on that call so i think this is more wai.

PR #849, opened 2025-09-26
Upload inference logs

algomaster99:
It is nice to have inference logs for our own debugging and also for expanding the logic for agent workflow to infer source repository using the inference logs. We had some positive results with inferring source URL based on logs of manual inference.

I haven't configured `LogStore` in agent API and also in our Terraform configuration. I will do so once the integration of these changes look good.


algomaster99:
We have decided to move on with only error of manual inference as an input in the prompt. #860

PR #851, opened 2025-09-26
Use genai.Chat.Send where possible

msuozzo:


internal/llm/chat.go
```diff
@@ -99,23 +99,23 @@ func (cm *Chat) SendMessage(ctx context.Context, parts ...*genai.Part) (*genai.C
...
+			partsPtrs := make([]*genai.Part, len(currentParts))
 			for i, part := range currentParts {
-				partsValues[i] = *part
+				partsPtrs[i] = &part
+			}
+			if !yield(&genai.Content{Parts: partsPtrs, Role: UserRole}, nil) {
+				return
 			}
-			resp, err := cm.session.SendMessage(ctx, partsValues...)
+			resp, err := cm.session.SendMessage(ctx, currentParts...)
```
wbxyz: Is there a significant improvement here? Either way we slice it, we need both []Part (for `cm.session.SendMessage()`) and []*Part (for `genai.Content{Parts: ...}`).

Given that we always need to do a conversion, and many other interfaces use []*Part, maybe we should stick with the less-copy-intensive version?

wbxyz: Oh actually... it looks like session.SendMessage is just converting the non-pointer version *back* into pointers, then calling session.Send. Let's just use `session.Send` and keep pointers?

PR #855, opened 2025-09-29
Provide target JDK version to mvn build

algomaster99:


pkg/rebuild/maven/infer.go
```diff
@@ -160,48 +160,51 @@ func inferBuildTool(commit *object.Commit) (string, error) {
...
-	} else if JDKDownloadURLs[jdk] == "" {
-		log.Printf("%s has no associated JDK URL, falling back to JDK %s", jdk, fallbackJDK)
-		jdk = fallbackJDK
-	}
-	return jdk, nil
-}
-
-// inferJDKVersion gets the JDK version from the MANIFEST or Java bytecode.
-func inferJDKVersion(ctx context.Context, name, version string, mux rebuild.RegistryMux) (string, error) {
+func inferJDKAndTargetVersion(ctx context.Context, name, version string, mux rebuild.RegistryMux) (string, string, error) {
```
msuozzo: name return type

pkg/rebuild/maven/infer.go
```diff
@@ -160,48 +160,51 @@ func inferBuildTool(commit *object.Commit) (string, error) {
...
+	}
+	targetJDK, err := inferJDKFromBytecode(zipReader)
+	if err != nil {
+		return "", "", errors.Wrap(err, "inferring target JDK version")
 	}
+	return buildJDK, targetJDK, nil
+}
+
+// inferJDKVersion gets the JDK version from the MANIFEST or Java bytecode.
+func inferJDKVersion(zipReader *zip.Reader) (string, error) {
```
msuozzo: let's fold this into the build+target function. that way we don't need to call the bytecode check twice.

pkg/rebuild/maven/infer.go
```diff
@@ -160,48 +160,51 @@ func inferBuildTool(commit *object.Commit) (string, error) {
...
+	if JDKDownloadURLs[jdk] != "" {
 		return jdk, nil
 	}
-	jdkInt, err := inferJDKFromBytecode(zipReader)
+	jdk, err = inferJDKFromBytecode(zipReader)
 	if err != nil {
 		return "", errors.Wrap(err, "inferring JDK from bytecode")
 	}
-	return fmt.Sprintf("%d", jdkInt), nil
+	if jdk == "" {
```
msuozzo: this predicate can be combined with the next

pkg/rebuild/maven/maveninfer.go
```diff
@@ -139,7 +139,8 @@ func MavenInfer(ctx context.Context, t rebuild.Target, mux rebuild.RegistryMux,
 			Dir:  dir,
 			Ref:  ref,
 		},
-		JDKVersion: jdk,
+		JDKVersion:    buildJDK,
+		TargetVersion: targetJDK,
```
msuozzo, line 144: as noted, let's add a TODO about supporting multi-release JARs (i.e. zeroing this out when it's detected)

msuozzo (approved):

pkg/rebuild/maven/infer_test.go
```diff
@@ -31,6 +32,189 @@ func (m *mockMavenRegistry) ReleaseFile(ctx context.Context, name string, versio
...
+				{
+					FileHeader: &zip.FileHeader{Name: "com/example/Main.class"},
+					Body:       []byte{0xCA, 0xFE, 0xBA, 0xBE, 0x00, 0x00, 0x00, 0x34}, // Java 8
+				},
+			},
+			wantBuildJDK:  "11",
+			wantTargetJDK: "8",
+		},
+		{
+			name: fmt.Sprintf("build JDK (not in manifest), target JDK (higher than fallback:%s) (URL present)", fallbackJDK),
```
msuozzo: let's remove fallbackJDK being templated here in all the names. we don't have version numbers for any of the other properties.

pkg/rebuild/maven/infer_test.go
```diff
@@ -31,6 +32,189 @@ func (m *mockMavenRegistry) ReleaseFile(ctx context.Context, name string, versio
...
+					FileHeader: &zip.FileHeader{Name: "com/example/Main.class"},
+					Body:       []byte{0xCA, 0xFE, 0xBA, 0xBE, 0x00, 0x00, 0x00, 0x35}, // Java 9
+				},
+			},
+			wantBuildJDK:  "17",
+			wantTargetJDK: "9",
+		},
+		{
+			name: fmt.Sprintf("build JDK (in manifest) (URL present), target JDK (lower than fallback:%s) (URL absent)", fallbackJDK),
+			input: []*archive.ZipEntry{
```
msuozzo: since this test table really only varies the two quantities in the manifest and class header, let's move the zipentry construction to the test body and just have the version identifiers referenced here in the testcases.

algomaster99:
@msuozzo ping for another review

algomaster99:
We tried to test if this feature has any gains, but we instead get more compilation failures. To understand this, let's consider
```
/usr/bin/docker run --rm --memory 10g ctl infer --ecosystem maven --package io.lettuce:lettuce-core --version 6.3.2.RELEASE --artifact lettuce-core-6.3.2.RELEASE.jar
 STDOUT:
{
  "maven_build": {
    "repo": "https://github.com/lettuce-io/lettuce-core",
    "ref": "8941aea2cb1e3124117cecf9a06c0e254d397911",
    "dir": ".",
    "jdk_version": "11",
    "target_version": "8"
  }
}
```
We infer its build command as `mvn clean package -DskipTests --batch-mode -f . -Dmaven.javadoc.skip=true -Dmaven.compiler.release=8`. However, this fails with compilation error. 
```
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  02:08 min
[INFO] Finished at: 2025-10-01T21:14:03Z
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.12.1:compile (java-compile) on project lettuce-core: Compilation failure: Compilation failure: 
[ERROR] /src/src/main/java/io/lettuce/core/ConnectionBuilder.java:[51,15] package jdk.net does not exist
[ERROR] /src/src/main/java/io/lettuce/core/cluster/event/JfrAskRedirectionEvent.java:[18,15] package jdk.jfr does not exist
[ERROR] /src/src/main/java/io/lettuce/core/cluster/event/JfrAskRedirectionEvent.java:[19,15] package jdk.jfr does not exist
[ERROR] /src/src/main/java/io/lettuce/core/cluster/event/JfrAskRedirectionEvent.java:[20,15] package jdk.jfr does not exist
[ERROR] /src/src/main/java/io/lettuce/core/cluster/event/JfrAskRedirectionEvent.java:[21,15] package jdk.jfr does not exist
[ERROR] /src/src/main/java/io/lettuce/core/cluster/event/JfrAskRedirectionEvent.java:[32,38] cannot find symbol
[ERROR]   symbol: class Event
[ERROR] /src/src/main/java/io/lettuce/core/cluster/event/JfrAskRedirectionEvent.java:[29,2] cannot find symbol
[ERROR]   symbol: class Category
[ERROR] /src/src/main/java/io/lettuce/core/cluster/event/JfrAskRedirectionEvent.java:[30,2] cannot find symbol
[ERROR]   symbol: class Label
[ERROR] /src/src/main/java/io/lettuce/core/cluster/event/JfrAskRedirectionEvent.java:[31,2] cannot find symbol
[ERROR]   symbol: class StackTrace
...
```

Let's look at 
```
[ERROR] /src/src/main/java/io/lettuce/core/cluster/event/JfrAskRedirectionEvent.java:[32,38] cannot find symbol
[ERROR]   symbol: class Event
```
This means that [Event](https://github.com/redis/lettuce/blob/8941aea2cb1e3124117cecf9a06c0e254d397911/src/main/java/io/lettuce/core/event/connection/JfrReconnectFailedEvent.java#L33) is not found here. If we go to the documentation of [Event](https://docs.oracle.com/en/java/javase/11/docs/api/jdk.jfr/jdk/jfr/Event.html), we see that the API has only been available since version 9, but in our compilation we are trying to make it compatible with 8. But then the question is how did the original developers compile to target JVM 8?

`maven-compiler-plugin:3.12.1` (which is used in this project uses this version) sets `source` and `target` to `8`. But what they forget to provide (at least mention in the pom) is `-Xbootclasspath`. Without this option, JDK 11 would simply compile to 8 using APIs like `jdk.jfr.Event` from 11. However, the drawback of this approach is that it would still fall to execute with JVM8 because JVM8 does not have the API `jdk.jfr.Event`.

Because of this issue, we run into compilation failures as we pass `release` instead of `source` and `target`. `release` replaces `source`, `target`, and `-Xbootclasspath`. Thus even when compiling with JDK11, `release` ensures that the APIs from 8 are used.

Thus, we should not explicitly provide `release` and let `mvn` build tool read from pom. Some more regressions we had:
```
-avro-1.11.3.jar
-brave-6.0.3.jar
-crud-1.1.0.8.jar
-h2-2.3.232.jar
-jakarta.annotation-api-3.0.0.jar
-langx-java-5.4.4.jar
-lettuce-core-6.3.2.RELEASE.jar
-mybatis-3.5.16.jar
-mybatis-plus-boot-starter-3.5.5.jar
-netty-buffer-4.1.109.Final.jar
-nutz-1.r.72.jar
-spring-boot-starter-web-3.4.4.jar
-wildfly-elytron-security-manager-2.6.0.Final.jar
-woodstox-core-6.6.2.jar
```
@msuozzo you are correct about the intuition. The compilation failures is mostly observed when target <=8.

PR #860, opened 2025-09-30
Use strategy hint if initial reference fails

algomaster99:
I haven't tested this yet. But the logic is to try inference again with Google Search to get the correct source repository.

### DIFF
--- a/internal/agent/logic.go
+++ b/internal/agent/logic.go
@@ -15,6 +15,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/go-git/go-billy/v5"
 	"github.com/go-git/go-billy/v5/memfs"
 	"github.com/go-git/go-git/v5"
 	"github.com/go-git/go-git/v5/plumbing"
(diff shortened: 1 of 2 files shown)


internal/agent/logic.go
```diff
@@ -336,6 +337,58 @@ func (a *defaultAgent) getTools() []*llm.FunctionDefinition {
 	}
 }
 
+func (a *defaultAgent) proposeInferenceWithAIAssist(ctx context.Context, initialErr error, wt billy.Filesystem, str *memory.Storage) (*schema.StrategyOneOf, error) {
+	log.Println("Getting repository hint from AI...")
+	prompt := []string{
```
msuozzo, line 341: we discussed this in another PR but let's try to keep these to a minimal number of raw string literals so there's less syntax noise from quotes and commas.

internal/agent/logic.go
```diff
@@ -359,7 +412,14 @@ func (a *defaultAgent) proposeNormalInference(ctx context.Context) (*schema.Stra
 		},
 	)
 	if err != nil {
-		return nil, errors.Wrap(err, "inferring initial strategy")
+		log.Printf("Normal inference failed: %v", err)
+		wt := memfs.New()
+		str := memory.NewStorage()
+		s, err = a.proposeInferenceWithAIAssist(ctx, err, wt, str)
```
msuozzo, line 416: i'd expect us to limit this to cases where the inference fails in the ways that might indicate a LocationHint might help. not sure whether it's worthwhile to do this unconditional second pass in all cases but we'd certainly want to do so in certain cases.

internal/llm/llm.go
```diff
@@ -61,6 +62,7 @@ type ScriptResponse struct {
 
 func GenerateTextContent(ctx context.Context, client *genai.Client, model string, config *genai.GenerateContentConfig, prompt ...*genai.Part) (string, error) {
 	contents := []*genai.Content{{Parts: prompt, Role: "user"}}
+	log.Printf("%s\n\n", FormatContent(*contents[0]))
```
msuozzo: these look like debugging statements. let's remove

algomaster99: I added them to mimic to the chat session, but you are right I don't want users of GenerateTextContent to be affected.
```
2025/10/15 12:14:16 --- Role: user ---

>>> Type: *genai.Part

  Based on the following inference failure error "rpc error: code = Internal desc = failed to infer strategy: clone failed [repo=https://www.aliyun.com/product/oss]: cloning repo: Get "https://www.aliyun.com/product/oss/info/refs?service=git-upload-pack": dial tcp: lookup www.aliyun.com on 127.0.0.1:53: server misbehaving" for package 'com.aliyun.oss:aliyun-sdk-oss', find the correct source code repository URL.
  Just return the URL WITHOUT any additional text or formatting.
  For example, for the package 'org.apache.camel:camel-support', return 'https://github.com/apache/camel' not 'https://github.com/apache/camel/tree/main/core/camel-support'.
  Use the tools you have at your disposal to find the URL.
  Finally, if you don't find the URL, just return an empty string.

2025/10/15 12:14:29 --- Role: model ---

>>> Type: *genai.Part

  https://github.com/aliyun/aliyun-oss-java-sdk
```

internal/agent/logic.go
algomaster99: >  might indicate a LocationHint might help

I can try to document cases where AI inference failed.

algomaster99: Looking through cases where inference failed despite AI, I think one failure where we can most likely skip this AI inference is when we have `build tool * is recognized but not yet supported`

internal/agent/logic.go
```diff
@@ -359,7 +411,13 @@ func (a *defaultAgent) proposeNormalInference(ctx context.Context) (*schema.Stra
 		},
 	)
 	if err != nil {
-		return nil, errors.Wrap(err, "inferring initial strategy")
+		wt := memfs.New()
+		str := memory.NewStorage()
```
wbxyz: Don't assign new variables for this, overwrite the existing ones. The code further down is going to rely on the worktree and git storage being valid when doing git.Open() and storing a reference to the repo.

internal/agent/session.go
```diff
@@ -26,6 +26,7 @@ type AgentDeps struct {
 	LogsBucket     string
 	GCSClient      *gcs.Client
 	MaxTurns       int
+	Client         *genai.Client
```
wbxyz: nit: I think it's worth differentiating these, I think it should be GenaiClient

We should probably also rename the SessionDeps.Client to match, but that can be outside this PR.

wbxyz (approved):

internal/agent/logic.go
algomaster99: I think I created new instances because there was an edge case where `proposeInferenceWithAIAssist` failed at cloning if the same repository was inferred. I can take a another look.

wbxyz: Oh you can make new values, just re-use the variables to make sure the values escape the code block (`:=` vs `=`):

```
		wt = memfs.New()
		str = memory.NewStorage()
```

wbxyz (approved):

internal/agent/logic.go
algomaster99: It seems that memory storage only allows to clone one repo per instance. It fails with repository already exists if I try to clone any git repo.

PR #866, opened 2025-09-30
Add thinking history

wbxyz:
This replaces a lot of the noise from logging all the iterations, with a
history of the diagnosis and the snippet that was introduced.


internal/agent/logic.go
```diff
@@ -278,12 +279,19 @@ func (a *defaultAgent) genericPrompt() []string {
...
 		"You SHOULD NOT change the ref in the location.",
-		"You might need to update the `dir` in the location to match the root of the source repo",
 		"To debug the build, you might want to use the read_build_logs tool to view the build errors to understand what's going wrong.",
 		"You might also want to inspect the contents of the source repo using the read_repo_file or list_repo_files tools.",
 	}
 }
 
+func (a *defaultAgent) diagnoseOnly() []string {
+	return []string{
+		"Please explain what went wrong with the rebuild, and what might need to be changed to resolve the build.",
```
msuozzo, line 289: okay so for clarity, i think it makes sense to keep these lines as a single string, not split them across entries.

for the cases where you want markdown blocks inside raw golang strings, you could use a small transform to construct them and insert them either via sprintf or concatenation:

```
func block(content, ...tags string) string{
	return "```"+strings.Join(tags, " ")+"\n"+content+"\n```"
}
```

internal/agent/logic.go
```diff
@@ -407,54 +447,87 @@ func (a *defaultAgent) generate(ctx context.Context, prompt []string, opts ...ai
 	return modelResp, err
 }
 
+func (a *defaultAgent) makeDiagnosticPrompt() []string {
+	return append(a.genericPrompt(), a.diagnoseOnly()...)
+}
+
+// One "cycle" of the LLM produces a thoughtT
+type thoughtT struct {
```
msuozzo: we tend to use T suffixes for generic types so maybe "thought" would suffice.

internal/agent/logic.go
wbxyz: Agreed, but I'll do this in a follow on PR, that uses templating.

internal/agent/logic.go
wbxyz: Changed to thoughtData

PR #873, opened 2025-10-01
Check for job execution crashes when running agent benchmark

wbxyz:


tools/ctl/ctl.go
```diff
@@ -707,6 +713,14 @@ var runAgentBenchmark = &cobra.Command{
 				if session.Status == schema.AgentSessionStatusCompleted {
 					break
 				}
+				exec, err := runService.Projects.Locations.Jobs.Get(resp.JobName).Do()
```
msuozzo: might be nice to also clean up and complete the session but either way is fine

wbxyz: Yeah I thought about it originally, but I didn't want to add a new session writer. I was thinking about maybe using the complete session handler in this code but eventually decided the complexity wasn't worth it yet.

I added a TODO.

PR #889, opened 2025-10-07
Write agent messages to session bucket

wbxyz:


internal/agent/logic.go
```diff
@@ -520,14 +517,25 @@ func (a *defaultAgent) makePrompt(ctx context.Context) []string {
...
 	contentParts := []*genai.Part{genai.NewPartFromText(strings.Join(prompt, "\n"))}
 	for content, err := range a.deps.Chat.SendMessageStream(ctx, contentParts...) {
 		if err != nil {
 			return "", errors.Wrap(err, "chat error")
 		}
 		log.Printf("%s\n\n", llm.FormatContent(*content))
+		if opts.chatUploadPath != "" {
+			// Use timestamp as filename to avoid needing a cross-loop count of messages written so far
+			contentPath := path.Join(opts.chatUploadPath, time.Now().UTC().Format(time.RFC3339Nano)+"-"+content.Role+".json")
+			writer := a.deps.GCSClient.Bucket(a.deps.SessionBucket).Object(contentPath).NewWriter(ctx)
```
algomaster99: This line is throwing nil pointer dereference in my experiments. I think it is happening because `SessionsBucket` is not registered in ctl.go. `SessionsBucket: "", // TODO: Add this once it's being used.`.

internal/agent/session.go
```diff
@@ -47,7 +56,10 @@ type RunSessionDeps struct {
 }
 
 func doIteration(ctx context.Context, sessionID string, iterNum int, agent Agent, deps RunSessionDeps) (*schema.AgentIteration, error) {
-	s, err := agent.Propose(ctx)
+	s, err := agent.Propose(ctx, &ProposeOpts{
+		// Storage scheme is gs://<sessionBucket>/<sessionID>/messages/<iternNum>/*
+		chatUploadPath: path.Join(sessionID, "messages", fmt.Sprintf("%d", iterNum)),
```
algomaster99: Related comment to the previous comment, but you should only set this if session bucket is configured.

internal/agent/logic.go
wbxyz: Yeah sorry, I wasn't testing with local-agent.

Updated to only upload when the ProposeOpt is populated. I also changed the ProposeOpt to be a URL to capture the bucket name in addition to the path. That should also make it easy to add local-file storage if we need to.

internal/agent/session.go
wbxyz: Reworked this section a bit, no longer an issue.

internal/agent/logic.go
```diff
@@ -520,14 +517,32 @@ func (a *defaultAgent) makePrompt(ctx context.Context) []string {
...
+			contentPath := path.Join(opts.ChatUploadURL.Path, time.Now().UTC().Format(time.RFC3339Nano)+"-"+content.Role+".json")
+			var writer io.WriteCloser
+			switch opts.ChatUploadURL.Scheme {
+			case "gs":
+				writer = a.deps.GCSClient.Bucket(opts.ChatUploadURL.Host).Object(contentPath).NewWriter(ctx)
+			// TODO: implement local fs storage
+			default:
+				return "", fmt.Errorf("unsupported chat upload scheme: \"%s\"", opts.ChatUploadURL.Scheme)
+			}
+			// Use timestamp as filename to avoid needing a cross-loop count of messages written so far
```
msuozzo: this looks to be in the wrong spot

internal/agent/logic.go
```diff
@@ -520,14 +517,32 @@ func (a *defaultAgent) makePrompt(ctx context.Context) []string {
...
+func (a *defaultAgent) generate(ctx context.Context, prompt []string, opts *ProposeOpts) (string, error) {
 	var response genai.Content
 	contentParts := []*genai.Part{genai.NewPartFromText(strings.Join(prompt, "\n"))}
 	for content, err := range a.deps.Chat.SendMessageStream(ctx, contentParts...) {
 		if err != nil {
 			return "", errors.Wrap(err, "chat error")
 		}
 		log.Printf("%s\n\n", llm.FormatContent(*content))
+		if opts.ChatUploadURL != nil {
+			contentPath := path.Join(opts.ChatUploadURL.Path, time.Now().UTC().Format(time.RFC3339Nano)+"-"+content.Role+".json")
```
msuozzo: would you want to just index them instead? e.g. 0-user.json, 1-model.json

if you're intending on storing them in a flat directory and are concerned about collisions, you could initialize a fixed identifier for the generate call at the start like a timestamp or uuid and use that as the prefix or additional disambiguation.

internal/agent/logic.go
wbxyz: Whoops, didn't get moved with a refactor. Fixed.

internal/agent/logic.go
wbxyz: Updated to use the same timestamp across all message in the generate() call. The path is starting to have a lot of elements to it but I think they're all pretty reasonable. The updated schema is

```
gs://<session-bucket>/<session-id>/<iteration-num>/<generate-timestamp>-<part-num>-<role>.json
```

msuozzo: Yeah I think this works. Lgtm

PR #890, opened 2025-10-07
Agent API can read metadata

wbxyz:
This is necessary for the API to do artifact comparison


cmd/agent/main.go
```diff
@@ -88,10 +89,15 @@ func main() {
 	if err != nil {
 		log.Fatal("Failed to create genai client: ", err)
 	}
+	gcsClient, err := gcs.NewClient(ctx)
+	if err != nil {
+		log.Fatal("Failed to create GCS client: ", err)
+	}
```
algomaster99: Also needs to be configured in ctl.

PR #897, opened 2025-10-10
Enable running benchmark locally

algomaster99:
The PR would enable running benchmark locally. This helps to not rely on Google Cloud infrastructure to run rebuilds. This also helps leveraging tui to view the local runs.

Todos:
- [x] support for putting inference logs in assets
- [x] it seems the local runs are hardcoded to `smoketest`; we would want it to be flexible.
~- [ ] nit: the docker run output is also redirected to stdout. This causes the progress bar to render multiple times thus creating disruption in the log.~ not that important for this issue.
  Example:
  ```
   0 / 1 [------------------------------------------------------------------------------------]   0.00%Get:1 http://deb.debian.org/debian trixie InRelease [140 kB]
	Get:2 http://deb.debian.org/debian trixie-updates InRelease [47.3 kB]
	Get:3 http://deb.debian.org/debian-security trixie-security InRelease [43.4 kB]
	Get:4 http://deb.debian.org/debian trixie/main amd64 Packages [9669 kB]
	 0 / 1 [------------------------------------------------------------------------------------]   0.00%Get:5 http://deb.debian.org/debian trixie-updates/main amd64 Packages [5412 B]
	Get:6 http://deb.debian.org/debian-security trixie-security/main amd64 Packages [52.2 kB]
	 0 / 1 [------------------------------------------------------------------------------------]   0.00%Fetched 9957 kB in 1s (9788 kB/s)
	 0 / 1 [------------------------------------------------------------------------------------]   0.00%
	 0 / 1 [------------------------------------------------------------------------------------]   0.00%
	Reading state information...
	 0 / 1 [------------------------------------------------------------------------------------]   0.00%
  ```
~- [ ] make output directory configurable~ reading and writing of logs both are tied to this so we should not do this in this PR.
- [x] retrieve results using `get-results`

### DIFF
--- a/pkg/build/local/run_planner.go
+++ b/pkg/build/local/run_planner.go
@@ -104,7 +104,7 @@ func (p *DockerRunPlanner) generateScript(instructions rebuild.Instructions, inp
 	if err := dockerRunScriptTpl.Execute(&buf, dockerRunScriptArgs{
 		Inst:           instructions,
 		PackageManager: pkgMgr,
-		UseTimewarp:    timewarpURL != "",
+		UseTimewarp:    opts.UseTimewarp,
 		TimewarpURL:    timewarpURL,
 		TimewarpAuth:   timewarpAuth,
 	}); err != nil {
--- a/tools/benchmark/run/local.go
+++ b/tools/benchmark/run/local.go
@@ -132,7 +132,9 @@ func executeBuild(ctx context.Context, t rebuild.Target, strategy rebuild.Strate
 			ToolURLs: map[build.ToolType]string{
 				build.TimewarpTool: opts.PrebuildURL + "/timewarp",
 			},
+			BaseImageConfig: build.DefaultBaseImageConfig(),
 		},
+		UseTimewarp: meta.AllRebuilders[t.Ecosystem].UsesTimewarp(input),
 	}
 	handle, err := executor.Start(ctx, input, buildOpts)
 	if err != nil {
--- a/tools/ctl/rundex/rundex.go
+++ b/tools/ctl/rundex/rundex.go
@@ -617,7 +617,7 @@ func (f *LocalClient) WatchRebuilds() <-chan *Rebuild {
 }
 
 func (f *LocalClient) WriteRebuild(ctx context.Context, r Rebuild) error {
-	path := filepath.Join(layout.RundexRebuildsDir, r.Ecosystem, r.Package, r.Artifact, rebuildFileName)
+	path := filepath.Join(layout.RundexRebuildsDir, r.RunID, r.Ecosystem, r.Package, r.Artifact, rebuildFileName)
 	file, err := f.fs.Create(path)
 	if err != nil {
 		return errors.Wrap(err, "creating file")


tools/benchmark/run/local.go
```diff
@@ -131,7 +147,9 @@ func executeBuild(ctx context.Context, t rebuild.Target, strategy rebuild.Strate
 			ToolURLs: map[build.ToolType]string{
 				build.TimewarpTool: opts.PrebuildURL + "/timewarp",
 			},
+			BaseImageConfig: build.DefaultBaseImageConfig(),
 		},
+		UseTimewarp: meta.AllRebuilders[t.Ecosystem].UsesTimewarp(input),
 	}
```
algomaster99, line 138: No need this for as `UseTimewarp` is false.

algomaster99: @msuozzo won't we need this for other ecosystems though?

tools/benchmark/run/local.go
```diff
@@ -77,7 +78,11 @@ func (s *localExecutionService) RebuildPackage(ctx context.Context, req schema.R
 		}
 	}
 	verdict := &schema.Verdict{Target: t}
+	outBuf := &bytes.Buffer{}
+	log.SetOutput(io.MultiWriter(outBuf, s.logsink))
 	strategy, err := s.infer(ctx, t, mux)
+	uploadInferenceLogs(ctx, s.store, t, outBuf.Bytes())
+	log.SetOutput(s.logsink)
```
algomaster99: @msuozzo this time we don't set log output to `nil` so it should be fine. I am inclined to keep it as is. Okay for you?

tools/benchmark/run/local.go
wbxyz: I'm not sure I understand the question?

Some ecosystems enable timewarp while others don't. That's why there's a switch based on ecosystem here.

pkg/build/local/run_planner.go
wbxyz: Can this be a separate PR?

tools/ctl/ctl.go
```diff
@@ -244,12 +244,17 @@ var getResults = &cobra.Command{
 		if err != nil {
 			log.Fatal(err)
 		}
-		fireClient, err := rundex.NewFirestore(cmd.Context(), *project)
-		if err != nil {
-			log.Fatal(err)
+		var client rundex.Reader
```
wbxyz: nit: We setup a lot of client objects, I think this should probably be named `dex`

tools/ctl/ctl.go
```diff
@@ -515,7 +520,7 @@ var runBenchmark = &cobra.Command{
 				ID:            runID,
 				BenchmarkName: filepath.Base(args[1]),
 				BenchmarkHash: hex.EncodeToString(set.Hash(sha256.New())),
-				Type:          string(schema.SmoketestMode),
```
wbxyz, line 521: I think I'd prefer to keep this, or maybe make a new execution mode to hardcode along the lines of "LocalMode".

The alternative from SmoketestMode is AttestMode which a local execution definitely isn't doing because it's not going to create an attestation. Also the execution mode will be used to locate the different build assets, which is why you're getting that error about info.json (an expected side effect of a GCB based build). The SmoketestMode probably works better out of the box, because that expects all the artifacts to be located in a single dir.

Were you having problems with SmoketestMode before?

wbxyz:
Thanks for all the work you did to diagnose and improve this!

Let's pull out the inference logs capturing into a separate PR. It's definitely helpful but I'm not sure how we might do that in the hosted version of OSS Rebuild yet. Ideally I'd like this "enable running benchmark locally" PR to bring the local execution *closer* to the remote execution and that's kind of moving in the opposite direction.

tools/benchmark/run/local.go
algomaster99: The suggestion here was to set `UseTimewarp` to `false`. However, I suggested to keep `meta.AllRebuilders[t.Ecosystem].UsesTimewarp(input)` since it decides the boolean based on ecosystem.

tools/ctl/ctl.go
algomaster99: > Were you having problems with SmoketestMode before?

While running `SmoketestMode`, I was only getting `firestore.json` and now when I inspect it, I see that [smoketest is not implemented for smoketest worker](https://github.com/google/oss-rebuild/blob/595126577a646505ed0daa4da8299666fad0917f/tools/benchmark/run/local.go#L228). Even when I implement it using `rebuilderservice.RebuildSmoketest`, I see errors `rpc error: code = Internal desc = fetching package version: 400 Bad Request` which is returned when pulling package versions from `https://search.maven.org/solrsearch`.

I will investigate the bad request later, but we probably need to  first implement the smoketest build for smoketest worker?

tools/benchmark/run/local.go
```diff
@@ -196,7 +199,10 @@ func compare(ctx context.Context, t rebuild.Target, store rebuild.LocatableAsset
 			return errors.Wrap(err, "getting debian artifact URL")
 		}
 	case rebuild.Maven:
-		return errors.New("maven comparison not implemented")
+		upstreamURL, err = mux.Maven.ReleaseURL(ctx, t.Package, t.Version, maven.TypeJar)
```
msuozzo: wanna split this out, too?

tools/ctl/ctl.go
msuozzo: we've discussed imminently removing smoketest altogether so let's not put work into adding that functionality

wbxyz: Whoops I didn't send my message.

Yeah agreed, don't implement smoketest mode stuff. I think the weird behavior we're relying on is that we mark the run as smoketest (to help the butler fetch artifacts correctly) while we actually execute the run in attest mode, which is the only implemented mode for local executor.

wbxyz: (And I think that weird behavior is fine for now, I'll sort it out as part of a wider Butler cleanup and smoketest deprecation that I have in my backlog).

tools/benchmark/run/local.go
algomaster99: Into what? Should this not be a part of this PR?

tools/ctl/ctl.go
algomaster99: I reverted the change. Now it will run in smoketest mode which means there will still be issues fetching outputs but they can be resolved in a follow-on.

tools/benchmark/run/local.go
msuozzo: Well, this seems to be meaningfully different from the rest of the bug fixes. This is explicitly adding platform support from maven.

tools/ctl/ctl.go
```diff
@@ -247,12 +247,17 @@ var getResults = &cobra.Command{
 		if err != nil {
 			log.Fatal(err)
 		}
-		fireClient, err := rundex.NewFirestore(cmd.Context(), *project)
-		if err != nil {
-			log.Fatal(err)
+		var dex rundex.Reader
```
msuozzo: on second look, this is a change to getResults which is more about viewing results locally, not running local benchmarks. so this, too, should be split into a separate change, right?

tools/ctl/ctl.go
algomaster99: This change does not matter as this value is being written to `rundex/runs_metadata/<timestamp>`. The mode to run in is decided by `RunBench` method later in the code.  For now, **only attest works for running the benchmark and getting results** (export is broken).

PR #898, opened 2025-10-13
Strip minor and patch version from JDK

algomaster99:
This is to ensure that we don't fallback to JDK 11 even when we have inferred JDKs like `17.0.6`.


pkg/rebuild/maven/infer.go
```diff
@@ -161,18 +161,45 @@ func inferBuildTool(commit *object.Commit) (string, error) {
...
 	}
-	return jdk, nil
+	// First, check if the inferred version is directly available for download.
+	if _, ok := JDKDownloadURLs[rawJDK]; ok {
+		log.Printf("Using directly inferred JDK version: %s", rawJDK)
+		return rawJDK, nil
+	}
+	// If not directly available, try parsing its major version.
+	log.Printf("Inferred JDK \"%s\" is not directly available, attempting to parse major version.", rawJDK)
+	var parsedJDK string
```
msuozzo: nit: rename this and the associated comments to "majorJDK" since the process here is attempting to build with the closest major version.

PR #904, opened 2025-10-16
Make previous instructions an input to historyContext

wbxyz:


internal/agent/logic.go
```diff
@@ -415,78 +414,99 @@ func (a *defaultAgent) ecosystemExpertise() []string {
...
+		})
+		if err != nil {
+			log.Println(errors.Wrap(err, "generating instructions"))
+		} else {
+			d.Instructions = inst
+		}
+	}
+	// Dockerfile
+	{
+		dockerfile, err := a.dockerfile(ctx, iteration.ObliviousID)
```
msuozzo: d.Dockerfile, err = ...

internal/agent/logic.go
```diff
@@ -415,78 +414,99 @@ func (a *defaultAgent) ecosystemExpertise() []string {
...
+	if iteration.Strategy == nil {
+		return d
+	}
+	s, err := iteration.Strategy.Strategy()
+	if err != nil {
+		return d
+	}
+	// Instructions
+	{
+		inst, err := s.GenerateFor(a.t, rebuild.BuildEnv{
```
msuozzo, line 456: want to add a note here that this can be inconsistent with what actually was used for the Dockerfile?

internal/agent/logic.go
```diff
@@ -415,78 +414,99 @@ func (a *defaultAgent) ecosystemExpertise() []string {
...
+func (a *defaultAgent) execDetails(ctx context.Context, iteration *schema.AgentIteration) *executionDetails {
+	d := &executionDetails{Iteration: *iteration}
+	if iteration.Strategy == nil {
+		return d
+	}
+	s, err := iteration.Strategy.Strategy()
+	if err != nil {
+		return d
+	}
+	// Instructions
```
msuozzo: want to fold these onto trailing comments on the line below? you also might add a bit more here to make this more than a word

PR #908, opened 2025-10-17
Add initial version of diffr tool

msuozzo:


pkg/diffr/diffr_test.go
```diff
@@ -0,0 +1,398 @@
...
+			}
+			rightFile := File{
+				Name:   tc.rightName,
+				Reader: bytes.NewReader(rightData),
+			}
+
+			// Test text output
+			t.Run("text_output", func(t *testing.T) {
+				opts := Options{OutputJSON: false}
+				match, output, err := Diff(leftFile, rightFile, opts)
```
wbxyz: Diff returning a boolean True as a way to indicate no diff seems counterintuitive to me. Is there a better way to express that? Maybe Diff could return empty output on match? Or maybe we could return a new object type DiffResult that has an Empty() method? We could push the string vs json logic into that DiffResult with AsJSON() and AsString() methods.

Alternatively, maybe Diff should just return True when there *is* a diff?

pkg/diffr/filetype.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	peek := buf[:n]
+	switch {
+	case bytes.HasPrefix(peek, gzipMagic):
+		return TypeGzip, nil
+	case bytes.HasPrefix(peek, zipMagic):
+		return TypeZip, nil
+	case len(peek) >= tarMagicOffset+len(tarMagic) && bytes.Equal(peek[tarMagicOffset:tarMagicOffset+len(tarMagic)], tarMagic):
+		return TypeTar, nil
+	case isBinary(peek[:min(len(peek), 1024)]):
```
wbxyz: Nit: use minPeekSize for the const?

And maybe this isn't necessary because buf is already at most minPeekSize?

pkg/diffr/zip.go
```diff
@@ -0,0 +1,34 @@
...
+}
+
+// formatZipListing produces a one-line summary of a zip file entry
+// -rw-r--r-- Deflate 4          1980-01-01 00:00:00.000000 foo.txt
+func formatZipListing(f *zip.FileHeader) string {
+	return fmt.Sprintf("%-10s %-8s %-12d %-26s %s\n",
+		f.Mode().String(),
+		formatMethod(f.Method),
+		f.UncompressedSize64,
+		f.Modified.UTC().Format("2006-01-01 15:04:05.000000"),
```
wbxyz: Hmmm I think this is using the month twice? 2006-01-02

pkg/diffr/tar.go
```diff
@@ -0,0 +1,23 @@
...
+
+// formatTarListing produces a one-line summary of a tar file entry
+// -rw-r--r-- 0 0            5 1970-01-01 00:00:00.000000 file.txt
+func formatTarListing(h *tar.Header) string {
+	return fmt.Sprintf("%-10s %d %d %12d %-26s %s\n",
+		os.FileMode(h.Mode).String(),
+		h.Uid,
+		h.Gid,
+		h.Size,
+		h.ModTime.UTC().Format("2006-01-01 15:04:05.000000"),
```
wbxyz: Same, month is being used twice?

pkg/diffr/output.go
```diff
@@ -0,0 +1,102 @@
...
+			builder.WriteString(prefix)
+			builder.WriteString(detailGlyph)
+			builder.WriteString("  --- ")
+			builder.WriteString(node.Source1)
+			builder.WriteString("\n")
+			builder.WriteString(prefix)
+			builder.WriteString(branchGlyph)
+			builder.WriteString("+++ ")
+			builder.WriteString(node.Source2)
+			builder.WriteByte('\n')
```
wbxyz: Can we reduce the noise here to make the output more obvious? Elsewhere in the codebase we have been trying to avoid complicated string construction.

I suspect the usual options aren't viable due to some speed requirements (templating, fmt.Sprintf, etc).

If so wdyt of something like this?

```golang
type builder struct {
  strings.Builder
}

func (b *builder) WriteLine(s ...string) {
    for _, st := range s {
        b.WriteString(st)
    }
    b.WriteByte('\n')
}

builder.WriteLine(prefix, detailGlyph, "  --- ", node.Source1)
builder.WriteLine(prefix, detailGlyph, "  +++ ", node.Source2)
```

pkg/diffr/output.go
```diff
@@ -0,0 +1,102 @@
...
+			// Standard formatting for all other diffs
+			builder.WriteString(prefix)
+			builder.WriteString(branchGlyph)
+			builder.WriteString(node.Source1)
+			builder.WriteByte('\n')
+			// Content prefix for nested content
+			contentPrefix := prefix + detailGlyph
+			// Format and print any comments for this specific node
+			if len(node.Comments) > 0 {
+				commentPrefix := strings.TrimSuffix(contentPrefix, detailGlyph) + commentGlyph
```
wbxyz: nit: Can this be constructed as `prefix + commentGlyph`?

pkg/diffr/tar.go
```diff
@@ -0,0 +1,185 @@
...
+			header:        hdr,
+			contentOffset: offset,
+		}
+		listing2.WriteString(formatTarListing(hdr))
+	}
+	// Compare listings
+	match := true
+	if listing1.String() != listing2.String() {
+		match = false
+		listingDiff, err := gitdiff.Strings(listing1.String(), listing2.String())
```
wbxyz: nit: Avoid casting to string twice?

pkg/diffr/output.go
msuozzo: yeah this is a good idea. i'd actually considered this initially and it got removed in a refactor. the code's a lot cleaner using this form, though.

pkg/diffr/zip.go
msuozzo: oof good catch

pkg/diffr/output.go
```diff
@@ -0,0 +1,89 @@
...
+	// Write top-level comments from the root node
+	for _, comment := range node.Comments {
+		builder.WriteLine(commentGlyph, comment)
+	}
+	// Write root node's unified diff if present
+	if node.UnifiedDiff != nil {
+		content := strings.TrimSuffix(*node.UnifiedDiff, "\n")
+		for line := range strings.SplitSeq(content, "\n") {
+			builder.WriteLine(line)
+		}
```
wbxyz: Now that it's written like this, why not directly add the UnifiedDiff?

Is it something about ensuring exactly one trailing newline for non-empty diffs?

wbxyz: Oh I guess it's mirroring the code below that needs to add a prefix on the line. Maybe worth keeping as is? Up to you.

wbxyz (approved):

pkg/diffr/output.go
msuozzo: yeah let me add a comment to specify this.

pkg/diffr/diffr_test.go
msuozzo: yeah agreed. i projected everything onto the error which i think is a lot clearer: "Diff returns nil when there is no diff"

pkg/diffr/output.go
msuozzo: eh i just inlined it

pkg/diffr/diffr_test.go
wbxyz: Nice I think that's a big readability improvement.

PR #921, opened 2025-10-24
Support pypi routes specifying content-type

msuozzo:


internal/timewarp/timewarp.go
```diff
@@ -405,13 +420,96 @@ func timeWarpPyPIProjectRequest(client httpx.BasicClient, obj map[string]any, at
...
+			fixedVersions = append(fixedVersions, vStr)
+		}
+	}
+	obj["versions"] = fixedVersions
 	return nil
 }
 
+// extractVersionFromFilename parses a wheel or sdist filename to find its version.
+// This is a simplified parser inspired by Python's packaging.utils.
+func extractVersionFromFilename(filename string) string {
```
wbxyz: Worth adding that this is python-specific? Maybe extractVersionFromPYPIFilename?

internal/timewarp/timewarp.go
```diff
@@ -405,13 +420,96 @@ func timeWarpPyPIProjectRequest(client httpx.BasicClient, obj map[string]any, at
...
+	if strings.HasSuffix(filename, ".whl") {
```
wbxyz, line 492: filepath.Ext()?

internal/timewarp/timewarp.go
msuozzo: Done

internal/timewarp/timewarp.go
msuozzo: eh then it wouldn't work with ".tar.gz" below. i think keeping them consistent is tolerable

PR #928, opened 2025-11-05
Display all details in the TUI modal

wbxyz:


tools/ctl/ide/details/details.go
```diff
@@ -46,10 +46,13 @@ func Format(example rundex.Rebuild) (string, error) {
 
 func View(example rundex.Rebuild) (modal.InputCaptureable, error) {
 	details := tview.NewTextView()
-	text, err := Format(example)
```
msuozzo, line 49: is Format still used? can we delete it?

msuozzo (approved):

tools/ctl/ide/details/details.go
wbxyz: It gets used in rebuildhistory, so I pushed the new indent thing into Format

PR #929, opened 2025-11-05
Generate new debian benchmark from stable popcon

wbxyz:


msuozzo (commented):
For these cases, we've copied the code to generate a new named benchmark. we should continue that pattern here (unless we want to abandon the non-stable benchmark)

wbxyz:
I don't think it would be a bad idea to move to stable, but I can keep both for backwards compatibility

PR #930, opened 2025-11-06
Fix build definiton path

algomaster99:
In https://github.com/google/oss-rebuild/commit/d37534d36f3e8883f87678de1836a7dd8d026a58, we introduced a file name with `:`. This breaks `go get` on any commit after this commit because `:` is not allowed to be in path. I found this [issue comment](https://github.com/golang/go/issues/28001#issuecomment-427031333) which tells that `:` is reserved character on Windows so to preserve cross-compatibility, we should not have `:` in filenames.

~I changed it to `!` because I saw that as a delimiter in TUI, but I can change them to whatever is more preferred.~
Chose to go with `-` instead. `!` _may_ pose a problem as it may be a special character somewhere :sweat_smile:


algomaster99:
Another side effect is reflected on https://deps.dev/go/github.com%2Fgoogle%2Foss-rebuild. There are no versions of the go module after we introduced the definition.

msuozzo:
okay yeah this is breaking things. I don't think this is a sufficient fix but we need to get the module unbroken.

msuozzo (approved):

PR #938, opened 2025-11-09
Add interface to encode Target attributes

msuozzo:


pkg/rebuild/rebuild/sanitize_test.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+func TestDefaultEncoder(t *testing.T) {
+	canonical := Target{
+		Ecosystem: PyPI,
+		Package:   "absl-py",
+		Version:   "2.0.0",
+		Artifact:  "absl_py-2.0.0-py3-none-any.whl",
+	}
+	encoded := FilesystemTargetEncoding.Encode(canonical)
```
algomaster99, line 135: You can also add `FirestoreTargetEncoding` for completeness.

algomaster99 (approved):
Great solution, and it looks good to me. Just pointed out a nitpick.

pkg/rebuild/npm/sanitize.go
```diff
@@ -0,0 +1,42 @@
...
+//   - Note: '@' is allowed in Firestore document IDs, so it's kept as-is
+//
+// Filesystem/GCS Encoding:
+//   - TODO: Breaks backward compatibility
+//   - Will use same encoding as Firestore when implemented
+
+// firestoreEncoder encodes NPM package identifiers for Firestore document IDs.
+// Replaces forward slash with exclamation mark since it's forbidden in Firestore.
+var firestoreEncoder = &rebuild.TargetEncoder{
+	Package:  rebuild.MapTransform(map[rune]rune{'/': '!'}),
```
algomaster99, line 35: `!` is also a separator between `Package`, `Version`, and `Artifact`. Maybe it is better to choose a different delimiter to clearly identify an npm package?

pkg/rebuild/debian/sanitize.go
```diff
@@ -0,0 +1,21 @@
+// Copyright 2025 Google LLC
+// SPDX-License-Identifier: Apache-2.0
+
+package debian
+
+// Debian Package Identifier Constraints
+//
+// Debian uses a single package name format:
+//   - Package names like "libc6", "python3.11", "g++"
+//   - Strictly lowercase alphanumeric with limited special characters
```
wbxyz: Worth mentioning that debian has a something called "Component" which is a grouping of packages, and we include that in the Target.Package. For example, `main/git`. Adding a filesystem sanitizer for that would be helpful.

pkg/rebuild/rebuild/sanitize.go
```diff
@@ -0,0 +1,180 @@
...
+	//
+	// Filesystem (Windows):
+	//   - Forbidden characters: \ / : * ? " < > |
+	//   - Colon (:) is reserved for alternate file streams
+	//   - Reference: https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file
+	//
+	// GCS Object Names:
+	//   - Technically allows all Unicode, but same restrictions as Windows filesystem
+	//     should be applied for cross-platform compatibility
+	//   - Reference: https://cloud.google.com/storage/docs/objects
```
wbxyz, line 56: Can we express these constraints as a lambda or a regex (or list of regexes) on targetEncoding?

That way a validation can happen during Encode that verifies the ecosystem implemented those targets correctly.

pkg/rebuild/npm/sanitize.go
wbxyz: Do you mean for visualization in the TUI? I think that's configured here: https://github.com/google/oss-rebuild/blob/5e7d3c7a3836f9876e1349999e5bb9e580c75306/tools/ctl/rundex/rundex.go#L77

And I think that separator is only used for human visualization, and should be fine to overlap with this internal encoding. If we need to we can change that human visualization more easily than the existing NPM encoding escape sequences.

pkg/rebuild/debian/sanitize.go
```diff
@@ -0,0 +1,50 @@
...
+// Filesystem/GCS Encoding:
+//   - Replaces '/' with '~' (tilde)
+//   - Example: "main/gcc-12" → "main~gcc-12"
+//
+// Firestore Encoding:
+//   - Replaces '/' with '!' (exclamation mark)
+//   - Example: "main/gcc-12" → "main!gcc-12"
+
+var filesystemEncoder = &rebuild.TargetEncoder{
+	Package:  rebuild.MapTransform(map[rune]rune{'/': '~'}),
```
wbxyz, line 36: Why use `~` instead of `!`? I know debian packages at the very least use `~` in version numbers, idk about package names. Might be good to avoid if possible?

pkg/rebuild/npm/sanitize.go
algomaster99: I suggested using another delimiter because I assumed we may need to decode the target to `make benchmark`, as an example. But I guess we would still have access to canonical versions so it is not necessary.

msuozzo: fair point. but considering we're really just changing the implementation of the current behavior, we can tackle that issue later (and i think it'd be easier to change the rundex representation rather than what we use for firestore.

algomaster99: > and i think it'd be easier to change the rundex representation rather than what we use for firestore.

agreed

pkg/rebuild/rebuild/sanitize.go
msuozzo: i'd be on board with this but, considering this PR is no worse than what we have now, i'm inclined to have that be a FR for a follow-on.

pkg/rebuild/debian/sanitize.go
msuozzo: my rationale is nothing more than `!` has context-independent special meaning in bash and we're more likely to be typing out filesystem encodings. `~` also has meaning but dir expansion isn't done in the context of a string so the net effect is you can use it without surrounding the path in quotes.

the presence in versions isn't such an issue in my mind considering we have part-specific encoding support (largely for this exact scenario)

pkg/rebuild/rebuild/sanitize_test.go
msuozzo: eh they use the same default pattern so this is more testing that defaulting itself works, not the fs encoder in particular.

pkg/rebuild/debian/sanitize.go
msuozzo: fixed

pkg/rebuild/rebuild/sanitize.go
wbxyz: Fair enough, worth adding a TODO?

wbxyz (approved):

pkg/rebuild/debian/sanitize.go
wbxyz: Adding a test case with a debian package that uses "~" in the version mitigates this concern, thanks.

pkg/rebuild/rebuild/sanitize.go
msuozzo: eh i'll just start drafting the follow-on

PR #941, opened 2025-11-10
Add Privileged option to RequiredEnv

wbxyz:


pkg/build/gcb/planner.go
```diff
@@ -328,6 +328,7 @@ type Planner struct {
 	timewarpHost           string
 	hasRepo                bool
 	preferPreciseToolchain bool
+	allowPrivileged        bool
```
msuozzo, line 331: i don't love "allow" here. this isn't conditional afaict, it means privilege _will_ be enabled. so we could keep this "privileged" or "usePrivilege" or "root" or "execAsRoot" or "privilegedExecution" or something else

pkg/build/gcb/planner.go
```diff
@@ -430,10 +433,10 @@ func (p *Planner) getToolURL(toolType build.ToolType, opts build.PlanOptions) (t
 }
 
 // generateSteps creates the Cloud Build steps using remote rebuild patterns
-func (p *Planner) generateSteps(target rebuild.Target, dockerfile string, opts build.PlanOptions) ([]*cloudbuild.BuildStep, error) {
+func (p *Planner) generateSteps(target rebuild.Target, dockerfile string, envReqs rebuild.EnvReqs, opts build.PlanOptions) ([]*cloudbuild.BuildStep, error) {
```
msuozzo: maybe just `reqs`

msuozzo (approved):

pkg/build/gcb/planner.go
wbxyz: >  this isn't conditional afaict, it means privilege will be enabled

It is conditional, this one is the gate on the planner that "grants" an instructions request to be privileged:

```golang
	var privileged bool
	if reqs.Privileged {
		if p.allowPrivileged {
			privileged = true
		} else {
			log.Println("Warning: instructions requested privileged execution but this planner does not allow privileged builds.")
		}
	}
```

PR #942, opened 2025-11-10
Move SystemDeps to EnvReqs

wbxyz:
This makes it easier to expand Instructions with future requirements on
the build environement. In a followup PR, we can add a boolean field for
privileged execution

### DIFF
--- a/internal/agent/logic.go
+++ b/internal/agent/logic.go
@@ -691,7 +691,7 @@ func (a *defaultAgent) proposeAgentInference(ctx context.Context, opts *ProposeO
 	// TODO: Try to format the bash script into a structured strategy?
 	strat := rebuild.ManualStrategy{
 		Location:   a.loc,
-		SystemDeps: prev.Instructions.SystemDeps,
+		Requires:   prev.Instructions.Requires,
 		Deps:       prev.Instructions.Deps,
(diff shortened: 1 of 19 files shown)


internal/agent/logic.go
```diff
@@ -691,7 +691,7 @@ func (a *defaultAgent) proposeAgentInference(ctx context.Context, opts *ProposeO
 	// TODO: Try to format the bash script into a structured strategy?
 	strat := rebuild.ManualStrategy{
 		Location:   a.loc,
-		SystemDeps: prev.Instructions.SystemDeps,
+		EnvReqs:    prev.Instructions.EnvReqs,
```
msuozzo: Maybe "Requires" as this field name? Then we get Requires.Privilege and Requires.SystemDeps.

Not sure what the struct should be named tho. Open to suggestions. It's going to live alongside BuildEnv so maybe RequiredEnv or RequiresEnv?

wbxyz: Yeah Requires is goo. I had to update a handful of places with that (WorkflowStrategy, ManualStrategy, Instructions). I also updated the json and yaml encoding to be `"requires"`. I went with RequiredEnv for the struct.

PR #943, opened 2025-11-10
Change debrebuild to use sbuild+unshare

wbxyz:


pkg/rebuild/debian/strategy.go
```diff
@@ -217,7 +211,10 @@ Signed-By: /usr/share/keyrings/debian-archive-keyring.gpg`))),
 		Name: "debian/build/debrebuild",
 		Steps: []flow.Step{
 			{
-				Runs: "debrebuild --buildresult=./out --builder=mmdebstrap {{ .With.buildinfo }}",
+				Runs: textwrap.Dedent(`
+				echo "root:100000:65536" | tee /etc/subuid
```
msuozzo: want to add a short code comment either linking to the docs for this or explaining the rationale for this setup?

wbxyz: Added.

PR #946, opened 2025-11-14
Add ini parser

msuozzo:


pkg/ini/ini.go
```diff
@@ -0,0 +1,199 @@
...
+func (f *File) EnsureSection(name string) *Section {
+	if s, exists := f.Sections[name]; exists {
+		return s
+	}
+	s := &Section{
+		Name:   name,
+		Values: make(map[string]string),
+	}
+	f.Sections[name] = s
+	return s
```
wbxyz: Wdyt of collapsing the extra local variables like

```go
if _, exists := f.sections[name]; !exists {
  f.Sections[name] = &Section{
   ...
  }
}
return f.Sections[name]
```

pkg/ini/ini.go
```diff
@@ -0,0 +1,199 @@
...
+				// Comments ignored and don't break the block
+				continue
+			} else if isEmpty {
+				// Blanks only kept if followed by another continuation
+				blankLineBuffer++
+				continue
+			} else if indent > keyIndent {
+				// Line is indented deeper than the key so we're in a continuation
+				for range blankLineBuffer {
+					currentValue.WriteByte('\n')
```
wbxyz, line 107: This replaces any whitespace lines with simple newlines right? Is that significant?

pkg/ini/ini.go
```diff
@@ -0,0 +1,199 @@
+// Copyright 2025 Google LLC
+// SPDX-License-Identifier: Apache-2.0
+
+// Package ini provides a generic INI file parser that supports multiline values,
```
wbxyz: Let's make it explicit that this parser is targeting setup.cfg files. The dialect we're parsing makes multiple assumptions based on that target, so it's worth being explicit in the top line doc.

If we need ini parsing for other files in the future, we can probably extend or re-use this lib, but it's worth re-evaluating if there are differences when doing so.

pkg/ini/ini_test.go
```diff
@@ -0,0 +1,404 @@
...
+						Name: "section",
+						Values: map[string]string{
+							"description": "This is a long\ndescription that spans\nmultiple lines",
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "comments",
```
wbxyz, line 95: Worth adding a line that includes a comment char without a preceding space?

`key3 = value3a; value3b`

wbxyz (approved):

pkg/ini/ini.go
msuozzo: i think the rest of the description does that but I'll remove "generic" as that's kind of at odds with it.

pkg/ini/ini.go
msuozzo: technically one more map access but ok

pkg/ini/ini.go
msuozzo: yeah that's the observed behavior. not sure what significance that would have here but it is the behavior nonetheless

PR #954, opened 2025-11-17
Allow concurrent local execution

wbxyz:


msuozzo (approved):

algomaster99:
Always had this as a patch :sweat_smile: nice!

wbxyz:
It used to be an important check but one of our build library changes made it unnecessary

PR #955, opened 2025-11-17
Nest docker working dirs in subdir of oss rebuild data

wbxyz:


pkg/build/local/run_executor.go
```diff
@@ -178,7 +178,7 @@ func (e *DockerRunExecutor) executeBuild(ctx context.Context, handle *localHandl
 		return
 	}
 	// Create temporary directory for build output
-	hostOutputPath := filepath.Join(e.tempDirBase, fmt.Sprintf("oss-rebuild-%s", handle.id))
+	hostOutputPath := filepath.Join(e.tempDirBase, "tmp", fmt.Sprintf("oss-rebuild-%s", handle.id))
```
msuozzo, line 181: Not sure about this. What's the motivation?

wbxyz: Ah sent too quickly, agreed that it's not needed.

PR #957, opened 2025-11-18
Add KeepAlive flag to docker run executor

wbxyz:
This keeps "run local" containers available for attaching and
exerimenting even after the build completes.

### DIFF
--- a/pkg/build/local/run_executor.go
+++ b/pkg/build/local/run_executor.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/google/oss-rebuild/internal/bufiox"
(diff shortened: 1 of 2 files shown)


pkg/build/local/run_executor.go
```diff
@@ -224,7 +227,14 @@ func (e *DockerRunExecutor) executeBuild(ctx context.Context, handle *localHandl
 	}
 
 	runArgs = append(runArgs, plan.Image)
-	runArgs = append(runArgs, "/bin/sh", "-c", plan.Script)
+	if e.keepalive {
+		// To keep the container alive, we need to execute the build script in the background and keep an infinte process in the forground.
+		// Write the script to a file then execute it in the background
+		wrappedScript := fmt.Sprintf("cat << 'EOF' > /build.sh\n%s\nEOF\n/bin/sh /build.sh &\ntail -f /dev/null\n", plan.Script)
```
msuozzo, line 240: will the script have EOF? worth checking for it?

wbxyz: I did a cursory check when validating that this works. No guarantee that a future script impl doesn't use heredocs with the same sentinel though. Thoughts on a safer value?

 My original implementation actually used EOF2 to be safe but it looked kind of gross.

msuozzo: maybe just worth check-fail / panic. if we hit it, easy fix. otherwise, pretty painless check to keep around.

PR #963, opened 2025-11-20
Add validate to target encoding

msuozzo:


pkg/rebuild/rebuild/sanitize.go
```diff
@@ -50,11 +55,18 @@ var (
...
+	//   - Reference: https://go.dev/ref/mod#go-mod-file-ident
+	//
 	// GCS Object Names:
 	//   - Technically allows all Unicode, but same restrictions as Windows filesystem
 	//     should be applied for cross-platform compatibility
 	//   - Reference: https://cloud.google.com/storage/docs/objects
-	FilesystemTargetEncoding = &targetEncoding{name: "filesystem"}
+	FilesystemTargetEncoding = &targetEncoding{
+		name:      "filesystem",
+		forbidden: []rune{'/', '\\', ':', '*', '?', '"', '<', '>', '|'},
```
wbxyz, line 68: Should this also prohibit `!`?

msuozzo: eh iirc we said it was suboptimal but not expressly forbidden.

PR #966, opened 2025-11-20
Move stabilizers to dedicated package

msuozzo:


pkg/stabilize/zip.go
```diff
@@ -0,0 +1,139 @@
...
+	Func func(*archive.MutableZipFile)
+}
+
+// Stabilize applies the stabilizer function to the given MutableZipFile.
+func (z ZipEntryStabilizer) Stabilize(arg any) {
+	z.Func(arg.(*archive.MutableZipFile))
+}
+
+// AllZipStabilizers is the list of all available zip stabilizers.
+var AllZipStabilizers = []Stabilizer{
```
wbxyz: nit: I think it would be helpful to move these lists of "AllXStabilizers" to the top of their respective files as documentation of what stabilizers they define.

PR #968, opened 2025-11-21
Add MaxDepth option to diffr

msuozzo:


pkg/diffr/diffr.go
```diff
@@ -18,6 +18,19 @@ import (
...
+type compareContext struct {
+	context.Context
+	maxDepth int
+	depth    int
+}
+
+func (c compareContext) Child() compareContext {
+	c.depth++
+	return c
 }
```
wbxyz, line 34: On first pass, this seems... broken?

Primarily, IIUC this doesn't make a copy of the original, so the original's depth will be incorrectly set to depth+1 after the child is done processing.

Also, (still guessing from the code) sibling nodes will have increasingly large starting depth? IIUC this wasn't caught in the test because the test only has one node at each level of the archive.

msuozzo: It's not using a pointer receiver so `c` itself is a copy

wbxyz: Ahh, my bad. I didn't realize non-pointer receivers make a copy. All makes sense now.

PR #969, opened 2025-11-21
Add format depth metadata to archive.Format

msuozzo:


wbxyz:
Where does this get used? Could we combine the introduction with the first usage?

I checked the following PR (#970) but I don't see any usage there either.

wbxyz:
Ah, it gets used later in #977 

One concern I have is that even if the outer container is .tar.gz or .zip, that doesn't necessarily mean the contents aren't other containers. For example deb packages are the AR format, containing multiple TARs inside. What would you think of setting a reasonable default and just using it everywhere to start?

wbxyz:
> > Ah, it gets used later in #977
> > One concern I have is that even if the outer container is .tar.gz or .zip, that doesn't necessarily mean the contents aren't other containers. For example deb packages are the AR format, containing multiple TARs inside. What would you think of setting a reasonable default and just using it everywhere to start?
> 
> Not sure I understand what you mean by setting a reasonable default. Could we add an archive format for "Deb" which encodes its value (2?)?

It's less a question of supporting .deb files specifically, but more a question of how we handle archives that go beyond this depth without an indication at the file extension (like crates with an archive inside it).

Exploring it more, it seems like stabilizers don't apply to nested archive types currently, so let's file an issue to support archives beyond this depth?

wbxyz (approved):

PR #975, opened 2025-11-24
Add stabilizer docs

msuozzo:


docs/stabilizers.md
```diff
@@ -0,0 +1,190 @@
+# Stabilization
+
+Stabilization transforms software packages to remove observable changes due to
```
wbxyz, line 3: I think it's really important to prioritize in the opening line that we believe these details are minimal in their security relevance (maybe mentioning that it's explained below).

docs/stabilizers.md
```diff
@@ -0,0 +1,190 @@
...
+abstraction for evaluating artifacts' _functional equivalence_.
+
+Strictly speaking, though, any observable difference in program content, no
+matter how trivial, can technically be conditioned upon by an executing system:
+A piece of software _could_ check its own creation timestamp and delete the
+filesystem if it was built on a Tuesday. It _could_ behave differently based on
+whether the builder's UID was 1000 or 1001, or whether gzip level 6 or 9 was
+used. However we assert that the required reflection would require such
+contrived logic to exploit that we should instead look to detect them by other
+means.
```
wbxyz: I thing this is a strong paragraph, but I think we can make an even stronger claim at the end. Apologies, this is kind of wordy, might need a new paragraph.

"We assert that any such stabilized attribute is not a malicious behavior in its own right, but would require reflective code to check the condition and additional malicious payloads to execute. By matching the post-stabilized artifacts bit-for-bit, we know that any such reflection or payload is included in both the rebuild and upstream artifact.

Furthermore, by linking source to binary, OSS Rebuild forces the attacker to include the code for this reflection and payload in the upstream source. This moves the discovery of attacks from the reverse engineering domain into static code analysis of the source, which is a much more tractable problem."

(Very much open to re-wording this)

docs/stabilizers.md
wbxyz: Namely, I don't think "preserving the software's semantics" as the strongest phrasing of how minor most of these stabilization passes are

docs/stabilizers.md
```diff
@@ -0,0 +1,190 @@
...
+```
+
+Each stabilizer is a pure function, handling a small, focused transformation.
+For example, one stabilizer might zeroe timestamps, another normalizes file
+ordering, and so on. Like stabilization, each accepts a single package and
+outputs a single package:
+
+```
+stabilizer(package) → package
+```
```
wbxyz, line 48: It's not really clear what's different between these two example snippets besides one being called "stabilization" and the second being "stabilizer". Can they be combined? Or should stabilization be visualized as applying multiple stabilizers?

docs/stabilizers.md
```diff
@@ -0,0 +1,190 @@
+# Stabilization
+
+Stabilization transforms software packages to remove observable changes due to
+the build environment while preserving the software's semantics. This is the
+foundation of how OSS Rebuild asserts that two packages are equivalent.
+
+## The Equivalence Philosophy
```
wbxyz, line 8: Here, or maybe in the concrete example towards the bottom, is it worth mentioning why we don't fix the problems upstream? Specifically that it's not feasible to address all these reproducibility issues on every CI worker for every package in every ecosystem? I think that's also part of the philosophy is that this process is pragmatic about improving the security of all the packages we're trying to cover.

docs/stabilizers.md
msuozzo: yeah agreed. i added a follow-up that defines stabilization as a composition of stabilizers

docs/stabilizers.md
msuozzo: what about "identity and semantics". idk i don't feel strongly about whether we should put a conclusion in the intro but I've also added a bit about this not being a meaningful regression in security.

docs/stabilizers.md
msuozzo: alright how about this? It's not exhaustive but does shoehorn in the fact that we can and should fix the issues but are addressing problems where we are now.

docs/stabilizers.md
msuozzo: eh idk. so what you're interested in adding here is that it forces the attack into the source code? that feels at least obliquely captured in what's there now that "the exploit would look weird." is there a tweak within the existing text you think you more precisely capture what you're after?

docs/stabilizers.md
wbxyz: Yeah I think "without compromising real security value" is a good additon.

docs/stabilizers.md
wbxyz: Yeah this looks good thanks.

docs/stabilizers.md
wbxyz: Yeah that example definitely makes it much clearer.

docs/stabilizers.md
```diff
@@ -0,0 +1,198 @@
...
+Each stabilizer is a pure function, handling a small, focused transformation.
+For example, one stabilizer might zeroe timestamps, another normalizes file
+ordering, and so on. Like stabilization, each accepts a single package and
+outputs a single package:
+
+```
+stabilizer(package) → package
+```
+
+The full stabilization process is simply the composition of all stabilizers:
```
wbxyz: nit: "The full stabilization process is a composition of multiple stabilizers:"

I don't think we quite want to say "all" stabilizers, some stabilizers don't make sense for certain ecosystems, and some stabilizers might be disabled if they're experimental.

msuozzo: fair. what about "of all applied stabilizers"

wbxyz: Yeah that works. Feels a little circular to define how stabilizers are applied in terms of "all applies stabilizers" but I don't think it probably reads ok.

msuozzo: flip-flopped to "all configured stabilizers"

PR #976, opened 2025-11-25
Increasing support for pyproject.toml discovery and parsing

gtystahl:
This is the cleaned PR from my fork. Still needs a file name change as a test and then potentially removing an old TODO from a previous version of SS-Rebuild.

### DIFF
--- a/pkg/rebuild/pypi/infer.go
+++ b/pkg/rebuild/pypi/infer.go
@@ -257,6 +257,8 @@ func (Rebuilder) InferStrategy(ctx context.Context, t rebuild.Target, mux rebuil
 		if buildReqs, newFoundDir, err := pypiresolver.ExtractAllRequirements(ctx, tree, name, version, dir); err != nil {
 			log.Println(errors.Wrap(err, "Failed to extract reqs from pyproject.toml."))
 		} else {
+			// NOTE - This should NOT overwrite the hint dir if one exists, but utilize it and return it again
+			//   Test "pyproject.toml - Detect package with dir hint" showcases this
 			dir = newFoundDir
 			existing := make(map[string]bool)
 			pkgname := func(req string) string {


pkg/rebuild/pypi/infer.go
```diff
@@ -251,9 +251,10 @@ func (Rebuilder) InferStrategy(ctx context.Context, t rebuild.Target, mux rebuil
 		if err != nil {
 			return cfg, errors.Wrapf(err, "Failed to get tree")
 		}
-		if buildReqs, err := pypiresolver.ExtractAllRequirements(ctx, tree, name, version); err != nil {
```
wbxyz: This works for now, but I think we might want to refactor this once the upcoming PRs are merged.

Previously ExtractAllRequirements was doing just that, but now it's also finding the directory in which the rebuild should take place which is a different job entirely.

We need to be careful we don't end up pushing more inference logic into ExtractAllRequirements just because it's the most convenient.

I think the right way to do it ultimately will be to split this function into two separate steps:
1. the config file search which verifies that a config file matches the expected package and version
1. the build requirements extracting from the config files found in step 1

If we expose those two functions separately, I think infer.go can make its own decision about dir without relying on ExtractAllRequirements.

In the meantime, can you add a comment like

```
// TODO: Split ExtractAllRequirements into 1) file discovery and 2) requirement extraction
// This allows infer.go to own the logic deciding on dir
```

pkg/rebuild/pypi/infer.go
```diff
@@ -251,9 +251,10 @@ func (Rebuilder) InferStrategy(ctx context.Context, t rebuild.Target, mux rebuil
...
+		if buildReqs, newFoundDir, err := pypiresolver.ExtractAllRequirements(ctx, tree, name, version, dir); err != nil {
 			log.Println(errors.Wrap(err, "Failed to extract reqs from pyproject.toml."))
 		} else {
+			dir = newFoundDir
```
wbxyz, line 262: Can you add a note that this might overwrite the dir provided by location hint above?

pkg/rebuild/pypi/parsing/discovery.go
```diff
@@ -128,7 +128,7 @@ func findRecursively(fileType string, tree *object.Tree, name, version string) (
 	var foundFiles []foundFile
 
 	tree.Files().ForEach(func(f *object.File) error {
-		if filepath.Base(f.Name) == fileType {
+		if filepath.Base(f.Name) == fileType && (hintDir == "" || filepath.Dir(f.Name) == hintDir) {
```
wbxyz, line 131: Awesome!

pkg/rebuild/pypi/infer.go
gtystahl: I added functionality in `findRecursively` to search the hint directory only (and not any others) if it is specified on line 131 of `discovery.go`. This way, it should never overwrite the hint provided. I had a test to make sure it doesn't get replaced (lines 122 - 146) which goes with the hinted directory and pyproject even when there is matching project and version information in a different sub directory.

If replacing is the wanted functionality, I can remove that change. Otherwise, feel free to verify my explanation and test.

pkg/rebuild/pypi/infer.go
gtystahl: I have been working on something similar to what you described in the commit finder rework I have been working on. I will add the comment here and add make the commit finder PR next week.

wbxyz: Sounds good.

pkg/rebuild/pypi/infer.go
wbxyz: Addressing in the followup with refactor sgtm.

PR #977, opened 2025-11-25
Migrate custom stabilizer generator to diffr

msuozzo:


tools/ctl/ide/commands/commands.go
```diff
@@ -264,33 +264,42 @@ func NewRebuildCmds(app *tview.Application, executor build.Executor, prebuildCon
...
+					)
+					if err != nil && !errors.Is(err, diffr.ErrNoDiff) {
+						log.Println(errors.Wrap(err, "running diffr"))
+						return
+					}
+					if buf.Len() > 0 {
+						var root diffr.DiffNode
+						if err := json.Unmarshal(buf.Bytes(), &root); err != nil {
+							log.Println(errors.Wrap(err, "parsing diffr output"))
 							return
```
wbxyz, line 300: Is json parsing a common enough usage of Diff() that we should expose this as a convenience function in the diff library?

tools/ctl/ide/commands/commands.go
```diff
@@ -756,3 +764,41 @@ func NewBenchmarkCmds(app *tview.Application, executor build.Executor, prebuildC
 		},
 	}
 }
+
+// collectDiffPaths walks the diff tree and collects file paths by status.
+// It returns files only in first archive, files in both, and files only in second.
+func collectDiffPaths(n *diffr.DiffNode) (onlyFirst, both, onlySecond []string) {
```
wbxyz, line 770: Similarly, should we put this as a helper in the diff library?

msuozzo: It feels bespoke enough that I thought it best to wait until we have another user before making a library of it.

tools/ctl/ide/commands/commands.go
msuozzo: Yeah I think it's probably worth it. Will do so in a follow on

msuozzo: done in https://github.com/google/oss-rebuild/pull/988

tools/ctl/ide/commands/commands.go
msuozzo: will wait for an additional use-case before abstracting it

PR #979, opened 2025-11-26
Setup cfg implementation

gtystahl:
Adding another build file (setup.cfg) implementation. Followed the same principles as pyproject with testing and implementation. However, relies on changes implemented in [PR](https://github.com/google/oss-rebuild/pull/976) so acceptance of that one should be priority.

### DIFF
--- a/pkg/rebuild/pypi/parsing/setupcfg.go
+++ b/pkg/rebuild/pypi/parsing/setupcfg.go
@@ -62,16 +62,10 @@ func verifySetupCfgFile(ctx context.Context, found foundFile, name, version stri
 
 		if editDist == 0 {
 			verificationResult.nameMatch = true
+		}
 
-			if fv && version == foundVersion {
-				verificationResult.versionMatch = true
-			}
-		} else {
-			verificationResult.partialNameMatch = true
-
-			if fv && version == foundVersion {
-				verificationResult.partialVersionMatch = true
-			}
+		if fv && version == foundVersion {
+			verificationResult.versionMatch = true
 		}
 	}
 
--- a/pkg/rebuild/pypi/parsing/discovery.go
+++ b/pkg/rebuild/pypi/parsing/discovery.go
@@ -26,13 +26,11 @@ type foundFile struct {
 }
 
 type fileVerification struct {
-	foundF              foundFile
-	main                bool
-	nameMatch           bool
-	versionMatch        bool
-	partialNameMatch    bool
-	partialVersionMatch bool
-	levDistance         int
+	foundF       foundFile
+	main         bool
+	nameMatch    bool
+	versionMatch bool
+	levDistance  int
 }
 
 // minEditDistance computes the Levenshtein distance between two strings.
@@ -68,15 +66,16 @@ func minEditDistance(s1, s2 string) int {
 // verificationScore assigns a numeric priority to the verification object.
 // Higher score means higher priority (comes first).
 func verificationScore(v fileVerification) int {
-	if v.versionMatch {
+	if v.nameMatch && v.versionMatch {
+		return 4
+	} else if v.nameMatch {
 		return 3
-	}
-	if v.nameMatch {
+	} else if v.main {
 		return 2
-	}
-	if v.main {
+	} else if v.versionMatch {
 		return 1
 	}
+
 	return 0
 }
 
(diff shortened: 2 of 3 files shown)


pkg/rebuild/pypi/parsing/setupcfg.go
```diff
@@ -0,0 +1,101 @@
...
+	for _, line := range lines {
+		line = strings.TrimSpace(line)
+		if line != "" {
+			result = append(result, line)
+		}
+	}
+	return result
+}
+
+func verifySetupCfgFile(ctx context.Context, foundFile foundFile, name, version string) (fileVerification, error) {
```
wbxyz: Instead of `foundFile foundFile` wdyt of `found foundFile`?

It's good practice to avoid using a variable name that overshaddows and existing value, which can make it confusing when reading code to know which value is being referred to.

pkg/rebuild/pypi/parsing/setupcfg.go
```diff
@@ -0,0 +1,101 @@
...
+	"context"
+	"log"
+	"strings"
+
+	"github.com/go-git/go-git/v5/plumbing/object"
+	"github.com/google/oss-rebuild/pkg/ini"
+	"github.com/pkg/errors"
+)
+
+func parseMultiLineValue(value string) []string {
```
wbxyz: nit: splitVersions?

It's not always multiline (sometimes list-semi) and it's specific to version fields IIUC

pkg/rebuild/pypi/parsing/setupcfg.go
```diff
@@ -0,0 +1,101 @@
...
+			verificationResult.nameMatch = true
+
+			if fv && version == foundVersion {
+				verificationResult.versionMatch = true
+			}
+		} else {
+			verificationResult.partialNameMatch = true
+
+			if fv && version == foundVersion {
+				verificationResult.partialVersionMatch = true
```
wbxyz: Should this be a regular versionMatch? Because they're exact?

pkg/rebuild/pypi/parsing/setupcfg.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	setupRequires, found := cfg.GetValue("options", "setup_requires")
+	if found {
+		setupRequiresList := parseMultiLineValue(setupRequires)
+		reqs = append(reqs, setupRequiresList...)
+	}
+
+	log.Println("Added these reqs from setup.cfg: " + strings.Join(reqs, ", "))
+	return reqs, nil
```
wbxyz: I think you can skip the "found" check, setupRequires will just be an empty string if it's not found.

```golang
setupRequires, _ := cfg.GetValue("options", "setup_requires")
reqs := parseMultiLineValue(setupRequires)
log.Println("Added these reqs from setup.cfg: " + strings.Join(reqs, ", "))
return reqs, nil
```

pkg/rebuild/pypi/parsing/setupcfg.go
```diff
@@ -0,0 +1,101 @@
...
+				verificationResult.partialVersionMatch = true
+			}
+		}
+	}
+
+	return verificationResult, nil
+}
+
+func extractSetupCfgRequirements(ctx context.Context, f *object.File) ([]string, error) {
+	var reqs []string
```
wbxyz: nit: Does "reqs" need to be defined so early? Why not define it as the return of the parsing call?

pkg/rebuild/pypi/parsing/setupcfg.go
gtystahl: I did partial here to mirror the partial name match condition above it. I didn't want it to overshadow the other matches just because of a weird case where the versions are the same across all sub packages in a repo (e.g. all v1)

pkg/rebuild/pypi/parsing/setupcfg.go
gtystahl: changed

pkg/rebuild/pypi/parsing/setupcfg.go
gtystahl: I changed it to splitRequiresList to be even more specific since it will also include the dependency name with the version

pkg/rebuild/pypi/parsing/setupcfg.go
gtystahl: Check skipped now.

pkg/rebuild/pypi/parsing/setupcfg.go
gtystahl: Changed to the return of the parse.

PR #981, opened 2025-11-26
Add attestation update API

msuozzo:
This adds versioning to the attestation GCS bucket, an initial list of
changes to consider meaningful for updating existing attestations, and
adding an explicit option to trigger overwriting on the rebuild API.

### DIFF
--- a/cmd/api/main.go
+++ b/cmd/api/main.go
@@ -69,20 +69,27 @@ var (
 
 var httpcfg = httpegress.Config{}
 
-func makeKMSSigner(ctx context.Context, cryptoKeyVersion string) (*dsse.EnvelopeSigner, error) {
+func makeKMSSignerVerifier(ctx context.Context, cryptoKeyVersion string) (*dsse.EnvelopeSigner, *dsse.EnvelopeVerifier, error) {
 	kc, err := kms.NewKeyManagementClient(ctx)
 	if err != nil {
-		return nil, errors.Wrap(err, "creating KMS client")
+		return nil, nil, errors.Wrap(err, "creating KMS client")
 	}
-	kmsSigner, err := kmsdsse.NewCloudKMSSignerVerifier(ctx, kc, cryptoKeyVersion)
+	kmsSignerVerifier, err := kmsdsse.NewCloudKMSSignerVerifier(ctx, kc, cryptoKeyVersion)
 	if err != nil {
-		return nil, errors.Wrap(err, "creating Cloud KMS signer")
+		return nil, nil, errors.Wrap(err, "creating Cloud KMS signer/verifier")
 	}
-	dsseSigner, err := dsse.NewEnvelopeSigner(kmsSigner)
+	dsseSigner, err := dsse.NewEnvelopeSigner(kmsSignerVerifier)
 	if err != nil {
-		return nil, errors.Wrap(err, "creating envelope signer")
+		return nil, nil, errors.Wrap(err, "creating envelope signer")
 	}
-	return dsseSigner, nil
+	// Create verifiers for both new (gcpkms://) and legacy (https://) keyid formats
+	// to support verification of existing attestations with legacy keyids
+	legacyVerifier := kmsdsse.NewLegacyKeyIDVerifier(kmsSignerVerifier)
+	dsseVerifier, err := dsse.NewEnvelopeVerifier(kmsSignerVerifier, legacyVerifier)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "creating envelope verifier")
+	}
+	return dsseSigner, dsseVerifier, nil
 }
 
 func RebuildPackageInit(ctx context.Context) (*apiservice.RebuildPackageDeps, error) {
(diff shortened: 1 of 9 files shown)


internal/api/apiservice/rebuild.go
```diff
@@ -283,14 +285,64 @@ func rebuildPackage(ctx context.Context, req schema.RebuildPackageRequest, deps
...
+			if err != nil {
+				v.Message = errors.Wrap(err, "parsing attestation bundle").Error()
+				return &v, nil
+			}
+			rebuildAtt, err := attestation.FilterForOne[attestation.RebuildAttestation](
+				bundle,
+				attestation.WithBuildType(attestation.BuildTypeRebuildV01),
+			)
+			if err != nil {
+				v.Message = api.AsStatus(codes.FailedPrecondition, errors.Wrap(err, "overwrite denied: no rebuild attestation in bundle")).Error()
```
wbxyz: Why is this constructing a Status if it's just going to be a string? And why do the other attestation parsing errors not warrant a FailedPrecondition?

internal/api/apiservice/rebuild.go
```diff
@@ -283,14 +285,64 @@ func rebuildPackage(ctx context.Context, req schema.RebuildPackageRequest, deps
...
+		default:
+			allow = false
+		}
+		if !allow {
 			v.Message = api.AsStatus(codes.AlreadyExists, errors.New("conflict with existing attestation bundle")).Error()
 			return &v, nil
 		}
+		deps.OverwriteAttestations = true
+	} else {
+		deps.OverwriteAttestations = true
```
wbxyz: This means a bundle doesn't exist, right? If so, then maybe OverwriteAttestations should be false?

wbxyz: Is there a way to avoid having two separate fields passed into buildAndAttest that allow overwrites? What if this processing block just sets the overwrite mode used by buildAndAttest and we delete the deps.OverwriteAttestations field?

If we don't think an overwrite is going to be needed, then the overwrite mode can be set to empty?

internal/verifier/attestation.go
```diff
@@ -122,10 +122,11 @@ func CreateAttestations(ctx context.Context, t rebuild.Target, defn *schema.Buil
...
 	externalParams := attestation.RebuildParams{
-		Ecosystem: string(t.Ecosystem),
-		Package:   t.Package,
-		Version:   t.Version,
-		Artifact:  t.Artifact,
+		Ecosystem:     string(t.Ecosystem),
+		Package:       t.Package,
+		Version:       t.Version,
+		Artifact:      t.Artifact,
+		OverwriteMode: mode,
```
wbxyz, line 129: Related to the comment about buildAndAttest, I don't think this necessarily indicates whether an overwrite happened. If a previous bundle doesn't exist, then the OverwriteMode would still just be plumed through to here. Is that correct? If so, is that intended?

pkg/attestation/types.go
```diff
@@ -71,6 +72,8 @@ type RebuildParams struct {
 	Package string `json:"package"`
 	// Version is the specific version of the package to rebuild
 	Version string `json:"version"`
+	// OverwriteMode specifies the justification for overwriting an existing attestation
+	OverwriteMode schema.OverwriteMode `json:"overwrite_mode,omitempty"`
```
wbxyz, line 76: I'm still not really sure what the intended use case for including the OverwriteMode in the attestation is. Do we expect pubsub subscribers to make a decision about whether they're going to replace existing attestations based on this signal? What semantic difference would they expect between attestations caused by service updates vs forced?

internal/api/apiservice/rebuild.go
msuozzo: yeah I'll clean this up in a later change. right now, it's not actively breaking anything since we set it to false.

msuozzo: added a comment. it's to prevent a failure when 2 simultaneous rebuilds of the same package are run.

internal/api/apiservice/rebuild.go
msuozzo: status encodes the HTTP-convertible status code so it's not just a 503.

pkg/attestation/types.go
msuozzo: resolved offline. it's a way of indicating in-band that there is a previous version of the attestation. it's not much data and this allows us to avoid designing/communicating an additional attestation-specific construct and can just go with the same format as in the request.

internal/verifier/attestation.go
msuozzo: it will be empty and it's omitempty so it will be absent for first-writes

internal/api/apiservice/rebuild.go
msuozzo: added a TODO

cmd/api/main.go
wbxyz, line 1: Are these verifier changes supposed to be in a different PR?

internal/api/apiservice/rebuild.go
msuozzo: okay so finally dug into this. added a comment: currently, the "write to Message" idiom should really be across a stub boundary since it's suppressing what are really rpc errors in the cases where we're confident we can write out a firestore record. as such, we should optimistically keep these in api format until we refactor to actually be a separate (cleaner) stub.

cmd/api/main.go
msuozzo: yep for sure. done

internal/verifier/attestation.go
msuozzo: added some logic to clean up the edge cases when the bundle doesn't exist but overwrites were requested.

PR #982, opened 2025-12-01
Support attestation overwrite in ctl

msuozzo:


tools/ctl/ctl.go
```diff
@@ -1635,6 +1637,7 @@ var (
 	bootstrapVersion  = flag.String("bootstrap-version", "", "the version of bootstrap tools to use")
 	useNetworkProxy   = flag.Bool("use-network-proxy", false, "request the newtwork proxy")
 	useSyscallMonitor = flag.Bool("use-syscall-monitor", false, "request syscall monitoring")
+	overwriteMode     = flag.String("overwrite-mode", "", "reason to overwrite existing attestation (SERVICE_UPDATE or FORCE)")
```
wbxyz: Is it worth validating that the flag value is one of these in the functions that use it?

msuozzo: yeah i think so. turns out the API doesn't validate and interprets unknown values as default no-overwrite

msuozzo: done

PR #985, opened 2025-12-03
Make a new pkg/act for internal/api types

msuozzo:


pkg/act/api/http.go
```diff
@@ -0,0 +1,221 @@
...
+	codes.Internal:           http.StatusInternalServerError,
+	codes.Unavailable:        http.StatusServiceUnavailable,
+	codes.DataLoss:           http.StatusInternalServerError,
+	codes.Unauthenticated:    http.StatusUnauthorized,
+}
+
+func Handler[I act.Input, O any, D act.Deps](initDeps InitDeps[D], handler HandlerFunc[I, O, D]) http.HandlerFunc {
+	return func(rw http.ResponseWriter, r *http.Request) {
+		ctx := context.Background()
+		r.ParseForm()
```
wbxyz, line 141: Should we check the err on this call? Maybe in a followup, I know this is largely a refactor PR.

wbxyz (approved):

pkg/act/api/http.go
msuozzo: potentially. this is meant to be an in-place move, though, so this should certainly be a follow-on

PR #986, opened 2025-12-03
Add pkg/act/cli for using act with cobra

msuozzo:


pkg/act/cli/io.go
```diff
@@ -0,0 +1,13 @@
...
+// Package cli provides utilities for building CLI commands using the act framework.
+package cli
+
+import "io"
+
+// IO provides input/output streams for CLI commands.
+type IO struct {
+	Out io.Writer // stdout
+	Err io.Writer // stderr
+}
```
wbxyz, line 14: Can we add stdin for completeness?

pkg/act/cli/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+### 4. Create the Command Factory
+
+Wire everything together:
+
+```go
+func Command() *cobra.Command {
+    cfg := Config{}
+    cmd := &cobra.Command{
+        Use:   "my-command --repository <uri> --ref <ref>",
```
wbxyz, line 84: Is it possible to generate the flag list here using the flagset below? One problem I've noticed is that when we add a new flag we forget to update the usage docs.

wbxyz (approved):

pkg/act/cli/README.md
msuozzo: nope :/

wbxyz: Can't we use set.VisitAll() after building the flag set but before returning the command? https://cs.opensource.google/go/go/+/refs/tags/go1.25.5:src/flag/flag.go;l=456

msuozzo: Oh sure we can implement a function that does it but we can also use this as a way of eliding common options or separating mutually dependent sets of arguments. This is informational and tbh superfluous so I don't think we should spend too much effort on it.

PR #1046, opened 2026-01-14
Redesign stabilizer API

msuozzo:


pkg/stabilize/context.go
```diff
@@ -0,0 +1,51 @@
...
+	Entry  *EntryInfo
+}
+
+// ArchiveLevel represents a single level in a nested archive.
+type ArchiveLevel struct {
+	Format archive.Format
+	Path   string
+}
+
+// EntryInfo provides information about the current entry being stabilized.
```
wbxyz, line 22: This is relative to the last ArchiveLevel struct in the stack?

pkg/stabilize/custom.go
```diff
@@ -121,42 +117,36 @@ func (rp *ReplacePattern) Validate() error {
...
-				}
-				transformed := re.ReplaceAll(content, []byte(rp.Replace))
-				zf.SetContent(transformed)
-			},
-		}, nil
-	default:
-		return nil, errors.New("unsupported format")
-	}
+	tarfn := TarEntryFn(func(te *archive.TarEntry) {
+		if match, err := multiMatch(rp.Paths, te.Name); err != nil || !match {
```
wbxyz, line 124: Should these matches include the context's chain of outer archive paths?

wbxyz (approved):

pkg/stabilize/context.go
msuozzo: yeah the entry within the last ArchiveLevel

pkg/stabilize/custom.go
msuozzo: i think it's fine to limit support to the current entry for now. we'd want to design and document a syntax for specifying layers of nesting here that we don't do at the moment. i think we can leave that for a future feature addition, if we run into the need for it.

pkg/stabilize/context.go
wbxyz: Could you add a comment on the Path field to clarify that? Maybe something like "Path within the immediate archive, excluding any outer archive elements"

pkg/stabilize/custom.go
wbxyz: Worth adding a comment?

wbxyz (approved):

PR #1170, opened 2026-04-09
Implement custom file patching in the docker proxy

andusy:
Many tools and ecosystems (such as Bazel, apt, etc.) require special config files to be written to the filesystem to use custom CA certificates. This change adds a generic way to write arbitrary files to a Docker container so downstream consumers of this package can adapt to new ecosystems by passing in the filepaths and contents.

### DIFF
--- a/pkg/proxy/docker/docker.go
+++ b/pkg/proxy/docker/docker.go
@@ -407,6 +407,19 @@ type patchSet struct {
 	Patches []patch
 }
 
+type FileWriteMode int
+
+const (
+	OverwriteExisting FileWriteMode = iota
+	AppendToExisting
+	ErrorOnExisting
+)
+
+type CustomFile struct {
+	Content []byte
+	Mode    FileWriteMode
+}
+
 // ContainerTruststorePatcher provides a Docker API proxy that patches the container truststore while running.
 type ContainerTruststorePatcher struct {
 	cert                 x509.Certificate
@@ -414,7 +427,7 @@ type ContainerTruststorePatcher struct {
 	truststoreEnvVars    []string
 	javaTruststoreEnvVar bool
 	bazelTruststore      bool
-	customFiles          map[string][]byte
+	customFiles          map[string]CustomFile
 	networkOverride      string // TODO: Not a good fit for this abstraction
 	proxySocket          string
 	patchMap             map[string]*patchSet
@@ -428,7 +441,7 @@ type ContainerTruststorePatcherOpts struct {
 	TruststoreEnvVars    []string
 	JavaTruststoreEnvVar bool
 	BazelTruststore      bool
-	CustomFiles          map[string][]byte
+	CustomFiles          map[string]CustomFile
 	RecursiveProxy       bool
 	NetworkOverride      string
 }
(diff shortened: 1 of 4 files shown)


pkg/proxy/docker/docker.go
```diff
@@ -663,6 +666,34 @@ func (d *ContainerTruststorePatcher) proxyRequest(clientConn, serverConn net.Con
...
+						customFilesFailed = true
+						break
+					}
+				} else {
+					log.Printf("Reading custom file %s: %v", path, err)
+					customFilesFailed = true
+					break
+				}
+			} else {
+				f.Contents = append(f.Contents[:], content...)
```
msuozzo: hmm i'm skeptical of this behavior as a default. do we want this to always "append"? i can imagine file types where that could cause unexpected or corrupting behavior. not sure how we want to change the interface but i suspect we want some configurability by location.

pkg/proxy/docker/docker.go
```diff
@@ -663,6 +666,34 @@ func (d *ContainerTruststorePatcher) proxyRequest(clientConn, serverConn net.Con
 				}
 			}
 		}
+		customFilesFailed := false
```
msuozzo, line 683: we also landed full tests under e2e/ so you could consider adding a case there to exercise this new behavior.

pkg/proxy/docker/docker.go
andusy: Good point. What about letting the user choose what to do with the file if it exists? Something like

```
type FileWriteMode int

const (
    Overwrite FileWriteMode = iota // Overwrite the file if it exists
    Append                         // Append to the file if it exists
    ErrorIfExists                  // Fail if the file already exists
)

type CustomFile struct {
    Content []byte
    Mode    FileWriteMode
}

type ContainerTruststorePatcherOpts struct {
	...
	CustomFiles          map[string]CustomFile
    ...
}
```

msuozzo: yeah sounds good. naming nit: what about OverwriteExisting, AppendToExisting, ErrorOnExisting?

andusy: Done.

pkg/proxy/docker/docker.go
andusy: I added e2e tests for all the write modes we're supporting. Also added a flag to main.go so that we can pass this option to the proxy.

msuozzo: ah i meant the test in ./e2e. the unit tests are good but, as you saw, a bit cumbersome. this works, though.

pkg/proxy/docker/docker.go
```diff
@@ -663,6 +679,46 @@ func (d *ContainerTruststorePatcher) proxyRequest(clientConn, serverConn net.Con
...
+		for path, file := range d.customFiles {
+			f, err := dfs.OpenAndResolve(path)
+			if err != nil {
+				if err == iofs.ErrNotExist {
+					if err := createFile(dfs, file.Content, path); err != nil {
```
msuozzo, line 688: i guess the only thing I'm still hesitant about is that this is fixed content string might be less valuable than supporting some sort of templating with the cert paths (that could support the bazel case above). but still i don't think that should block merging, just maybe a good idea to include as a TODO for us to come back to later.

msuozzo (approved):

pkg/proxy/docker/docker.go
andusy: Added a TODO for this

PR #1176, opened 2026-04-13
Make RebuildAttempt stateful

msuozzo:
This change allows us to track pending attempts out-of-band of the
request. This helps with the implementation of a long-running operation
API as we can use the attempt object itself to represent the incomplete
state in the Operation.


spencerschrock (approved):
I'll defer to William on the rundex bits, but other parts LGTM

wbxyz (approved):

PR #1177, opened 2026-04-14
Implement Resource abstraction for attempts, runs

msuozzo:


internal/db/db.go
```diff
@@ -0,0 +1,128 @@
...
+
+	"cloud.google.com/go/firestore"
+	"github.com/google/oss-rebuild/pkg/rebuild/schema"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// Resource is a typed CRUD handle for a single Firestore collection (or
+// nested-collection path). T is the value type; K is the primary key type.
+type Resource[T, K any] interface {
```
spencerschrock, line 20: This is a nit, but can we use order the types `K`, and then `V` throughout this file? It's a bit jarring to see it the other way around

spencerschrock (approved):
Seems like a nice abstraction, for other local DBs and testing

internal/db/db.go
msuozzo: yeah I can see the confusion. I guess I view this K as more of an afterthought with T being the main param on Resource. I'd initially considered breaking out cases where we need separate K values into a separate generic but ultimately decided against it. it would be nice if there were a generic default as we'll generally want K to just be a string and we could still but having the option of a structured type here makes things a bit cleaner. I guess that may not really answer your question but could tweaking the comment here clear things up?

spencerschrock: It mainly just looked weird on the aliases you defined, but I wont block anything

```

	Runs     = Resource[schema.Run, string]
	Attempts = Resource[schema.RebuildAttempt, AttemptKey]
```

msuozzo: yeah fair. I think i'll leave it for now but probably good to revisit.

wbxyz (approved):

PR #1183, opened 2026-04-16
Set up ruby toolchain

msuozzo:


pkg/build/baseimage.go
```diff
@@ -21,8 +21,9 @@ func DefaultBaseImageConfig() BaseImageConfig {
 	return BaseImageConfig{
 		Default: "docker.io/library/alpine:3.21",
 		Ecosystems: map[rebuild.Ecosystem]string{
-			rebuild.Debian: "docker.io/library/debian:stable-20251103-slim",
-			rebuild.Maven:  "docker.io/library/debian:trixie-20250203-slim",
+			rebuild.Debian:   "docker.io/library/debian:stable-20251103-slim",
+			rebuild.Maven:    "docker.io/library/debian:trixie-20250203-slim",
+			rebuild.RubyGems: "docker.io/library/debian:trixie-20250203-slim",
```
wbxyz: nit: Can we use the stable variant? Or is trixie necessary? I'd like to avoid adding a new dependency on a 1-year old image (before trixie became stable?) if a newer stable version works.

wbxyz (approved):

pkg/build/baseimage.go
msuozzo: good call. done

PR #1185, opened 2026-04-16
Add timewarp support for Ruby

msuozzo:


internal/timewarp/timewarp.go
```diff
@@ -288,6 +322,29 @@ func (h Handler) handleRequest(rw http.ResponseWriter, r *http.Request) error {
 	return nil
 }
 
+// proxyUpstream forwards a request to the upstream registry without time filtering.
+func (h Handler) proxyUpstream(rw http.ResponseWriter, r *http.Request) error {
```
wbxyz, line 326: What about using httputil.ReversProxy?

https://pkg.go.dev/net/http/httputil#ReverseProxy

pkg/rebuild/rubygems/strategy.go
```diff
@@ -41,6 +49,14 @@ func (b *GemBuild) ToWorkflow() *rebuild.WorkflowStrategy {
...
 		})
 	}
+	if registryTime != "" {
+		deps = append(deps, flow.Step{
+			Uses: "rubygems/setup-registry",
+			With: map[string]string{
+				"registryTime": registryTime,
+			},
+		})
+	}
```
wbxyz, line 59: Worth adding a test for this?

wbxyz (approved):

wbxyz:
Not really in scope for this PR, but the timewarp.go file is starting to get a lot of interleaved ecosystem-specific logic.

In a followup, it might be worth trying to have a switch at the top of timewarp and routing the implementation to ecosystem-specific handlers.

internal/timewarp/timewarp.go
msuozzo: We'd need h.Client.Do to somehow be wrapped up in a http.Transport and we'd lose the herror pattern (ReverseProxy writes the error to the rw directly). This is shim pretty modest by comparison.

pkg/rebuild/rubygems/strategy.go
msuozzo: yep good call. done.

msuozzo:
> Not really in scope for this PR, but the timewarp.go file is starting to get a lot of interleaved ecosystem-specific logic.
> 
> In a followup, it might be worth trying to have a switch at the top of timewarp and routing the implementation to ecosystem-specific handlers.

yeah agreed

PR #1187, opened 2026-04-16
Add npm filesystem sanitizer

wbxyz:


msuozzo:
This changes the output attestation path so I think we need more of a plan for rollout.

wbxyz:
Yeah I think we'd need to do a migration and also ensure the analyzers are using stabilzers to decode the storage object events. This change is noteworthy in context of #1191 which uses the stabilized target for path elements in a web-based dashboard.

One alternative would be to add an entirely new flavor of stabilization for http path targets?

PR #1189, opened 2026-04-16
Change Translator to operate on request instead of io.Reader

wbxyz:


msuozzo (approved):
Yeah this is a lot better

PR #1190, opened 2026-04-16
Add template-based http act handler

wbxyz:


internal/api/rpc.go
```diff
@@ -50,6 +51,10 @@ func Handler[I act.Input, O any, D act.Deps](initDeps InitT[D], handler HandlerT
 	return actapi.Handler(initDeps, handler)
 }
 
+func TmplHandler[I act.Input, O any, D act.Deps](initDeps InitT[D], handler HandlerT[I, O, D], tmpl *template.Template) http.HandlerFunc {
```
msuozzo: I'm cool with this but can we call these `HTMLHandler`?

msuozzo (approved):

internal/api/rpc.go
wbxyz: Yeah sgtm.

pkg/act/api/http.go
```diff
@@ -135,7 +136,31 @@ var grpcToHTTP = map[codes.Code]int{
...
+	return func(rw http.ResponseWriter, o *O) error {
+		if o != nil {
+			return tmpl.Execute(rw, o)
+		}
+		return nil
+	}
+}
+
 func Handler[I act.Input, O any, D act.Deps](initDeps InitDeps[D], handler HandlerFunc[I, O, D]) http.HandlerFunc {
+	return handUsingResponder(initDeps, handler, jsonResponse)
```
msuozzo: ah and i think this should be "handleUsingResponder"

msuozzo (approved):

pkg/act/api/http.go
wbxyz: Yeah good catch

PR #1194, opened 2026-04-17
Refactor dashboard impls to use dedicated data types

wbxyz:


internal/api/dashboard/attempt.go
```diff
@@ -0,0 +1,66 @@
...
+type AttemptData struct {
+	Ecosystem       string
+	PackageName     string
+	EncodedPackage  string
+	Attempt         *RebuildView
+	AttemptStrategy string
+	AttemptDuration string
+}
+
+func Attempt(ctx context.Context, req AttemptRequest, deps *Deps) (*AttemptData, error) {
```
msuozzo, line 35: idle thought: we should be type-asserting these to their api.Handler types more regularly. no need here but might be a good idea for a follow-on across the depot

internal/api/dashboard/dashboard.go
```diff
@@ -50,6 +41,8 @@ func init() {
 	LogsTmpl = template.Must(template.New("logs").Parse(logsHTML))
 }
 
+var encoding = rebuild.FilesystemTargetEncoding
```
msuozzo, line 44: this tripped me up since it looked like a package name. maybe `var packagePathEncoding` or something descriptive like that?

internal/api/dashboard/index.go
```diff
@@ -0,0 +1,138 @@
...
+
+type IndexData struct {
+	BenchmarkTitle   string
+	BenchmarkStats   BenchmarkStats
+	BenchmarkTargets []BenchmarkTarget
+	RecentRebuilds   []RebuildView
+}
+
+func Index(ctx context.Context, _ IndexRequest, deps *Deps) (*IndexData, error) {
+	ctx, cancel := context.WithTimeout(ctx, 60*time.Second)
```
msuozzo, line 46: configure via deps?

internal/api/dashboard/package.go
```diff
@@ -0,0 +1,63 @@
...
+	Ecosystem string
+	Package   string
+}
+
+func (PackageRequest) Validate() error { return nil }
+
+type PackageData struct {
+	Ecosystem       string
+	PackageName     string
+	EncodedPackage  string
```
msuozzo, line 26: non-blocking: would be nice to have these be a single value instead.  could make a value type (or just use Target) with a method exposing the encoded form. definitely overkill at this point but just frustrating that the same output carries both forms.

msuozzo (approved):

msuozzo:
yeah this looks a lot better

internal/api/dashboard/dashboard.go
wbxyz: Yeah that's pretty reasonable, especially crossing file boundaries.

internal/api/dashboard/attempt.go
wbxyz: Easy enough, for this package at least. Added.

internal/api/dashboard/index.go
wbxyz: I think we could actually make an api helper that wraps Handlers with a timeout so this can be configured statically at the same location the route is configured. See https://github.com/google/oss-rebuild/pull/1236

msuozzo: Why is this a property inherent to the endpoint? I'm not a huge fan of building this sort of thing into the interface but api.WithTimeout could be way of abstracting it. But why not configure it via deps?

wbxyz: > I'm not a huge fan of building this sort of thing into the interface

wdym? api.WithTimeout is just a helper that wraps a provided handler to set the ctx's timeout, so no change has to happen on the api interface. Context is one (the canonical?) way of configuring the timeout for an operation, adding a field on deps feels a bit like duplication.

Wrapping the handler with api.WithTimeout would happen at the same location that dep initialization is configured so any caller that controls deps would also have the option to set context timeout.

internal/api/dashboard/package.go
wbxyz: That's what we do in the RebuildView:

```golang
type RebuildView struct {
	rundex.Rebuild
	Encoded rebuild.EncodedTarget
}
```

The problem here is that PackageDeps doesn't refer to a specific target, but really a group of targets. Even if we exposed the encoder as a function in the template evaluation, we'd need to represent these things as targets which feels wrong.

We could definitely refine this further if we wanted, but for now I think it's a good improvement.

internal/api/dashboard/index.go
wbxyz: Let's move this conversation over to https://github.com/google/oss-rebuild/pull/1236 instead, I think the other refactors here are worthwhile in the meantime.

PR #1198, opened 2026-04-20
Remove obsolete context vars

msuozzo:


wbxyz:
Maybe we should set a context timeout here? IIUC the GCB executor responds to context cancellation appropriately. We can read the ExecutionHint at this point and decide on 55min or maybe 23hr timeouts?

wbxyz (approved):

msuozzo:
> Maybe we should set a context timeout here? IIUC the GCB executor responds to context cancellation appropriately. We can read the ExecutionHint at this point and decide on 55min or maybe 23hr timeouts?

eh i think doing so at a higher level of abstraction would work. i believe cancellation in cloud run should work but we'll likely want to add a constraint like this to the cloud run jobs path.

msuozzo:
> > Maybe we should set a context timeout here? IIUC the GCB executor responds to context cancellation appropriately. We can read the ExecutionHint at this point and decide on 55min or maybe 23hr timeouts?
> 
> eh i think doing so at a higher level of abstraction would work. i believe cancellation in cloud run should work but we'll likely want to add a constraint like this to the cloud run jobs path.

added in #1206

PR #1199, opened 2026-04-20
Add SizeHint to RebuildRequest

wbxyz:
This also adds the jumbo worker pool to GCB. Other builders don't do
anything special for JumboSize builds yet, but the information should be
available in the build options and plan options.

### DIFF
--- a/pkg/build/gcb/executor.go
+++ b/pkg/build/gcb/executor.go
@@ -18,6 +18,7 @@ import (
 	"github.com/google/oss-rebuild/pkg/build"
 	"github.com/google/oss-rebuild/pkg/gcb"
 	"github.com/google/oss-rebuild/pkg/rebuild/rebuild"
+	"github.com/google/oss-rebuild/pkg/rebuild/schema"
 	"github.com/pkg/errors"
 	"google.golang.org/api/cloudbuild/v1"
 	"google.golang.org/grpc/codes"
@@ -35,6 +36,7 @@ type Executor struct {
 	serviceAccount     string
 	logsBucket         string
 	privatePool        *gcb.PrivatePoolConfig
+	jumboPool          *gcb.PrivatePoolConfig
 	outputBufferSize   int
 	builderName        string
 	terminateOnTimeout bool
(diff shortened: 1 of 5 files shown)


pkg/build/local/run_executor.go
msuozzo, line 1: can we add to the build.Options instead? i think layering-wise, plan is meant to represent the build instructions / process while the options are the scheduling / execution and i think this falls squarely into the latter case.

msuozzo: ah wait you've added it to both places. i think we should configure the conditional override outside of makeBuild as we do for timeout instead of threading it through the plan

pkg/build/gcb/executor.go
```diff
@@ -305,8 +306,17 @@ func (e *Executor) makeBuild(t rebuild.Target, plan *Plan) *cloudbuild.Build {
 		Logging: "GCS_ONLY",
 	}
 	if e.privatePool != nil {
-		buildOptions.Pool = &cloudbuild.PoolOption{
-			Name: e.privatePool.Name,
+		if plan.JumboPool {
+			buildOptions.Pool = &cloudbuild.PoolOption{
+				// We assume there's a similarly named -jumbo pool.
+				// TODO: Check this pool exists when the Executor is started.
```
msuozzo: yeah we could add a PrivatePoolJumbo config item to the executor here and conditionally route to it if the hint is Jumbo.

pkg/build/gcb/executor.go
```diff
@@ -123,6 +127,11 @@ func (e *Executor) Start(ctx context.Context, input rebuild.Input, opts build.Op
 	if opts.Timeout > 0 {
 		gcbBuild.Timeout = fmt.Sprintf("%ds", int(opts.Timeout.Seconds()))
 	}
+	if opts.SizeHint == schema.JumboSize && e.jumboPool != nil {
```
msuozzo: i think it might be better to move the pool selection out of makeBuild at this point. that way we could keep the logic here and it'd have a more similar structure to what you initially had here.

msuozzo (approved):

pkg/build/gcb/executor.go
wbxyz: I think at that point it makes inline the rest of makeBuild logic into Executor.Start which is relatively minimal. That way we can construct the build object all at once.

pkg/build/gcb/executor.go
wbxyz: Executor now has a `jumboPool` config item, so the JumboSize is only honored if there's a jumboPool configured.

pkg/build/local/run_executor.go
wbxyz: Removed it. In the future I do think it's appropriate for the planner to be aware of build size hints, which would allow size tradeoff decisions during planning. Now that we've inlined the makeBuild logic though it's not immediately necessary.

PR #1200, opened 2026-04-20
Update buildAndAttest to take build.Options

wbxyz:


internal/api/apiservice/rebuild.go
```diff
@@ -196,21 +196,15 @@ func buildAndAttest(ctx context.Context, deps *RebuildPackageDeps, mux rebuild.R
...
-		UseSyscallMonitor:  useSyscallMonitor,
-		SaveContainerImage: true,
-		Resources: build.Resources{
-			AssetStore:       buildStore,
-			ToolURLs:         toolURLs,
-			ToolAuthRequired: authRequired,
-			BaseImageConfig:  build.DefaultBaseImageConfig(),
-		},
-	})
+	buildOpts.BuildID = obID
```
msuozzo, line 199: while i understand the inclination, i think we're actually more likely to run into issues by splitting initialization across two functions here. moderate preference for keeping the status quo.

PR #1211, opened 2026-04-21
Replace pool-specific GCB client with region-specific GCB client

wbxyz:


pkg/gcb/gcb.go
```diff
@@ -87,46 +90,47 @@ func (c *gcsLogsClient) ListStepLogs(ctx context.Context, buildID string) (int,
...
-			Name: c.privatePoolConfig.Name,
+	if build.Options != nil && build.Options.Pool != nil {
+		if c.region != "" {
+			return nil, fmt.Errorf("attempting to use a non-regional client with private pool: %s", build.Options.Pool.Name)
 		}
-	}
-	if c.privatePoolConfig != nil && c.privatePoolConfig.Region != "" {
 		// For private pools, use the regional API endpoint if specified
-		return c.service.Projects.Locations.Builds.Create(fmt.Sprintf("projects/%s/locations/%s", project, c.privatePoolConfig.Region), build).Context(ctx).Do()
+		// TODO: Should we verify these regions match?
```
msuozzo, line 123: yeah i think we should do this. it's possible there are synonymous regions but i don't know of any

msuozzo (approved):

pkg/gcb/gcb.go
wbxyz: Ah, stale TODO. These are extracted using poolPat and compared with expected values below.

PR #1212, opened 2026-04-21

MarlinDiary:
Fixes #1123

PyPI wheel inference currently assumes that `.dist-info` paths exactly match the
registry package name after only replacing `-` with `_`.

That breaks on wheels whose `.dist-info` directory uses an equivalent historical
form, such as lowercase names or other unnormalized separators. For example,
`inferRequirements("128Autograder", "5.2.3", ...)` on `main` looks for:

`128Autograder-5.2.3.dist-info/WHEEL`

but the wheel may actually contain:

`128autograder-5.2.3.dist-info/WHEEL`

This change updates PyPI wheel inference to:

- look up normalized `.dist-info` paths first
- accept equivalent historical `.dist-info` name/version forms when reading
  `WHEEL` and `METADATA`
- keep the fallback restricted to top-level `.dist-info` files

This matches the PyPA packaging guidance that tools consuming `.dist-info`
directories should treat unnormalized historical forms as equivalent to their
normalized counterparts.

Tests added:
- normalized `.dist-info` path
- lowercase historical `.dist-info` path (`128Autograder`-style case)
- punctuation-equivalent names
- uppercase historical hyphenated path

Validation:
- `docker run --rm -v /Users/marlin/Documents/oss-rebuild:/src -w /src golang:1.25 go test ./pkg/rebuild/pypi -count=1`
- `docker run --rm -v /Users/marlin/Documents/oss-rebuild:/src -w /src golang:1.25 go run ./ci -v build`
- `docker run --rm -v /Users/marlin/Documents/oss-rebuild:/src -w /src golang:1.25 go run ./ci -v test`
- `docker run --rm -v /Users/marlin/Documents/oss-rebuild:/src -w /src golang:1.25 go run ./ci -v lint`
- `docker run --rm -v /Users/marlin/Documents/oss-rebuild:/src -w /src golang:1.25 go run ./ci -v fmt-check`
- `docker run --rm -v /Users/marlin/Documents/oss-rebuild:/src -w /src golang:1.25 go run ./ci -v imports-check`

### DIFF
--- a/pkg/rebuild/pypi/infer.go
+++ b/pkg/rebuild/pypi/infer.go
@@ -11,6 +11,7 @@ import (
 	"io"
 	"io/fs"
 	"log"
+	"path"
 	re "regexp"
 	"slices"
 	"strconv"
@@ -37,6 +38,8 @@ var commonRepoLinks = []string{
 	"project",
 	"github",
 }
+
+var distInfoFieldPat = re.MustCompile(`[-_.]+`)
 
 // There are two places to find the repo:
 // 1. In the ProjectURLs (project links)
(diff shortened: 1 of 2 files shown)


MarlinDiary:
@googlebot rescan

PR #1213, opened 2026-04-21

MarlinDiary:
Fixes #1123

PyPI wheel inference currently assumes that `.dist-info` paths match the
registry package name after replacing `-` with `_`. This breaks on wheels whose
`.dist-info` directory uses an equivalent historical form, for example
`128autograder-5.2.3.dist-info/WHEEL` for `128Autograder`.

This change first looks up the normalized `.dist-info` path, then falls back to
equivalent historical name/version forms when reading `WHEEL` and `METADATA`,
while keeping the fallback restricted to top-level `.dist-info` files.

Added regression tests covering normalized and historical equivalent
`.dist-info` names.

Validated with `go run ./ci -v build`, `go run ./ci -v test`, `go run ./ci -v lint`,
`go run ./ci -v fmt-check`, and `go run ./ci -v imports-check`.

### DIFF
--- a/pkg/rebuild/pypi/infer.go
+++ b/pkg/rebuild/pypi/infer.go
@@ -11,6 +11,7 @@ import (
 	"io"
 	"io/fs"
 	"log"
+	"path"
 	re "regexp"
 	"slices"
 	"strconv"
@@ -37,6 +38,8 @@ var commonRepoLinks = []string{
 	"project",
 	"github",
 }
+
+var distInfoFieldPat = re.MustCompile(`[-_.]+`)
 
 // There are two places to find the repo:
 // 1. In the ProjectURLs (project links)
@@ -164,10 +167,7 @@ func FindSourceDist(artifacts []pypireg.Artifact) (*pypireg.Artifact, error) {
 }
 
 func inferRequirements(name, version string, zr *zip.Reader) ([]string, error) {
-	// Name and version have "-" replaced with "_". See https://packaging.python.org/en/latest/specifications/recording-installed-packages/#the-dist-info-directory
-	// TODO: Search for dist-info in the gzip using a regex. It sounds like many tools do varying amounts of normalization on the path name.
-	wheelPath := fmt.Sprintf("%s-%s.dist-info/WHEEL", strings.ReplaceAll(name, "-", "_"), strings.ReplaceAll(version, "-", "_"))
-	wheel, err := getFile(wheelPath, zr)
+	wheel, wheelPath, err := getDistInfoFile(name, version, "WHEEL", zr)
 	if err != nil {
 		return nil, errors.Wrapf(err, "[INTERNAL] Failed to extract upstream %s", wheelPath)
 	}
@@ -180,9 +180,7 @@ func inferRequirements(name, version string, zr *zip.Reader) ([]string, error) {
 		// setuptools already set.
 		return reqs, nil
 	}
-	// TODO: Also find this with a regex.
-	metadataPath := fmt.Sprintf("%s-%s.dist-info/METADATA", strings.ReplaceAll(name, "-", "_"), strings.ReplaceAll(version, "-", "_"))
-	metadata, err := getFile(metadataPath, zr)
+	metadata, _, err := getDistInfoFile(name, version, "METADATA", zr)
 	if err != nil {
 		return nil, errors.Wrapf(err, "[INTERNAL] Failed to extract upstream dist-info/METADATA")
 	}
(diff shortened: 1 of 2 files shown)


pkg/rebuild/pypi/infer.go
```diff
@@ -203,6 +201,50 @@ func inferRequirements(name, version string, zr *zip.Reader) ([]string, error) {
...
+	expectedPath := fmt.Sprintf("%s-%s.dist-info/%s", normalizeDistInfoName(name), normalizeDistInfoVersion(version), fileName)
+	b, err := getFile(expectedPath, zr)
+	if err == nil || !errors.Is(err, fs.ErrNotExist) {
+		return b, expectedPath, err
+	}
+	for _, f := range zr.File {
+		if path.Base(f.Name) != fileName {
+			continue
+		}
+		if strings.Count(f.Name, "/") != 1 {
```
msuozzo, line 223: or `path.Dir(f.Name) != ""`

pkg/rebuild/pypi/infer.go
```diff
@@ -203,6 +201,50 @@ func inferRequirements(name, version string, zr *zip.Reader) ([]string, error) {
...
+			continue
+		}
+		dir := path.Base(path.Dir(f.Name))
+		stem, ok := strings.CutSuffix(dir, ".dist-info")
```
msuozzo, line 227: i suspect we could decompose this logic a bit better by creating a function that returns the "*.dist-info" dir name (which should be stable) and then do the normalization logic in this higher-level function.

we could even break it up further:

`getFile(getDistInfoDir(name, version, zr) +"/METADATA", zr)`

(we'd probably store the distinfo dir as a var and would want to use path.Join)

pkg/rebuild/pypi/infer.go
```diff
@@ -203,6 +201,50 @@ func inferRequirements(name, version string, zr *zip.Reader) ([]string, error) {
...
+		if !ok {
+			continue
+		}
+		dash := strings.LastIndexByte(stem, '-')
+		if dash == -1 {
+			continue
+		}
+		foundName, foundVersion := stem[:dash], stem[dash+1:]
+		if normalizeDistInfoName(foundName) != normalizeDistInfoName(name) {
```
msuozzo, line 236: in what circumstances would we expect this mismatch or the one below?

pkg/rebuild/pypi/infer.go
```diff
@@ -164,10 +167,7 @@ func FindSourceDist(artifacts []pypireg.Artifact) (*pypireg.Artifact, error) {
 }
 
 func inferRequirements(name, version string, zr *zip.Reader) ([]string, error) {
-	// Name and version have "-" replaced with "_". See https://packaging.python.org/en/latest/specifications/recording-installed-packages/#the-dist-info-directory
```
msuozzo, line 167: similarly to this comment here, i'd like to see an external link here that we can use to document the format we expect and any links on the canonicalization used.

pkg/rebuild/pypi/infer_test.go
```diff
@@ -73,3 +75,96 @@ func TestInferPythonVersion(t *testing.T) {
...
+		version string
+		files   map[string]string
+		want    []string
+	}{
+		{
+			name:    "exact normalized path",
+			pkg:     "friendly-bard",
+			version: "1.2.3",
+			files: map[string]string{
+				"friendly_bard-1.2.3.dist-info/WHEEL":    "Wheel-Version: 1.0\nGenerator: bdist_wheel (0.43.0)\nTag: py3-none-any\n",
```
msuozzo, line 92: given the above suggestion on simplifying this interface, the test would also be simplified to returning the expecting the right dir given a set of zip filenames. getFile itself gets tested here again when we're primarily just interested in the dirname selection

pkg/rebuild/pypi/infer_test.go
```diff
@@ -73,3 +75,96 @@ func TestInferPythonVersion(t *testing.T) {
...
+			for i := range tt.want {
+				if got[i] != tt.want[i] {
+					t.Fatalf("inferRequirements()[%d] = %q, want %q; got %v", i, got[i], tt.want[i], got)
+				}
+			}
+		})
+	}
+}
+
+func testZipReader(t *testing.T, files map[string]string) *zip.Reader {
```
msuozzo, line 148: we can use `pkg/archive/archivetest` which provides some nice helpers for this

msuozzo (commented):
thanks for the PR!

out of curiosity, do you have an idea of how often this normalization issue is occurring?

MarlinDiary:
Thanks, addressed in 69fce07.

I refactored the lookup to select the `.dist-info` directory first, then use `path.Join` for `WHEEL` and `METADATA`. I also simplified the tests to cover directory selection directly and switched the helper to `pkg/archive/archivetest`.

On frequency, I don't have a PyPI-wide number. As a quick bounded check, I looked at the repo's `pypi_top_1250_pure.json` sample: 86/1244 projects have PyPI `info.name` values where the old lookup form differs from the normalized dist-info form. I also checked a few actual wheels from that set, plus the reported `128Autograder` case, and confirmed several would miss `WHEEL` with the old lookup.

So I would not describe this as universal, but it does not appear to be isolated to the one reported package either.

msuozzo:
Apologies for not seeing your update! If you're willing to re-open, i actually ran into this problem this week, as well, so it'd be great to get a fix in.

MarlinDiary:
Re-opened, thanks for taking another look.

MarlinDiary:
Re-opened, thanks for taking another look.

MarlinDiary:
I could not reopen this PR because GitHub still has the old head repository recorded as unavailable after the fork was recreated. I opened the same fix as #1272.

PR #1216, opened 2026-04-22
Support rubygems in benchmark runner

msuozzo:


tools/benchmark/run/local.go
wbxyz, line 210: Can we update this to use rebuilder.UpstreamURL? This would match the way that internal/api/apiservice/rebuild.go implements it.

Might be worth doing in a follow-on

wbxyz (approved):

tools/benchmark/run/local.go
wbxyz: Sent in https://github.com/google/oss-rebuild/pull/1233

msuozzo: sg

PR #1217, opened 2026-04-22
Add gem metadata stabilizers

msuozzo:
These are regrettably implemented as regexes but the "yaml" used as
metadata is foreign enough to standard serialization that round-tripping
would introduce an unacceptable level of deviation from the common form.

### DIFF
--- a/pkg/stabilize/gem.go
+++ b/pkg/stabilize/gem.go
@@ -5,6 +5,7 @@ package stabilize
 
 import (
 	"bytes"
+	"regexp"
 	"slices"
 
 	"github.com/google/oss-rebuild/pkg/archive"
@@ -13,6 +14,8 @@ import (
 // AllGemStabilizers is the list of all available gem stabilizers.
 var AllGemStabilizers = []Stabilizer{
 	StableGemExcludeChecksums,
+	StableGemMetadataDate,
+	StableGemMetadataRubygemsVersion,
 	StableGemInnerArchives,
 }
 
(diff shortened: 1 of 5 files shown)


pkg/stabilize/gem.go
```diff
@@ -27,6 +30,26 @@ var StableGemExcludeChecksums = Stabilizer{
...
 
+var gemMetadataDateRe = regexp.MustCompile(`(?m)^date: .+$`)
+var gemMetadataRubygemsVersionRe = regexp.MustCompile(`(?m)^rubygems_version: .+$`)
+
+// StableGemMetadataDate normalizes the date field in gem metadata YAML.
+// The date is stamped by `gem build` at build time and so varies across rebuilds.
+var StableGemMetadataDate = Stabilizer{
+	Name: "gem-metadata-date",
+}.WithConstraints(AtDepth(1), ArchivePath("metadata.gz")).WithFn(GzipContentFn(func(b []byte) []byte {
+	return gemMetadataDateRe.ReplaceAll(b, []byte("date: 1980-01-01 00:00:00.000000000 Z"))
```
wbxyz: nit: Why 1980?

wbxyz (approved):

pkg/stabilize/gem.go
msuozzo: oops this should be 01-02 and definitely deserves the link to the source. in short: zip

https://github.com/ruby/rubygems/blob/master/lib/rubygems.rb#L149

PR #1218, opened 2026-04-22
Support multiple actions per ci command

msuozzo:


wbxyz (approved):

ci/main.go
```diff
@@ -26,33 +26,38 @@ func main() {
...
-		"fix":           fix,
+	// Resolve args to a flat task list.
+	var allTasks []task
+	var sequential bool
+	for _, name := range args {
+		if fn, ok := tasks[name]; ok {
+			allTasks = append(allTasks, task{name: name, fn: fn})
+		} else if comp, ok := composites[name]; ok {
+			allTasks = append(allTasks, comp.tasks...)
+			sequential = comp.sequential
```
wbxyz: How does this handle if there are multiple composite tasks and the last one has sequential=false?

ci/main.go
```diff
@@ -26,33 +26,38 @@ func main() {
...
-		"imports":       importsFix,
-		"imports-check": importsCheck,
-		"license":       licenseFix,
-		"license-check": licenseCheck,
-		"check":         check,
-		"fix":           fix,
+	// Resolve args to a flat task list.
+	var allTasks []task
+	var sequential bool
+	for _, name := range args {
```
wbxyz, line 33: Do you care if there are duplicates? Or if the composites have overlap?

wbxyz (approved):

ci/main.go
msuozzo: yeah there's only one non-sequential composite currently but cmp.Or should fix that.

ci/main.go
msuozzo: meh i think it's tolerable. makes it clearer that a call may be redundant.

PR #1223, opened 2026-04-22
Use host.docker.internal for local network testing

msuozzo:
This is the better supported mode for serving locally without needing
to be a part of the host network namespace.

### DIFF
--- a/internal/timewarp/integration_test.go
+++ b/internal/timewarp/integration_test.go
@@ -13,6 +13,7 @@ import (
 	"net/http"
 	"os"
 	"os/exec"
+
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


wbxyz:
This seems similar to the update to the local docker run executor proposed in https://github.com/google/oss-rebuild/pull/1015


Maybe once that's merged we could re-use the docker run executor for this test?

wbxyz (approved):

msuozzo:
> This seems similar to the update to the local docker run executor proposed in #1015
> 
> Maybe once that's merged we could re-use the docker run executor for this test?

yeah that's an idea. not immediately clear to me that we'd want timewarp to add a dep on pkg/build but i don't think it's out of the question.

that said, this is a pretty minor change so i think it's relatively safe to merge now and consider the refactor after the other change.

PR #1230, opened 2026-04-27
Update ctl to use lro endpoints

wbxyz:


tools/benchmark/run/run.go
```diff
@@ -60,7 +63,24 @@ func (s *remoteExecutionService) RebuildPackage(ctx context.Context, req schema.
...
+			return nil, errors.Wrap(err, "getting rebuild operation")
+		}
+	}
+	if op.Result != nil {
+		return op.Result, nil
+	}
+	if op.Error != nil {
+		return nil, errors.New(op.Error.Message)
+	}
+	return nil, errors.New("operation completed without response or error")
```
msuozzo: this breaks an invariant of the API so i'm not sure we should be handling it explicitly. i feel this is repetitive enough that we could provide a (generic) helper to poll these to completion and we could assert-panic this invariant there.

it's fine if that's a follow-on.

msuozzo (approved):

tools/benchmark/run/run.go
wbxyz: Refactored a bit, because longrunning.OperationError implements the Error interface I can actually just return op.Result, op.Error

PR #1231, opened 2026-04-27
Update cli.RunE to set SilenceUsage after input parsing

wbxyz:
By default, any error returned by RunE causes Cobra to print the usage string. This is confusing and noisy when a command fails for runtime reasons.

By keeping SilenceUsage=false for the first half of execution, any input argument failures, args parsing, Validate() or initDeps() failures will result in the usage string printing. Beyond this point in the code, failures will be logged and non-zero status will be returned, but the usage string will no longer be printed.

### DIFF
--- a/pkg/act/cli/cli.go
+++ b/pkg/act/cli/cli.go
@@ -45,6 +45,10 @@ func RunE[I act.Input, O any, D Deps](
 		if err != nil {
 			return errors.Wrap(err, "initializing dependencies")
 		}
+		// Beyond this point in the code, failures are not usually related to command usage.
+		// Instead, they're application failures which should be reflected in the error message.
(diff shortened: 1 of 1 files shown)


pkg/act/cli/cli.go
```diff
@@ -45,6 +45,7 @@ func RunE[I act.Input, O any, D Deps](
 		if err != nil {
 			return errors.Wrap(err, "initializing dependencies")
 		}
+		cmd.SilenceUsage = true
```
msuozzo, line 51: nice this looks a lot better. let's add a comment here explaining why we're setting this here.

msuozzo (approved):

pkg/act/cli/cli.go
wbxyz: Added a comment, and linked the thread from the library's github repo where I found this solution.

PR #1235, opened 2026-04-27
Populate operation.Error for RebuildStatusFailure attempts

wbxyz:


internal/api/apiservice/rebuild_lro.go
```diff
@@ -42,9 +42,9 @@ func ProjectRebuildAttempt(a schema.RebuildAttempt) longrunning.Operation[schema
...
 	}
 	switch a.Status {
-	case schema.RebuildStatusSuccess, schema.RebuildStatusFailure:
+	case schema.RebuildStatusSuccess:
 		op.Done = true
-	case schema.RebuildStatusError:
+	case schema.RebuildStatusFailure, schema.RebuildStatusError:
 		op.Done = true
 		op.Error = &longrunning.OperationError{
 			Code:    http.StatusInternalServerError,
```
msuozzo, line 50: is this an appropriate error code for this case?

msuozzo (approved):

internal/api/apiservice/rebuild_lro.go
wbxyz: Before mailing this I briefly looked over the http status codes, the closest one I found was `417 Expectation Failed` but that's a semantic meaning about Expect headers so I don't think it's appropriate here. https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/417

We were previously using Internal Server Error for the inference and build failures so I figured that's good enough for now. We could definitely revisit this though.

PR #1239, opened 2026-05-04
Convert stabilize cli to act

msuozzo:


pkg/stabilize/cli/action.go
```diff
@@ -0,0 +1,301 @@
...
+		if i == 0 {
+			result = stabs
+		} else if !slices.EqualFunc(result, stabs, func(s1, s2 stabilize.Stabilizer) bool {
+			return s1.Name == s2.Name
+		}) {
+			return nil, errors.Wrapf(ErrAmbiguousEcosystem, "ecosystem %s suggests different stabilizers than %s", candidates[0], e)
+		}
+	}
+	return result, nil
+}
```
wbxyz, line 262: Would it make sense for this logic to be outside the `action.go` file? StabilizerRegistry seems appropriate for pkg/stabilize/registry.go, maybe the eligibility logic should also be there?

wbxyz (approved):

pkg/stabilize/cli/action.go
msuozzo: that wasn't really meant to be public. it's a glorified ordered map.

msuozzo: we can pull things out piecemeal as we need them. off the top, I'm not sure there's a usage for them just yet since the prod uses of stabilizer use the Target instead of the filename.

PR #1241, opened 2026-05-04
Set prebuild file names inside the build

msuozzo:
Previously, the code assumed the filenames of the bootstrap containers.
While expected, this isn't guaranteed to be true. Additionally, the name
selection rules are sufficiently complex (and differ slightly between
wget and curl) that doing so statically is not reliable.


wbxyz (approved):
Nice

PR #1242, opened 2026-05-04
Add explicit run ID support to ctl run

msuozzo:


tools/ctl/command/runone/runone.go
```diff
@@ -262,6 +266,24 @@ func handleRemote(ctx context.Context, cfg Config, deps *Deps, enc *json.Encoder
...
+			runID = time.Now().UTC().Format(time.RFC3339)
+		} else {
+			// When a custom run ID is provided, ensure the run exists
+			// Use BenchmarkName to pass the desired run ID
+			createRunStub := api.Stub[schema.CreateRunRequest, schema.Run](client, apiURL.JoinPath("runs"))
+			run, err := createRunStub(ctx, schema.CreateRunRequest{
+				BenchmarkName: runID,
+				BenchmarkHash: "",
+				Type:          string(schema.AttestMode),
+			})
```
wbxyz, line 282: Could you mirror this logic for local rundex management in handleLocal? I think we'd need to create a document in the local rundex. This is how runbenchmark.go does it:

https://github.com/google/oss-rebuild/blob/4a267623b34beff7b11f6e76c31fee5bb990bf36/tools/ctl/command/runbenchmark/runbenchmark.go#L193-L201

wbxyz:
Is it worth logging a warning that re-using a run-id is liable to overwrite existing metadata of previous attempts for the same target?

wbxyz (approved):

msuozzo:
> Is it worth logging a warning that re-using a run-id is liable to overwrite existing metadata of previous attempts for the same target?

eh I think the unconditional log line should be enough given it's an opt-in feature. explicit detection for pre-existing attempts could be a future addition but i think that's out-of-scope here.

PR #1246, opened 2026-05-05
Refactor logs to use html templates

wbxyz:


internal/api/dashboard/logs.go
```diff
@@ -27,6 +29,69 @@ type LogsRequest struct {
...
+		Package:   req.Package,
+		Version:   req.Version,
+		Artifact:  req.Artifact,
+	}
+	obj, attrs, err := fetchLogObject(ctx, req, deps)
+	if err != nil {
+		return nil, err
+	}
+
+	if attrs.Size > 10*1024*1024 {
```
msuozzo: constant? or Dep really?

internal/api/dashboard/logs.go
```diff
@@ -27,6 +29,69 @@ type LogsRequest struct {
...
+	if err != nil {
+		return nil, fmt.Errorf("creating log reader: %w", err)
+	}
+	defer reader.Close()
+
+	var lines []LogLine
+	scanner := bufio.NewScanner(reader)
+	lineNumber := 1
+	for scanner.Scan() {
+		lines = append(lines, LogLine{
```
msuozzo, line 80: oof... well the raw threshold is low enough that this should be fine, i suppose.

msuozzo (approved):

internal/api/dashboard/logs.go
wbxyz: Agreed, though this kind of overhead is necessary when doing rich rendering with log lines each having an html element. That's the motivation for have the raw logs fallback.

internal/api/dashboard/logs.go
wbxyz: Going with constant for simplicity but we can make it a dep in the future.

PR #1252, opened 2026-05-07
pkg/archive: reject symlinks that escape the extraction root in ExtractTar

evilgensec:
## Summary

`ExtractTar` in `pkg/archive/tar.go` has two symlink path-traversal bypasses:

**Bypass 1 — symlink destination escapes the skip guard:**
The `if h.Linkname != ""` branch executes before the `if skip` check. A tar entry with `Name = "../../evil"` sets `skip = true`, but the symlink is still created because the skip guard only covers the `IsDir()` and regular-file branches.

**Bypass 2 — symlink target is not validated:**
`filepath.Rel(basepath, h.Linkname)` is called but its result is never checked for `..` components. A tar entry with `Name = "safe_link"` (skip=false) and `Linkname = "../../../../etc/passwd"` creates a symlink pointing outside the extraction root. An absolute `Linkname` (e.g. `/etc/passwd`) causes `filepath.Rel` to return an error that was previously propagated rather than handled.

Since `osfs.Symlink` calls `os.Symlink` directly (no chroot), the resulting symlink on the host filesystem is followed by subsequent `go-git` operations (`Open`, `Checkout`) outside `tmpDir`.

**Attack scenario:** A compromised or MITM'd gitcache server (or GCS bucket poisoning) serves a malicious `.tar.gz`. The rebuilder worker extracts it, creating a dangling symlink. Subsequent git operations follow the symlink outside the sandbox — enabling arbitrary file read/write on the rebuilder host.

## Fix

- Move the `skip` check to cover the symlink branch.
- Skip entries where `filepath.Rel` returns an error (absolute `Linkname`).
- Skip entries where the resolved `linkpath` contains `..` components.

## Tests

Added `TestExtractTarSymlinkTraversal` covering four escape patterns (dotdot name, dotdot target, absolute target, both) and `TestExtractTarSymlinkSafe` as a regression guard for well-formed symlinks.

### DIFF
--- a/pkg/archive/tar_test.go
+++ b/pkg/archive/tar_test.go
@@ -1,0 +1,89 @@
+// Copyright 2025 Google LLC
+// SPDX-License-Identifier: Apache-2.0
+
+package archive
+
+import (
+	"archive/tar"
+	"bytes"
+	"path/filepath"
+	"testing"
+
+	"github.com/go-git/go-billy/v5/memfs"
+)
+
+// buildTarWithSymlink creates a tar reader containing a single symlink entry.
+func buildTarWithSymlink(name, linkname string) *tar.Reader {
+	buf := &bytes.Buffer{}
+	tw := tar.NewWriter(buf)
+	_ = tw.WriteHeader(&tar.Header{
+		Typeflag: tar.TypeSymlink,
+		Name:     name,
+		Linkname: linkname,
+	})
+	_ = tw.Close()
+	return tar.NewReader(bytes.NewReader(buf.Bytes()))
+}
+
+// TestExtractTarSymlinkTraversal verifies that symlink entries whose destination
+// path or target escapes the extraction root are silently skipped rather than
+// written to the filesystem.
+func TestExtractTarSymlinkTraversal(t *testing.T) {
+	tests := []struct {
+		desc     string
+		name     string
+		linkname string
+	}{
+		{
+			desc:     "symlink destination escapes via dotdot",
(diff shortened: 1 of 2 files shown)


pkg/archive/tar_test.go
```diff
@@ -0,0 +1,89 @@
...
+		Linkname: linkname,
+	})
+	_ = tw.Close()
+	return tar.NewReader(bytes.NewReader(buf.Bytes()))
+}
+
+// TestExtractTarSymlinkTraversal verifies that symlink entries whose destination
+// path or target escapes the extraction root are silently skipped rather than
+// written to the filesystem.
+func TestExtractTarSymlinkTraversal(t *testing.T) {
```
msuozzo: can you use pkg/archive/archivetest for these?

pkg/archive/tar_test.go
```diff
@@ -0,0 +1,89 @@
...
+			if len(entries) != 0 {
+				t.Errorf("ExtractTar() created %d unexpected entries in root: %v", len(entries), entries)
+			}
+		})
+	}
+}
+
+// TestExtractTarSymlinkSafe verifies that a well-formed symlink (both name and
+// target within the extraction root) is extracted successfully.
+func TestExtractTarSymlinkSafe(t *testing.T) {
```
msuozzo: try to fold this into the above table-based test

msuozzo (approved):

msuozzo (approved):

PR #1253, opened 2026-05-07
Support custom stabilizers in differs

msuozzo:


spencerschrock (approved):

tools/ctl/difftool/diffr.go
```diff
@@ -12,11 +12,18 @@ import (
...
 
 // Diffr implements Differ using the diffr library.
-type Diffr struct{}
+//
+// ExtraStabilizers are appended to the default stabilizer set returned by
+// stability.StabilizersForTarget and support build-definition-scoped custom
+// stabilizers.
+type Diffr struct {
+	ExtraStabilizers []stabilize.Stabilizer
+}
```
spencerschrock: meant to ask if this is something you'd want the same extra stabilizers re-used across the differ, or if it should be an arg in the Differ `Diff` method interface

msuozzo: Yeah I can see that as an option. I don't have a strong preference one way or the other but I think one advantage of that it being part of the interface would be ensured we maintained support across the implementation. Let me make the change

PR #1254, opened 2026-05-08
Add a corpus of pypi definitions

msuozzo:
This covers a single release of 50 popular packages.

### DIFF
--- a/definitions/pypi/requests-toolbelt/1.0.0/requests_toolbelt-1.0.0-py2.py3-none-any.whl/build.yaml
+++ b/definitions/pypi/requests-toolbelt/1.0.0/requests_toolbelt-1.0.0-py2.py3-none-any.whl/build.yaml
@@ -1,0 +1,31 @@
+# inference differs:
+#  source diff:
+#    git clone https://github.com/requests/toolbelt .
+#    git checkout --force 'b7d1a1fcdda9ebcd9afe5011690ab860fce780c2'
+#   +git checkout 'da0306dd4f84cdcad6b0e694d674b9d9c9479c61^' -- requests_toolbelt/adapters/appengine.py
+flow:
+  location:
+    repo: https://github.com/requests/toolbelt
+    ref: b7d1a1fcdda9ebcd9afe5011690ab860fce780c2
+  src:
+    - uses: git-checkout
+    # The 1.0.0 release tag (b7d1a1f) does not contain
+    # requests_toolbelt/adapters/appengine.py — it was removed in da0306d
+    # ("Remove Google AppEngine support") two commits earlier — but the
+    # wheel that was uploaded to PyPI does. The published sdist does not.
+    # The wheel was built from a working tree that still had the deleted
+    # file. Restore it from the parent of the deletion commit so the
+    # rebuild matches.
+    - runs: git checkout 'da0306dd4f84cdcad6b0e694d674b9d9c9479c61^' -- requests_toolbelt/adapters/appengine.py
+  deps:
+    - uses: pypi/setup-venv
+      with:
+        locator: /usr/bin/
+        path: /deps
+    - uses: pypi/setup-registry
+      with:
+        registryTime: "2023-05-01T04:11:28Z"
+    - runs: /deps/bin/pip install wheel==0.40.0 setuptools==67.7.2 build
+  build:
+    - runs: /deps/bin/python3 -m build --wheel -n
+  output_dir: dist
--- a/definitions/pypi/shellingham/1.4.0/shellingham-1.4.0-py2.py3-none-any.whl/build.yaml
+++ b/definitions/pypi/shellingham/1.4.0/shellingham-1.4.0-py2.py3-none-any.whl/build.yaml
@@ -1,0 +1,54 @@
+# inference differs:
+#  deps diff:
+#    /usr/bin/uvx uv venv /deps --seed --python 3.11
+#    export PIP_INDEX_URL=http://pypi:2021-02-01T09:27:47Z@TIMEWARP/simple
+#   -/deps/bin/pip install build setuptools==56.2.0 wheel==0.36.2
+#   +/deps/bin/pip install --no-deps setuptools==52.0.0 wheel==0.36.2
+#  build diff:
+#   -/deps/bin/python3 -m build --wheel -n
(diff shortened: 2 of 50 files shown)


definitions/pypi/google-api-core/2.30.3/google_api_core-2.30.3-py3-none-any.whl/build.yaml
wbxyz, line 1: For all these definitions with vanilla pypi_pure_wheel_build strategies, how many could instead be just location hints? I'm guessing some of these are primarily resolving repo/ref difficulties.

wbxyz: If they're not, and it's more focused on requirements or registry_time, would you be able to document why those changes were needed?

definitions/pypi/requests-toolbelt/1.0.0/requests_toolbelt-1.0.0-py2.py3-none-any.whl/build.yaml
```diff
@@ -0,0 +1,26 @@
...
+  src:
+    - uses: git-checkout
+    # The 1.0.0 release tag (b7d1a1f) does not contain
+    # requests_toolbelt/adapters/appengine.py — it was removed in da0306d
+    # ("Remove Google AppEngine support") two commits earlier — but the
+    # wheel that was uploaded to PyPI does. The published sdist does not.
+    # The wheel was built from a working tree that still had the deleted
+    # file. Restore it from the parent of the deletion commit so the
+    # rebuild matches.
+    - runs: git checkout 'da0306dd4f84cdcad6b0e694d674b9d9c9479c61^' -- requests_toolbelt/adapters/appengine.py
```
wbxyz, line 19: Does building at that commit work instead of copy the file forward?

definitions/pypi/shellingham/1.4.0/shellingham-1.4.0-py2.py3-none-any.whl/build.yaml
```diff
@@ -0,0 +1,45 @@
...
+    - runs: /deps/bin/pip install --no-deps setuptools==52.0.0 wheel==0.36.2
+  build:
+    - runs: /deps/bin/python3 setup.py bdist_wheel
+  output_dir: dist
+custom_stabilizers:
+  - replace_pattern:
+      paths: ["*/zip-safe"]
+      pattern: "\r\n"
+      replace: "\n"
+    reason: |
```
wbxyz, line 36: This feels like it would be a good general stabilizer?

wbxyz (approved):
1) This is really cool!
2) The explanatory comments are very helpful. First thought is that we should make a dedicated field for them. Second thought is that a dedicated field might bloat the build definition that ends up in the attestation and infra, while comments get dropped during marshal/unmarshal. Third thought is that might actually be worthwhile to include comments as a sort of justification for the definition?
3) Should we make a re-usable flow for replacement (to avoid sed -i everywhere)? Could potentially re-use the replacement stabilizer with archive path support?

wbxyz:
Something else I was thinking while reviewing:

1) When we have a better process, we should maybe do 1 def per PR. That makes it easier to reject one but keep the rest, and could also make it easier to add links to evidence in the PR description

2) Can we improve the confidence in the repo URLs? I feel like that's the easiest way for something to sneak past review in a process like that, where the build def has a typo or a fork repo. If the build def is intentionally adding the repo that's one thing, but if we're just including the repo as part of the machinery to describe other steps it would be nice to avoid taking on this potential failure mode.

Idk what surface to display this on, but we could have a trusted inference service attempt to infer the build definition as part of the Pull request github actions (assuming one build def per PR), then that can be diffed against the proposed build def, in which case the repo being different would be obvious.

Even further down the road, we could add scorecards or some other metadata about these repos when the PR author is proposing a different git repo for building.

definitions/pypi/shellingham/1.4.0/shellingham-1.4.0-py2.py3-none-any.whl/build.yaml
msuozzo: it's an option but with the RECORD file stabilizer still unimplemented, it would always require an accompanying stabilizer anyway. Good to add when that's ready.

definitions/pypi/requests-toolbelt/1.0.0/requests_toolbelt-1.0.0-py2.py3-none-any.whl/build.yaml
msuozzo: nope. this was a case where there was almost certainly a local modification to the work dir that didn't make it into a specific commit.

msuozzo:
> Something else I was thinking while reviewing:
> 
> 1. When we have a better process, we should maybe do 1 def per PR. That makes it easier to reject one but keep the rest, and could also make it easier to add links to evidence in the PR description

I think bulk additions are more sustainable than individual PRs. It's easy to revert an individual file regardless. I agree, though, that tooling will help scale our process.

> 2. Can we improve the confidence in the repo URLs? I feel like that's the easiest way for something to sneak past review in a process like that, where the build def has a typo or a fork repo. If the build def is intentionally adding the repo that's one thing, but if we're just including the repo as part of the machinery to describe other steps it would be nice to avoid taking on this potential failure mode.
>
> Idk what surface to display this on, but we could have a trusted inference service attempt to infer the build definition as part of the Pull request github actions (assuming one build def per PR), then that can be diffed against the proposed build def, in which case the repo being different would be obvious.

you could potentially run that yourself locally. and again, i don't think there's any specifics here that necessitate one def per PR.
 
> Even further down the road, we could add scorecards or some other metadata about these repos when the PR author is proposing a different git repo for building.

Hmm not sure that would make sense. A malicious repo could have a high scorecard score but still not map to the intended package. I think a better approach would be to lint against the use of any repo not found in our ground truth sources and/or that recently changed.

msuozzo:
Okay so i managed a diff header formatter in #1264 that i feel balances brevity with clarity. I've annotated the headers on all the definitions here and we should probably make this part of the workflow moving forward.

PR #1255, opened 2026-05-11
Handle anchor URLs specifically to prevent navigating to site root in render-link.html

gabhisekh:
Ref: https://docs.oss-rebuild.dev/builds/ArtifactEquivalence@v0.1/  "section below" is transformed by hugo to `<a href="/#artifact-stabilization-details">section below</a>`.

Feel free to suggest if there are simpler ways to do this or to discard if there are other plans to handle this, etc.


msuozzo:
Hey thanks for pointing this out! this actually uncovered some additional issues with the approach we were using so, to avoid asking to expand the scope here too much, i put together #1268. Thanks again for the report!

PR #1259, opened 2026-05-14
Update terraform to avoid authoratative bindings

wbxyz:
This prevents oss rebuild from breaking other deployments in the same
cloud project.

### DIFF
--- a/deployments/iam.tf
+++ b/deployments/iam.tf
@@ -70,362 +70,369 @@ resource "google_project_iam_custom_role" "bucket-viewer-role" {
   title       = "Bucket Viewer"
   permissions = ["storage.buckets.get", "storage.buckets.list"]
 }
-resource "google_storage_bucket_iam_binding" "git-cache-manages-git-cache" {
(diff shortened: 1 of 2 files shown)


deployments/iam.tf
```diff
@@ -70,362 +70,369 @@ resource "google_project_iam_custom_role" "bucket-viewer-role" {
...
-resource "google_kms_crypto_key_iam_binding" "attestors-uses-signing-key" {
+  ))
   crypto_key_id = google_kms_crypto_key.signing-key.id
-  role          = "roles/cloudkms.signerVerifier"
-  members = concat(
+  role          = "roles/cloudkms.viewer"
+  member        = each.key
+}
+resource "google_kms_crypto_key_iam_member" "attestors-uses-signing-key" {
+  for_each = toset(concat(
```
msuozzo, line 221: what if we moved these for_each blocks down either before or after member so we retain the 'members declared ~last' property the old blocks had.

msuozzo (approved):
looks good. one small nit but lgtm

PR #1266, opened 2026-05-15
Add rubygems benchmark

msuozzo:


tools/benchmark/generate/main.go
```diff
@@ -1276,6 +1278,122 @@ LIMIT 2500
...
+			if err != nil {
+				log.Printf("error fetching versions for %s: %v", p.Package, err)
+				continue
+			}
+			// Heuristic for pure Ruby: all versions should be platform "ruby".
+			// Native gems (even source-only ones) eventually tend to publish pre-compiled binaries
+			// or have at least one version with a platform.
+			isPure := true
+			for _, v := range versions {
+				if v.Platform != "ruby" && v.Platform != "java" {
```
spencerschrock: not a ruby dev, is the inclusion of java here intended?

spencerschrock (approved):

tools/benchmark/generate/main.go
msuozzo: yeah I think this is how jRuby is specified but a quick search reveals there are some others i missed ("universal-java" may be another one). Added.

PR #1268, opened 2026-05-15
Fix docs linkification

msuozzo:
As surfaced in #1255, the previous version broke anchor links (#foo -> /#foo)
and also all markdown links of remote URLs
(github.com/foo/bar/blob/main/README.md -> .md stripped leading to 404).
This new linkifier only targets the site-local markdown links we'd initially
intended to fix.

The new version parses the URL to isolate rewrites to relative markdown
links. We then reconstruct it and materialize the link in the same
callsite as all others.

### DIFF
--- a/docs/layouts/_default/_markup/render-link.html
+++ b/docs/layouts/_default/_markup/render-link.html
@@ -7,15 +7,14 @@
 {{- $title := .Title -}}
 {{- $text := .Text -}}
 
-{{- if strings.HasSuffix $destination ".md" -}}
-  {{- $destination = strings.TrimSuffix ".md" $destination -}}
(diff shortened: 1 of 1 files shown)


spencerschrock (approved):
.md rewrite LGTM, anchors still broken (which commit message says isnt the aim here)

msuozzo:
> .md rewrite LGTM, anchors still broken (which commit message says isnt the aim here)

The anchors look to be fixed when accessed through a local dev server (although that may differ from how they'll render on sites).

spencerschrock:
Resolved confusion offline, unrelated to the trailing slash in Hugo routing, but rather rewriting of anchors in the linked PR.

PR #1272, opened 2026-05-18

MarlinDiary:
Fixes #1123

Follow-up to #1213; the original PR could not be reopened after the fork/head repository link was lost.

PyPI wheel inference currently assumes that `.dist-info` paths match the registry package name after replacing `-` with `_`. This breaks on wheels whose `.dist-info` directory uses an equivalent historical form, for example `128autograder-5.2.3.dist-info/WHEEL` for `128Autograder`.

This change first resolves the matching `.dist-info` directory, then uses that directory when reading `WHEEL` and `METADATA`. The fallback accepts equivalent historical name/version forms while remaining restricted to top-level `.dist-info` directories.

Added regression tests covering normalized and historical equivalent `.dist-info` names.

Validated with `go run ./ci -v build`, `go run ./ci -v test`, `go run ./ci -v lint`, `go run ./ci -v fmt-check`, and `go run ./ci -v imports-check`.

### DIFF
--- a/pkg/rebuild/pypi/infer.go
+++ b/pkg/rebuild/pypi/infer.go
@@ -11,6 +11,7 @@ import (
 	"io"
 	"io/fs"
 	"log"
+	"path"
 	re "regexp"
 	"slices"
 	"strconv"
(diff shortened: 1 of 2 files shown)


msuozzo (approved):
awesome work!

PR #1282, opened 2026-05-27
Add stabilizer ordering

msuozzo:


pkg/stabilize/context.go
```diff
@@ -48,8 +52,12 @@ func (ctx *StabilizationContext) Format() archive.Format {
 }
 
 // WithStabilizers returns a new context with the given stabilizers.
+// Stabilizers are stable-sorted by Ordering so the pipeline respects the
+// declared Stages regardless of input slice order.
 func (ctx *StabilizationContext) WithStabilizers(stabilizers []Stabilizer) *StabilizationContext {
```
wbxyz, line 57: Unrelated, should this extend ctx.Stabilizers instead of replacing it?

wbxyz (approved):

pkg/stabilize/context.go
msuozzo: It's a fair question. I think there could be cases where we'd want to subset the available stabilizers but I believe we only use it for supersetting right now. The proposed semantics could end up applying a stabilizer more than once (but, on the other hand, maybe that's functionality we'd want). It's certainly something to consider if/when we run into the need.

PR #1283, opened 2026-05-27
Add wheel RECORD file stabilizer

msuozzo:


pkg/stabilize/wheel.go
```diff
@@ -0,0 +1,86 @@
...
+//
+// NOTE: Runs at StageFinalize as any earlier stabilizer that altered an entry's
+// content or membership would render the RECORD inaccurate. This is notably
+// the case for custom stabilizers which often exclude paths from the package.
+var StableWheelRecord = Stabilizer{
+	Name: "wheel-record",
+}.WithFn(ZipArchiveFn(func(mr *archive.MutableZipReader) {
+	var record *archive.MutableZipFile
+	for _, f := range mr.File {
+		if strings.HasSuffix(f.Name, ".dist-info/RECORD") {
```
wbxyz: Should we add guards to make sure this only finds .dist-info at the root dir? Or if we're supporting RECORD files deeper in the stack maybe we should limit the entries to be in the same directory?

wbxyz (approved):

pkg/stabilize/wheel.go
msuozzo: yeah in fairness i actually introduced this simplification because i thought glob.Match was overkill given we should be able to assume dist-info at top-level. i suppose it's cheap to protect against this case, though.

PR #1284, opened 2026-05-28
Port more commit inference strategies to npm and pypi

pohlarized:
Across all ecosystems, i have seen the following commit inference strategies:
- git tag heuristic
- commit file hash overlap with artifact file hashes
- version change in manifest
- explicit ref in registry

I have ported the commit file hash overlap strategy that was exclusive to maven to npm and PyPI, and also ported the manifest version change strategy from npm to PyPI.
The registry ref strategy was not ported, because PyPI simply does not expose this information.
Thus, now npm supports all four heuristics, and PyPI supports the three that make sense.

Some notes:

- The commit file hash overlap was previously exclusive to maven, and thus lived in maven specific files. I have refactored out the generic part and moved it to a different file.
- I am unsure about the way and granularity you test these features. For now, i only introduced some patches to the tests so that they don't fail (and perhaps in a weird way, because I'm not so familiar with go and the way tests work).
- I have copied some models from `pkg/rebuild/pypi/parsing/pyproject.go` to `pypi/infer.go` because they were private, we have to see how we want to deal with that.

### DIFF
--- a/pkg/rebuild/maven/infer.go
+++ b/pkg/rebuild/maven/infer.go
@@ -15,7 +15,6 @@ import (
 	"strings"
 
 	"github.com/go-git/go-git/v5"
-	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/go-git/go-git/v5/plumbing/object"
 	"github.com/go-git/go-git/v5/plumbing/transport"
 	"github.com/google/oss-rebuild/internal/gitx"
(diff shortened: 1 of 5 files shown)


wbxyz:
Looks good! Would you mind splitting this up a little bit into separate PRs? Mainly the python code could benefit from some more discussion but shouldn't block the other ecosystems being merged in.

pohlarized:
> Looks good! Would you mind splitting this up a little bit into separate PRs? Mainly the python code could benefit from some more discussion but shouldn't block the other ecosystems being merged in.

Sure! I have created #1308 and #1309 so I am closing this PR in favor of those.

PR #1285, opened 2026-05-28
Add Jar signature stabilizer

msuozzo:


pkg/stabilize/jar.go
```diff
@@ -213,6 +216,95 @@ var StableMavenPomProperties = Stabilizer{
...
+		if mf.Name != "META-INF/MANIFEST.MF" {
+			continue
+		}
+		stripManifestDigestSections(mf)
+	}
+}))
+
+func hasJarSignature(mr *archive.MutableZipReader) bool {
+	for _, mf := range mr.File {
+		if match, _ := glob.Match("META-INF/*.SF", mf.Name); match {
```
wbxyz, line 267: Should this multiMatch over all jarSignaturePatterns?

pkg/stabilize/jar.go
```diff
@@ -213,6 +216,95 @@ var StableMavenPomProperties = Stabilizer{
...
+//
+// The jarsigner logic we are undoing adds file digests to the manifest one of
+// two ways (see JarSigner.{updateDigests,getDigestAttributes}):
+//   - When entry already has a manifest section: add an <alg>-Digest attribute
+//     and leave its other attributes untouched.
+//   - When entry has no section: create a new section holding only Name and the
+//     <alg>-Digest attribute(s) (only applied to non-directory entries).
+func stripManifestDigestSections(mf *archive.MutableZipFile) {
+	r, err := mf.Open()
+	if err != nil {
```
wbxyz, line 284: stripManifestDigestSections can return an error?

Might also be nice to log the returned error, but we could also push the "throw away the error" decision higher up?

wbxyz (approved):

pkg/stabilize/jar.go
msuozzo: so .SF is sort of the "signing manifest" while the other suffixes are for the signatures themselves. this is more used to reduce false positives where a file with the .EC suffix is inadvertently removed: we only do so when we're confident we should be expecting a signature file.

pkg/stabilize/jar.go
msuozzo: absolutely. i think this code is consistent with the rest of the library but it's certainly worth going back through pretty much everything in stabilizer to at least make the errors produce some observable signal. i've de-prioritized it so far since the policy will, in general, fail safe: not stabilizing something should generally result in _not_ publishing an attestation.

msuozzo: filed #1290

PR #1288, opened 2026-05-28
Add diffr helper for archive test output

msuozzo:


wbxyz (approved):
nit: Can we remove some of the stuttering in the paths? especially diffr/diffrtest/diffrtest_test.go

pkg/diffr/diffrtest/diffrtest.go
```diff
@@ -0,0 +1,42 @@
...
+		diffr.File{Name: "want", Reader: bytes.NewReader(want)},
+		diffr.File{Name: "got", Reader: bytes.NewReader(got)},
+		diffr.Options{Output: &buf},
+	)
+	if err != nil && !errors.Is(err, diffr.ErrNoDiff) {
+		t.Fatalf("diffr.Diff: %v", err)
+	}
+	if buf.Len() == 0 {
+		// Diffr normalized the byte difference away so fall back to cmp.Diff.
+		return "bytes differ but diffr reports no semantic diff; byte-level diff (-want +got):\n" + cmp.Diff(want, got)
```
wbxyz: This case shouldn't happen? IIUC it would be a violation of diffr.Diff to return an empty diff but not return ErrNoDiff.

msuozzo: yep i realized this after the fact. reverted.

msuozzo:
> nit: Can we remove some of the stuttering in the paths? especially diffr/diffrtest/diffrtest_test.go

yeah it's rough but it does conform to the idioms we use in the project and golang's more generally.

PR #1291, opened 2026-05-29
Use HEAD as default checkout target for NativeClone

msuozzo:
As opposed to master, this default will avoid reference not found errors
for repos where the master branch does not exist. This does leave us in
a detached HEAD state but, for rebuild purposes, this is no worse than
the alternative since we're not modifying any content.

### DIFF
--- a/internal/gitx/clone.go
+++ b/internal/gitx/clone.go
@@ -289,7 +289,8 @@ func NativeClone(ctx context.Context, s storage.Storer, fs billy.Filesystem, opt
 		if err != nil {
 			return nil, errors.Wrap(err, "getting worktree")
 		}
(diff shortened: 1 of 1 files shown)


wbxyz:
Will this be problematic if the build reflects on the branch being used for the build?

msuozzo:
> Will this be problematic if the build reflects on the branch being used for the build?

it's possible but these wouldn't have worked anyway when the checked out branch diverged from what was originally built. as an example, if a build records `main` as well as the commit hash, we wouldn't be able to do so without resetting the main branch back to that commit (which we currently do not do).

wbxyz (approved):

PR #1294, opened 2026-06-03
Add initial agent scratch framework

msuozzo:
This is *incomplete* but demonstrates the ability to create and teardown
VMs and exercises our state bookkeeping. Subsquent changes actually
enable us to make use of these constructs for remote execution and state
management.

### DIFF
--- a/cmd/agent-api/main.go
+++ b/cmd/agent-api/main.go
@@ -15,6 +15,7 @@ import (
 	"cloud.google.com/go/storage"
 	"github.com/google/oss-rebuild/internal/api"
 	"github.com/google/oss-rebuild/internal/api/agentapiservice"
+	"github.com/google/oss-rebuild/internal/db"
 	"github.com/google/oss-rebuild/internal/serviceid"
 	buildgcb "github.com/google/oss-rebuild/pkg/build/gcb"
 	"github.com/google/oss-rebuild/pkg/gcb"
@@ -33,6 +34,12 @@ var (
 	prebuildVersion      = flag.String("prebuild-version", "", "golang version identifier of the prebuild binary builds")
 	prebuildAuth         = flag.Bool("prebuild-auth", false, "whether to authenticate requests to the prebuild tools bucket")
 	port                 = flag.Int("port", 8080, "port on which to serve")
+
+	// Scratch flags. Gated by --scratch-enabled.
+	scratchEnabled      = flag.Bool("scratch-enabled", false, "register the scratch routes")
+	scratchZone         = flag.String("scratch-zone", "", "GCE zone for scratch VMs (required when scratch enabled)")
+	scratchStandardTmpl = flag.String("scratch-instance-standard-template", "", "GCE instance template URL for the standard machine class (required when scratch enabled)")
+	scratchJumboTmpl    = flag.String("scratch-instance-jumbo-template", "", "GCE instance template URL for the jumbo machine class (optional)")
 )
 
 // Link-time configured service identity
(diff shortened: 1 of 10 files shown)


deployments/services.tf
```diff
@@ -42,6 +42,46 @@ resource "google_cloudbuild_worker_pool" "jumbo-pool" {
...
+  machine_type = "n2-standard-8"
+  region       = "us-central1"
+
+  service_account {
+    email  = ""
+    scopes = []
+  }
+
+  disk {
+    source_image = "ubuntu-os-cloud/ubuntu-2204-lts"
```
spencerschrock: Is there any reason not to use a more recent LTS?

cmd/agent-api/main.go
```diff
@@ -116,11 +123,70 @@ func AgentCompleteInit(ctx context.Context) (*agentapiservice.AgentCompleteDeps,
...
+		return nil, errors.Wrap(err, "compute client")
+	}
+	return &agentapiservice.ScratchCreateDeps{
+		Scratches: db.NewFirestoreScratch(fs),
+		GCE:       gce,
+		Standard: agentapiservice.ClassConfig{
+			InstanceTemplate: *scratchStandardTmpl,
+			DiskType:         "pd-ssd",
+			DiskSizeGB:       200,
+		},
```
spencerschrock: would there be any benefit in having the disk size as a parameter in `schema.ScratchCreateRequest`?

deployments/services.tf
wbxyz: Can we select a newer lts here? Although maybe this is updated in a followup PR

internal/db/scratch.go
```diff
@@ -0,0 +1,186 @@
...
+	if status.Code(err) == codes.NotFound {
+		return ErrNotFound
+	}
+	return err
+}
+
+func (f *firestoreScratch) ListIdleSince(ctx context.Context, t time.Time) ([]schema.Scratch, error) {
+	q := f.client.Collection(scratchCollection).
+		Where("state", "==", string(schema.ScratchReady)).
+		Where("last_used", "<", t.UTC())
```
spencerschrock, line 87: do we expect there to be VMs that crash in other states (e.g. `deleting`), that we want the ability to cleanup?

internal/api/agentapiservice/scratch_gce.go
```diff
@@ -0,0 +1,208 @@
...
+	DeleteDisk(ctx context.Context, zone, name string) error
+	// ListStoppedInstances returns terminated VMs whose labels match every
+	// entry in labels (subset match).
+	ListStoppedInstances(ctx context.Context, zone string, labels map[string]string) ([]Instance, error)
+	// StartInstance brings a terminated VM up and returns its summary
+	// (including the (possibly new) InternalIP).
+	StartInstance(ctx context.Context, zone, name string) (Instance, error)
+	// StopInstance terminates a running VM (used at replenishment to
+	// return a freshly-inserted VM to stopped state).
+	StopInstance(ctx context.Context, zone, name string) error
```
wbxyz, line 43: Can zone be embeded in computGCE? This can push the zone configuration all the way up to main.go, removes it from all the Deps and function calls, and it leaves this interface much closer to something a local execution backend could implement.

From what I can tell we don't want to use multiple zones during the handling of one request, but we can add it back if we do?

internal/api/agentapiservice/scratch.go
```diff
@@ -0,0 +1,211 @@
...
+		if createdInstance {
+			if err := deps.GCE.DeleteInstance(bg, scratch.Zone, scratch.VMName); err != nil {
+				log.Printf("teardown DeleteInstance(%s): %v", scratch.VMName, err)
+			}
+		}
+		if createdDisk {
+			if err := deps.GCE.DeleteDisk(bg, scratch.Zone, scratch.DiskName); err != nil {
+				log.Printf("teardown DeleteDisk(%s): %v", scratch.DiskName, err)
+			}
+		}
```
wbxyz: Should we just attempt these unconditionally? If an opt times out during creation, then createdInstance might be false even if there's actually an instance that started.

cmd/agent-api/main.go
```diff
@@ -116,11 +123,70 @@ func AgentCompleteInit(ctx context.Context) (*agentapiservice.AgentCompleteDeps,
...
-	http.HandleFunc("/agent/session/complete", api.Handler(AgentCompleteInit, agentapiservice.AgentComplete))
-	if err := http.ListenAndServe(fmt.Sprintf(":%d", *port), nil); err != nil {
+	mux := http.NewServeMux()
+	mux.HandleFunc("/agent/session/iteration", api.Handler(AgentCreateIterationInit, agentapiservice.AgentCreateIteration))
+	mux.HandleFunc("/agent/session/complete", api.Handler(AgentCompleteInit, agentapiservice.AgentComplete))
+
+	if *scratchEnabled {
+		mux.HandleFunc("/scratch/create", api.Handler(ScratchCreateInit, agentapiservice.ScratchCreate))
+		mux.HandleFunc("/scratch/get", api.Handler(ScratchGetInit, agentapiservice.ScratchGet))
+		mux.HandleFunc("/scratch/delete", api.Handler(ScratchDeleteInit, agentapiservice.ScratchDelete))
```
wbxyz, line 182: Should scratch creation be on the standard API? And/or maybe it should just be a parameter on AgentCreate?

I'm concerned that we might not intend to give the agent ability to make multiple scratch instances.

internal/api/agentapiservice/scratch.go
```diff
@@ -0,0 +1,211 @@
...
+			return ClassConfig{}, errors.Errorf("machine_class %q not configured", class)
+		}
+		return *jumbo, nil
+	default:
+		return ClassConfig{}, errors.Errorf("unknown machine_class %q", class)
+	}
+}
+
+// ScratchCreateDeps wires ScratchCreate.
+type ScratchCreateDeps struct {
```
wbxyz, line 46: Should a scratch environment be associated to an agent session? And should we validate that a session doesn't already have a scratch that exists?

wbxyz (approved):

cmd/agent-api/main.go
msuozzo: yeah i can see that being a future extension. i suspect having reasonable defaults per-machine class is workable for now. these are probably overly generous to begin with so we'll probably find it easy to reduce them, in which case we'd want a bail-out to control it.

deployments/services.tf
msuozzo: nope. done.

cmd/agent-api/main.go
msuozzo: eh the scratch management should be feasible to be done within the scaffolding around the agent so i'm not particularly concerned about incorrect usage at this point. that said, i think there's a deeper integration we could make here with agent session but that requires deeper changes i think we can defer. happy to hear ideas of how you might like to see that api evolve.

internal/api/agentapiservice/scratch.go
msuozzo: addressed in the previous comment: it could and would make sense to but we're not there yet.

internal/db/scratch.go
msuozzo: fair point. having that intermediate state does introduce the risk of stragglers when instances/disks fail to delete. the actual cleanup process isn't implemented for another few PRs so another option would be either removing the Deleting state or doing a native sweep of GCE resources that need to be cleaned up. i'd consider these to be more polish and this specific API seems consistent with its purpose (listing the idle resources).

internal/api/agentapiservice/scratch_gce.go
msuozzo: eh i'm not so sure about this. i can certainly imagine cases where we'd be indifferent to where the VMs get provisioned but account for (very real) cases where stock-outs could mean needing to shift provisioning. this also lines up with the shape of the native gcp apis which also require location configuration per request.

internal/api/agentapiservice/scratch.go
msuozzo: mmm good catch. i'll try to figure out a fix

msuozzo: hmm on second thought, I think a context cancellation would propagate to the gcp api which leads me to believe this is a lot less likely. only a timeout occurring right on the cusp of the server-side transaction would leave us with a straggler. I think the GCE-native sweep for abandoned resources i mentioned above could probably better serve this case than over-eager deletion.

internal/api/agentapiservice/scratch_gce.go
wbxyz: Discussed offline, I think the plan is for this to be done when there's an abstraction layer which wraps the GCE interface.

cmd/agent-api/main.go
wbxyz: I was thinking more about either explicitly malicious usage (spawning multiple scratch instances for the same session) or the agent unintentionally over-provisioning.

We can push some limits on scratch usage into the agent scaffolding. My only hesitation is that's an untrusted element of the infrastructure. However, misbehaving with scratch instances would just result in resource consumption rather than attestations so the risk seems minimal.

internal/api/agentapiservice/scratch.go
wbxyz: I meant waitZoneOp, which does not propagate the OpTimeout to gcp even though theres an in flight create op.

deployments/services.tf
wbxyz: Thoughts on ubuntu-os-cloud/ubuntu-2604-lts-amd64? Although, it might be good to use debian given recent ubuntu service reliability issues and the cleanliness of a debian build environment.

internal/api/agentapiservice/scratch.go
msuozzo: ah yeah actually the use of lro here means the underlying operation doesn't necessarily get cancelled. went back to the more pessimistic cleanup.

deployments/services.tf
msuozzo: i mean the ubuntu service issues impacted debian, too, so i'm not sure that's a compelling reason to change. i didn't see the more recent LTS so that's an option although i think we may want to consider lagging a bit since most of our rebuilds are still historical. 26 is only 2 months old so it might be good to wait a bit.

msuozzo: gonna punt on this. we can revisit once we start exercising these machines

PR #1295, opened 2026-06-03
Add initial scratch worker

msuozzo:
This sidecar runs alongside any activity in the scratch environment and
mediates communication with the agent API. Notably, the worker has no
IAM access itself so the only way we get data back to durable storage
is pull-through via the agent API. We authenticate by pre-configuring
the sidecar to expect the agent API credentials and validating a signed
JWT sent alongside each request.

### DIFF
--- a/cmd/scratch-worker/main.go
+++ b/cmd/scratch-worker/main.go
@@ -1,0 +1,72 @@
+// Copyright 2025 Google LLC
+// SPDX-License-Identifier: Apache-2.0
+
+// Command scratch-worker is the per-VM scratch worker. It serves
+// /healthz and /stat over HTTP. The worker has no GCP credentials of
+// its own; the VM template should set service_account = null.
+package main
+
+import (
+	"context"
+	"flag"
+	"log"
+	"net/http"
+	"strings"
+
+	"github.com/google/oss-rebuild/internal/api"
+	"github.com/google/oss-rebuild/internal/api/scratchworkerservice"
+	"github.com/google/oss-rebuild/internal/api/scratchworkerservice/idauth"
+)
+
+var (
+	callerSA         = flag.String("caller-sa", "", "caller service account email (sender of incoming requests)")
+	audience         = flag.String("audience", "", "expected ID-token audience (e.g. https://builder/<vm-name>)")
+	dockerSocketPath = flag.String("docker-socket", "/var/run/docker.sock", "path probed by /stat for Docker daemon presence")
+	diskPaths        = flag.String("disk-paths", "", "comma-separated mount paths reported by /stat")
+	listen           = flag.String("listen", ":8080", "listen address")
+)
+
+// statDeps holds the singleton built once at startup.
+var statDeps *scratchworkerservice.StatDeps
+
+func statInit(_ context.Context) (*scratchworkerservice.StatDeps, error) { return statDeps, nil }
+
+func main() {
+	flag.Parse()
+	if *callerSA == "" {
+		log.Fatalf("--caller-sa is required")
+	}
+	if *audience == "" {
+		log.Fatalf("--audience is required")
+	}
+
+	statDeps = &scratchworkerservice.StatDeps{
+		DockerSocketPath: *dockerSocketPath,
+		DiskPaths:        splitCSV(*diskPaths),
+	}
+
+	mw := idauth.Middleware(idauth.NewGoogleValidator(*callerSA, *audience))
+
+	mux := http.NewServeMux()
+	mux.Handle("/stat", mw(api.Handler(statInit, scratchworkerservice.Stat)))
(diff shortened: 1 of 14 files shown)


deployments/platform.tf
```diff
@@ -313,3 +313,17 @@ resource "google_compute_firewall" "allow_outbound" {
   }
   destination_ranges = ["0.0.0.0/0"]
 }
+
+# Scratch workers: reachable on port 8080 only from agent-api's
+# Direct VPC egress range. Workers have no external IPs.
```
spencerschrock, line 318: I'll have some questions about this offline, in terms of other local services in the project

cmd/scratch-worker/main.go
```diff
@@ -0,0 +1,72 @@
...
+
+	statDeps = &scratchworkerservice.StatDeps{
+		DockerSocketPath: *dockerSocketPath,
+		DiskPaths:        splitCSV(*diskPaths),
+	}
+
+	mw := idauth.Middleware(idauth.NewGoogleValidator(*callerSA, *audience))
+
+	mux := http.NewServeMux()
+	mux.Handle("/stat", mw(api.Handler(statInit, scratchworkerservice.Stat)))
```
spencerschrock, line 51: so is the intent to have a future `/exec` or similar for more open-ended execution?

cmd/scratch-worker/main.go
```diff
@@ -0,0 +1,72 @@
...
+	"strings"
+
+	"github.com/google/oss-rebuild/internal/api"
+	"github.com/google/oss-rebuild/internal/api/scratchworkerservice"
+	"github.com/google/oss-rebuild/internal/api/scratchworkerservice/idauth"
+)
+
+var (
+	callerSA         = flag.String("caller-sa", "", "caller service account email (sender of incoming requests)")
+	audience         = flag.String("audience", "", "expected ID-token audience (e.g. https://builder/<vm-name>)")
```
spencerschrock, line 23: I'm sure it's coming, but i'm curious how these will be minted

cmd/scratch-worker/main.go
spencerschrock: didn't see the child PR, whoops

internal/api/scratchworkerservice/idauth/idauth.go
```diff
@@ -0,0 +1,76 @@
...
+
+// Validator verifies an inbound bearer token. Implementations return the
+// caller's email on success.
+type Validator interface {
+	Validate(ctx context.Context, token string) (email string, err error)
+}
+
+// googleValidator verifies a Google-signed ID token against an expected
+// audience and asserts the email claim matches the configured caller SA.
+type googleValidator struct {
```
wbxyz, line 26: nit: Assert that googleValidator is a Validator?

deployments/services.tf
```diff
@@ -77,6 +77,14 @@ resource "google_compute_instance_template" "scratch-standard" {
   tags   = ["scratch"]
   labels = { purpose = "scratch" }
 
+  metadata = {
+    startup-script = templatefile("${path.module}/scratch_startup.sh", {
+      worker_binary_uri = "gs://${google_storage_bucket.bootstrap-tools.name}/${module.prebuild_images["scratch-worker"].image_version}/scratch-worker"
```
wbxyz, line 83: The scratch_startup.sh assumes the worker_binary_uri is public, so this assumes the boostrap-tools bucket is public. Is there a way to reference the production oss-rebuild bootstrap tools when doing a deployment without re-hosting your own boostrap-tools in a public bucket?

wbxyz (approved):

cmd/scratch-worker/main.go
msuozzo: just a call to idtoken: https://github.com/google/oss-rebuild/pull/1296/changes/b54c4d5ab85f23bbc7ccfe379bbbb86d7714eef9#diff-342b8b9665cd15752ec1c18fa50cd41dffbf0449ecd3eb377473895e79e1d865R139

the SA is implicit since it's the caller

deployments/services.tf
msuozzo: eh it's a TODO in the script. i think we should be able to fix that (probably by adding back an SA to the worker with ~only access to a bootstrap bucket). hard-coding is also an option, though.

internal/api/scratchworkerservice/idauth/idauth.go
msuozzo: it is implicitly asserted via the NewGoogleValidator return type

deployments/platform.tf
msuozzo: will follow up

deployments/services.tf
wbxyz: Will probably be necessary to provide access if we want to deploy to this in a dev environment, but we can figure that out in a future PR if you like.

msuozzo: yeah fair plan. done in #1300

PR #1298, opened 2026-06-03
Only fetch and start the tetragon_sysgraph when needed

wbxyz:
This is the sysgraph monitor which consumes sysgraph events from
tetragon, no tetragon itself which is where the larger overhead comes
from. However, it's still a non-trivial fetch and conatiner start that
happens at the beginning of each build, even though it's usually
unnecessary.


msuozzo (approved):
Oof. Want to add a regression test?

PR #1299, opened 2026-06-03
Add periodic clean up of scratch and execs

msuozzo:
This adds support for out-of-band op finalization and scratch instances
deletion. Instance teardown happens after a fixed period of inactivity
(currently 30m) after the last op ran. Instances with ops outstanding
will never be cleaned up.

### DIFF
--- a/internal/api/agentapiservice/scratch_reap.go
+++ b/internal/api/agentapiservice/scratch_reap.go
@@ -1,0 +1,231 @@
+// Copyright 2025 Google LLC
+// SPDX-License-Identifier: Apache-2.0
+
+package agentapiservice
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/google/oss-rebuild/internal/api"
+	"github.com/google/oss-rebuild/internal/db"
+	"github.com/google/oss-rebuild/pkg/rebuild/schema"
+	pkgerrors "github.com/pkg/errors"
+	"google.golang.org/grpc/codes"
+)
+
+type ScratchReapRequest struct{}
+
+func (ScratchReapRequest) Validate() error { return nil }
+
+// ScratchReapResponse reports the counts from a single reap cycle.
+type ScratchReapResponse struct {
+	ScratchesReaped int `json:"scratches_reaped"`
+	OpsFinalized    int `json:"ops_finalized"`
+}
+
+// opDeadlineGrace pads each op's worker-enforced timeout to allow for
+// dispatch latency, clock skew, and sync slack before the reaper treats
+// the op as expired.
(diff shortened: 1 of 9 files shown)


internal/api/agentapiservice/scratch_reap.go
```diff
@@ -0,0 +1,179 @@
...
+	for _, scratch := range idle {
+		if deps.Syncer != nil {
+			syncPendingFor(ctx, deps, scratch)
+		}
+		if err := teardownScratch(ctx, deps, scratch); err != nil {
+			log.Printf("reap teardown scratch %s: %v", scratch.ID, err)
+			continue
+		}
+		scratchesReaped++
+	}
```
wbxyz, line 124: Parallelize? sync might be IO bound, but teardown might be waiting for GCE responses so there's probably some benefit here.

internal/api/agentapiservice/scratch_reap.go
```diff
@@ -0,0 +1,179 @@
...
+// per-op hard deadline are marked TimedOut. Best-effort: each item's failure
+// is logged and the loop continues so a single bad row can't block the rest.
+func ScratchReap(ctx context.Context, _ ScratchReapRequest, deps *ScratchReapDeps) (*ScratchReapResponse, error) {
+	now := time.Now().UTC()
+	idleCutoff := now.Add(-deps.idleThreshold())
+	opCutoff := now.Add(-deps.opDeadline())
+	// Reap idle scratches. Before tearing down each one, try
+	// to sync any pending ops on it so we capture exit codes while the
+	// worker is still reachable.
+	idle, err := deps.Scratches.ListIdleSince(ctx, idleCutoff)
```
wbxyz, line 98: What exactly does ListIdleSince capture? IIUC it's measuring Scratch.LastUsed which only gets updated when ScratchExecCreate is triggered, not during the life of the execution. Which I think means if an operation takes longer than 30 minutes the reaper could kill it.

Should we update LastUsed as part of the exec/get handler?

wbxyz (approved):

internal/api/agentapiservice/scratch_reap.go
msuozzo: we can't start tearing down the VM while the sync is happening

internal/api/agentapiservice/scratch_reap.go
msuozzo: yeah the semantics changed between initial design and the code and introduced a gap. previously, the reaper was also reaping ops which meant it knew not to turn down a scratch with ops still running. that was removed for simplicity but i think it was actually load-bearing in a few ways so it's now back in. also bump the LastUsed on op end so we aren't too eager about reclaiming a VM after a long-running op completes.

PR #1302, opened 2026-06-08
Add act streaming API

msuozzo:


wbxyz (approved):
I'm surprised the standard http library doesn't support `text/event-stream`.

This seems small enough to be manageable but at some point we might want to consider using an off-the-shelf RPC library for things like this.

PR #1304, opened 2026-06-08
Reimplement Output with streaming

msuozzo:
Note: this doesn't quite get us to the incremental point that we'd like
since we're still re-fetching the full file each time (coming in the
next change). This does, however avoid buffering the full file in the
worker before dispatch.


wbxyz (commented):
Looks good but multi-parent commits makes it hard to view just the relevant change.

wbxyz (approved):

PR #1307, opened 2026-06-09
Fix local git cache

pohlarized:
Fixes multiple issues with the git cache
- The GCS backend returns a 302 found redirect which was always expected, but the local backend returns a 200 response which would default to an error response by the cache
- At least the `ctl infer` command uses `git.DefaultSubmoduleRecursionDepth` which is set to 10, but the git cache explicitly only allowed recursion depth 0
- The cache server would pack the repo files into the tgz so that the contents of the .git dir are in the root. However, when extracting, it would expect them to be within a `.git` directory, and thus always extracted an empty git repo.

Some notes:
- I have no GCS backend so i couldn't / didn't test whether these changes might break the GCS backend (e.g. idk when GCS might return a 200 HTTP code and what that might mean)
- I have removed the recursion depth check entirely. In my few tests this wasn't an issue, but i don't know whether there might be times where this is an issue for the cache. Specifically, i don't know any repos that require submodules so in my tests this probably just didn't have any effect.

### DIFF
--- a/internal/gitcache/client.go
+++ b/internal/gitcache/client.go
@@ -69,12 +69,15 @@ func (c Client) GetLinkWithRef(repo string, contains time.Time, ref string) (uri
 	if err != nil {
 		return "", err
 	}
+	defer resp.Body.Close()
 	var errs []byte
 	switch resp.StatusCode {
 	case http.StatusFound:
 		uri = resp.Header.Get("Location")
 		// FIXME: Figure out why this URL parsing artifact is being reintroduced.
 		return strings.ReplaceAll(uri, "%252F", "%2F"), nil
+	case http.StatusOK:
+		return u.String(), nil
 	case http.StatusBadRequest:
 		errs, err = io.ReadAll(resp.Body)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


msuozzo (approved):

internal/gitcache/client.go
```diff
@@ -91,7 +94,7 @@ func (c Client) GetLinkWithRef(repo string, contains time.Time, ref string) (uri
 
 // Clone provides an interface to clone a git repo using the GitCache.
 func (c Client) Clone(ctx context.Context, s storage.Storer, fs billy.Filesystem, opt *git.CloneOptions) (*git.Repository, error) {
-	if opt.Auth != nil || opt.RemoteName != "" || opt.Depth != 0 || opt.RecurseSubmodules != 0 || opt.Tags != git.InvalidTagMode || opt.InsecureSkipTLS || len(opt.CABundle) > 0 {
```
msuozzo, line 94: hmm so the submodule question is tricky. on bare clone, we don't care about the flag bc it won't do anything. but for `wt.Checkout`, we do now want to propagate the flag there. On the downside, these submodules themselves _won't_ be cached which undercuts this a bit. I think it's fine to document that behavior in the doc comment and pass through the recurse value to CheckoutOptions.

pohlarized: Alright, i should have addressed the issues.
[The go-git `CheckoutOptions` do not have a way to pass any `SubmoduleRecursivity`](https://pkg.go.dev/github.com/go-git/go-git/v5#CheckoutOptions), so i added a manual step where the submodules are initialized and updated after checkout.

pohlarized:
Ah, sorry, i don't use github that much, should I have re-requested a review here? Or resolved the conversation?

PR #1323, opened 2026-06-11
Improve boundary semantics of scratch lifecycle

msuozzo:
This ensures that scratch VMs are eventually reaped, even if failures
occur during insertion or deletion.

> Children: [#1324](https://redirect.github.com/google/oss-rebuild/pull/1324)


wbxyz:
Is this still needed? Maybe slightly related in #1469 which I just merged.

Issue #1327, opened 2026-06-14
Missing User-Agent header causes 403 from crates.io in local execution


DEVSOG12:
Local execution against CratesIO packages fails with `403 Forbidden` at two stages:

1. Registry metadata fetch: `fetching crate metadata: 403 Forbidden`
2. Upstream artifact download: `fetching upstream artifact: 403 Forbidden`

crates.io requires a `User-Agent` header on all requests ([here](https://github.com/rust-lang/crates.io/blob/main/src/middleware/require_user_agent.rs#L1)). Several code paths create registry clients or HTTP requests using bare `http.DefaultClient` without a User-Agent:

Registry API calls (via `NewRegistryMux(http.DefaultClient)`):
- [`tools/benchmark/run/local.go` (lines 64, 95, 139)](https://github.com/google/oss-rebuild/blob/617d637/tools/benchmark/run/local.go#L64)
- [`tools/benchmark/run/run.go` (line 58)](https://github.com/google/oss-rebuild/blob/617d637/tools/benchmark/run/run.go#L58)
- [`tools/benchmark/artifact_backfill/main.go` (line 71)](https://github.com/google/oss-rebuild/blob/617d637/tools/benchmark/artifact_backfill/main.go#L71)
- [`tools/benchmark/generate/main.go` (line 105)](https://github.com/google/oss-rebuild/blob/617d637/tools/benchmark/generate/main.go#L105) using direct `HTTPRegistry{Client: http.DefaultClient}`
- [`tools/ctl/command/export/export.go` (line 162)](https://github.com/google/oss-rebuild/blob/617d637/tools/ctl/command/export/export.go#L162)
- [`tools/ctl/command/tui/tui.go` (line 126)](https://github.com/google/oss-rebuild/blob/617d637/tools/ctl/command/tui/tui.go#L126)

Artifact download (via `DoContext` fallback):
- [`pkg/rebuild/rebuild/http.go` (line 18)](https://github.com/google/oss-rebuild/blob/617d637/pkg/rebuild/rebuild/http.go#L18): `DoContext` fallback
- [`internal/verifier/summary.go` (line 44)](https://github.com/google/oss-rebuild/blob/617d637/internal/verifier/summary.go#L44):  bare `http.NewRequest`

The `httpx.WithUserAgent` wrapper already exists in `internal/httpx/http.go` but is not used in these paths.

### Reproducer
```
ctl run-one attest --local 
[...]

PR #1328, opened 2026-06-14
Add default User-Agent to outbound HTTP requests

DEVSOG12:
Some registries (e.g., crates.io) reject HTTP requests that lack a `User-Agent` header with `403 Forbidden`. This causes local execution (`run-one --local`, `run-bench --local`) to fail for CratesIO packages at two stages: Registry metadata fetch and Upstream artifact download

This PR adds a default `User-Agent` header; see issue #1327

### DIFF
--- a/pkg/rebuild/meta/meta.go
+++ b/pkg/rebuild/meta/meta.go
@@ -24,6 +24,11 @@ import (
 )
 
 func NewRegistryMux(c httpx.BasicClient) rebuild.RegistryMux {
+	// Ensure requests have a User-Agent header. Some registries (e.g.,
+	// crates.io) reject requests without one.
+	if _, ok := c.(*httpx.WithUserAgent); !ok {
+		c = &httpx.WithUserAgent{BasicClient: c, UserAgent: rebuild.DefaultUserAgent}
+	}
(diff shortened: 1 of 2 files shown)


MarlinDiary:
I tested this branch locally and noticed two crates.io tool paths still fail with `403 Forbidden` after this change:

- `go run ./tools/benchmark/generate -only cratesio_top_2000.json -output-dir /tmp/out`
- `go run ./tools/registryscan serde@1.0.219 /tmp/cache`

The benchmark generator uses a bare `http.DefaultClient` for the crates.io listing request before it reaches the per-crate `HTTPRegistry` call, and `registryscan` also constructs a direct crates.io `HTTPRegistry` with `http.DefaultClient`.

Happy to send a small follow-up if that would be useful.

msuozzo:
Hey there! thanks for the PR! i realized we didn't have a good story around this so i made some more systemic changes. #1338 and #1339 implement a mandatory "host" which is an sub-identifier within a project-wide user-agent scope. i think this will more reliably provide the right default while enabling instance/user-specific overrides for dedicated uses.

Again, thanks for bringing this up and for the first-pass at a solution!

PR #1333, opened 2026-06-16
Add manual strategy for OCI ecosystem

wbxyz:


pkg/rebuild/oci/strategy.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func init() {
+	flow.Tools.MustRegister(&flow.Tool{
+		Name: "oci/build",
+		Steps: []flow.Step{{
+			Runs: textwrap.Dedent(`
+				# Install docker CLI if not present
+				if ! command -v docker >/dev/null; then
+					apt-get update && apt-get install -y docker.io
```
msuozzo: hmm i don't think we want this here. if it can't be listed as systemdeps, i think we'd want to check-fail instead of conditionally installing dynamically (especially since aptitude isn't platform-agnostic)

msuozzo (approved):

pkg/rebuild/oci/strategy.go
wbxyz: Yeah this was left over from a previous version. Removing.

PR #1336, opened 2026-07-02
Implement package-level fallback for build definitions

wbxyz:


spencerschrock (approved):

pkg/builddef/set.go
msuozzo, line 35: Hmm thinking about this a bit, should this new default behavior really be built into the interface itself or should it instead be an explicitly supported idiom (empty version and artifact) which is documented here but used externally. Basically, would it make more sense if the caller was responsible for specifying what sort of fallback behavior they want? (i'd lean towards yes)

PR #1338, opened 2026-07-14
Add user-agent on missing paths

msuozzo:


wbxyz (approved):
Is it easy to add a test for this? Might be nice to catch a regression

msuozzo:
> Is it easy to add a test for this? Might be nice to catch a regression

I was hoping the log line (eventually upgraded to an error) would provide the in-band assertion that we're always populating the user agent. testing's a bit tricky given how high in hierarchy the callers are configured and how low in the stack the effect is observed.

PR #1342, opened 2026-07-14
Add firestore indexes to terraform

wbxyz:


deployments/platform.tf
```diff
@@ -179,6 +179,84 @@ resource "google_firestore_index" "scratch-state-last-used" {
   depends_on = [google_app_engine_application.dummy_app]
 }
 
+# Composite indexes for querying attempts
+resource "google_firestore_index" "attempts-ecosystem-package-created" {
```
msuozzo, line 185: nit: wanna include a brief text note about where/how these are used?

msuozzo (approved):
very nice

deployments/platform.tf
wbxyz: Added

PR #1356, opened 2026-07-16
Validate inherited Cargo workspace versions

MarlinDiary:
When a crate inherits its version from `workspace.package`, version validation currently treats it as a wildcard. This can allow a ref for a different workspace version to pass both ref selection and final strategy validation.

Resolve the workspace manifest using `package.workspace` when present, otherwise walk ancestor manifests, and compare `workspace.package.version` with the requested version. Also strip a leading UTF-8 BOM before parsing manifests, matching Cargo behavior.

Testing:
- `go test ./pkg/rebuild/cratesio`
- `go test ./...`


msuozzo (approved):
nice!

PR #1358, opened 2026-07-16
Set Cargo 1.85 floor for Edition 2024 crates

MarlinDiary:
Stable Cargo versions before 1.85 reject package manifests with `edition = "2024"`. If such a crate omits `rust_version`, the publication-date heuristic can select Cargo 1.82–1.84 for releases published before or shortly after Cargo 1.85's release.

This change reads the package edition and applies Cargo 1.85 as a hard lower bound after the existing heuristic bounds. Edition-like values in metadata or multiline strings are ignored.

Testing:
- `go test ./...`
- Repackaged 985 affected releases: Cargo 1.82–1.84 rejected all of them with `feature edition2024 is required`, while Cargo 1.85.0 matched the upstream `Cargo.toml`, `Cargo.lock`, and packaged file set (excluding `.cargo_vcs_info.json`).

### DIFF
--- a/pkg/rebuild/cratesio/hints.go
+++ b/pkg/rebuild/cratesio/hints.go
@@ -5,6 +5,9 @@ package cratesio
 
 import (
 	"regexp"
+
+	"github.com/google/oss-rebuild/internal/semver"
+	"github.com/pelletier/go-toml/v2"
 )
 
 // Compiled regex patterns for detecting Rust version requirements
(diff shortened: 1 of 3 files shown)


msuozzo (approved):

pkg/rebuild/cratesio/infer.go
```diff
@@ -256,6 +265,11 @@ func (Rebuilder) InferStrategy(ctx context.Context, t rebuild.Target, mux rebuil
 	if maxVer != "" && semver.Cmp(rustVersion, maxVer) > 0 {
 		rustVersion = maxVer
 	}
+	// Cargo 1.85 is the first stable release that can package Edition 2024
+	// manifests. Apply this hard requirement after the heuristic bounds.
+	if hasPackageEdition2024(string(cargoTomlText)) && semver.Cmp(rustVersion, "1.85.0") < 0 {
```
msuozzo: is this the appropriate place for this check? would this be better served being moved into detectRustVersionBounds?

msuozzo (approved):

pkg/rebuild/cratesio/infer.go
MarlinDiary: Good call, solved~

PR #1359, opened 2026-07-17
Treat declared Rust version as a lower bound

MarlinDiary:
A declared `rust_version` is the crate's minimum supported Rust version, not necessarily the Cargo version that packaged it. Using it as the initial inference pick skips the existing publish-date heuristic whenever the field is present.

Start from the publish-date estimate in all cases, then let `rust_version` raise that estimate as a lower bound. Structural and lockfile bounds still apply afterwards.

Testing:
- `go test ./...`
- `go vet ./...`

> Parents: [#1348](https://redirect.github.com/google/oss-rebuild/pull/1348)

### DIFF
--- a/pkg/rebuild/cratesio/infer.go
+++ b/pkg/rebuild/cratesio/infer.go
@@ -232,15 +232,20 @@ func (Rebuilder) InferStrategy(ctx context.Context, t rebuild.Target, mux rebuil
 	if ct.Version() != version && ct.Version() != reg.WorkspaceVersion {
 		return nil, errors.Errorf("mismatched version [expected=%s,actual=%s]", version, ct.Version())
 	}
-	rustVersion := vmeta.RustVersion
-	if rustVersion == "" {
-		// NOTE: Give a week's margin to allow for toolchain upgrades. Maybe raise.
-		rustVersion, err = reg.RustVersionAt(vmeta.Created.Add(-7 * 24 * time.Hour))
-		if err != nil {
-			return nil, errors.New("rust version heuristic failed")
+	// NOTE: Give a week's margin to allow for toolchain upgrades. Maybe raise.
+	rustVersion, err := reg.RustVersionAt(vmeta.Created.Add(-7 * 24 * time.Hour))
+	if err != nil {
+		return nil, errors.New("rust version heuristic failed")
+	}
+	// rust_version is the crate's minimum supported Rust version, not
+	// necessarily the toolchain used to publish it.
+	if declared := vmeta.RustVersion; declared != "" {
+		if strings.Count(declared, ".") == 1 {
+			declared += ".0"
 		}
-	} else if strings.Count(rustVersion, ".") == 1 {
-		rustVersion += ".0"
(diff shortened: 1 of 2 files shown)


pkg/rebuild/cratesio/infer.go
```diff
@@ -284,7 +280,7 @@ func (Rebuilder) InferStrategy(ctx context.Context, t rebuild.Target, mux rebuil
 		}
 		resp, err := stub(ctx, cratesregistryservice.FindRegistryCommitRequest{
 			LockfileBase64: base64.StdEncoding.EncodeToString(lockContent),
-			PublishedTime:  vmeta.Updated.Format(time.RFC3339),
+			PublishedTime:  vmeta.Created.Format(time.RFC3339),
```
msuozzo, line 292: IIUC, this seems like a separate change worth calling out in the PR description and/or pulling out into a separate PR.

pkg/rebuild/cratesio/infer.go
```diff
@@ -232,15 +223,20 @@ func (Rebuilder) InferStrategy(ctx context.Context, t rebuild.Target, mux rebuil
...
-		// NOTE: Give a week's margin to allow for toolchain upgrades. Maybe raise.
-		rustVersion, err = reg.RustVersionAt(vmeta.Updated.Add(-7 * 24 * time.Hour))
-		if err != nil {
-			return nil, errors.New("rust version heuristic failed")
+	// NOTE: Give a week's margin to allow for toolchain upgrades. Maybe raise.
+	rustVersion, err := reg.RustVersionAt(vmeta.Created.Add(-7 * 24 * time.Hour))
+	if err != nil {
+		return nil, errors.New("rust version heuristic failed")
+	}
+	// rust_version is the crate's minimum supported Rust version, not
```
msuozzo, line 240: for a separate change, this is probably worth changing in the field name of our struct. "MinRustVersion" wouldn't be much longer but would be quite a bit clearer on its semantics.

msuozzo (approved):

pkg/rebuild/cratesio/infer.go
MarlinDiary: This is inherited from the parent PR, #1348 (noted under Parents in the description), and will drop out of this diff once #1348 lands.

pkg/rebuild/cratesio/infer.go
MarlinDiary: Agreed. I can send the field rename as a follow-up after this stack lands.

msuozzo (approved):

pkg/rebuild/cratesio/infer.go
msuozzo: Ah so that annotation actually comes from github.com/msuozzo/jj-forge which notably adds a `forge-parent` trailer in the commit descriptions of all non-root commits. we have a repo rule configured to prevent us from merging these changes.

so stacking changes, while great in general, is risky without that tooling (unfortunately, this is described nowhere in our documentation...)

PR #1360, opened 2026-07-17
Remove resolver absence version bounds

MarlinDiary:
An explicit `resolver = "2"` provides a Cargo 1.51 lower bound, but the absence of that field does not identify a Cargo version range. Cargo 1.51-1.54 can package old-header manifests without it, and Cargo 1.56-1.63 can package Edition 2021 manifests without it.

Remove the absence-based lower and upper bounds while keeping the explicit resolver lower bound from #1349.

Testing:
- Repacked `lld-pg 0.1.0`: Cargo 1.57 and 1.59 reproduce the published artifact; Cargo 1.64 does not
- `go test ./...`
- `go vet ./...`

> Parents: [#1349](https://redirect.github.com/google/oss-rebuild/pull/1349)

### DIFF
--- a/pkg/rebuild/cratesio/hints.go
+++ b/pkg/rebuild/cratesio/hints.go
@@ -85,15 +85,6 @@ func detectRustVersionBounds(cargoTomlText string) (lo, hi string) {
 	}
 	if hasResolverTwo(cargoTomlText) {
 		lo = max("1.51.0", lo)
-	} else {
-		// If resolver pattern not found, we know the version lies outside the above range
-		if lo > "1.51.0" {
-			if hi >= "1.64.0" { // only raise lo if hi is in range
-				lo = max("1.64.0", lo)
-			}
-		} else if hi < "1.63.0" {
-			hi = min("1.50.0", hi)
-		}
 	}
 	if hi == "999" {
 		hi = ""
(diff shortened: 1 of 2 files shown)


pkg/rebuild/cratesio/hints.go
```diff
@@ -24,6 +26,23 @@ var (
...
+	}
+	var manifest struct {
+		Package struct {
+			Resolver string `toml:"resolver"`
+		} `toml:"package"`
+		Workspace struct {
+			Resolver string `toml:"resolver"`
+		} `toml:"workspace"`
+	}
+	return toml.Unmarshal([]byte(cargoTomlText), &manifest) == nil &&
```
msuozzo, line 61: can this not rely on the regex?

msuozzo: ah wait is this oriented towards parsing it out of the orig files?

MarlinDiary: It currently operates on the packaged Cargo.toml, rather than .orig. The regex was intended as a parsing prefilter, but you’re right that it can miss valid resolver declarations. I’ll remove it in the parent PR, #1349, and update this PR.

MarlinDiary: done～

msuozzo (approved):

PR #1363, opened 2026-07-17
Pre-populate proxy certificate volume at container creation time

andusy:
On Container-Optimized OS (COS), writing certificates into an unstarted container volume (`/var/cache/proxy.crt`) right before `docker start` fails with HTTP 500 because the volume is not yet mounted in the container filesystem. Injecting post-start also fails due to a race condition with short-lived commands.

This PR:
1. **Pre-populate volume at creation (`POST /containers/create`)**: Spawns a temporary helper container attached to the proxy volume to write the certificates (`proxy.crt` / `.jks`) into the volume ahead of time.
2. **Skip pre-start injection**: On `POST /containers/<id>/start`, inspects container mounts and skips Archive API writes if `/var/cache` is backed by a volume.

### DIFF
--- a/internal/proxy/dockerfs/dockerfs.go
+++ b/internal/proxy/dockerfs/dockerfs.go
@@ -94,7 +94,7 @@ func (c Filesystem) Stat(path string) (*FileInfo, error) {
 	case http.StatusNotFound:
 		return nil, fs.ErrNotExist
 	case http.StatusBadRequest:
-		return nil, errors.New("request error: bad parameter")
+		return nil, fs.ErrNotExist
 	default:
 		return nil, errors.Wrap(errors.New(resp.Status), "response error: Unexpected HTTP response")
 	}
--- a/e2e/proxy_test.go
+++ b/e2e/proxy_test.go
@@ -495,6 +495,7 @@ func TestTruststorePatching(t *testing.T) {
 			env := setupProxyTestEnv(t, bin, proxyTestEnvOpts{
 				WithDockerProxy: true,
 			})
+			registerLogDump(t, env)
 
 			// Install docker CLI in the build container so we can invoke docker commands.
 			env.runInBuild(t, "apk add --no-cache docker-cli")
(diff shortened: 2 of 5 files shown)


internal/proxy/dockerfs/dockerfs.go
```diff
@@ -94,7 +94,7 @@ func (c Filesystem) Stat(path string) (*FileInfo, error) {
 	case http.StatusNotFound:
 		return nil, fs.ErrNotExist
 	case http.StatusBadRequest:
-		return nil, errors.New("request error: bad parameter")
+		return nil, fs.ErrNotExist
```
msuozzo: this makes me a little uncomfortable. given you said this was being retried, can we instead use a separate error type that would indicate a non-transient error? i feel like these cases are worth differentiating.

andusy: Is there a error type you had in mind? We can also just throw a generic error like before if you prefer. The issue was mostly in the `createFile` function, so we just need to make sure we properly propagate all underlying errors instead of returning them as `iofs.ErrExist`.

msuozzo: Ugh right. I'm now remembering there was some ephemeral error conditional that was solved with a retryable case (maybe container creation?). might be good to try to recreate that error case to see if it was, indeed, a race condition on startup.

to better understand this change, you're trying to remove retry for this error in order to speed up some fallback?

andusy: Oh I think you might be thinking of a different issue. This PR doesn't touch any retry/fallback logic.

The context is, when running the proxy on COS, docker's `/archive` API throws a `http.StatusBadRequest` when attempting to patch the truststore. In `createFile` we're catching all non-`ErrNotExist` errors and masking it as `iofs.ErrExist`, causing the proxy to skip patching the container. So this PR treats `http.StatusBadRequest` responses as `fs.ErrNotExist` so that we can attempt to write the cert.

msuozzo: ah by fallback i just meant "trying the cert creation anyway". i'm trying to understand why we'd get a 400 and then still want to attempt to patch the truststore. is the 400 due to the symlink?

andusy: I haven't pinpointed where exactly the failure happens, but I think it happens around [FollowSymlinkInScope](https://github.com/moby/sys/blob/symlink/v0.3.0/symlink/fs.go#L51) when we make the `/archive` call. For some reason, it doesn't like that we're trying to resolve the symlink across mount boundaries (ro -> rw), so it throws a 400 and we end up skipping the file write.

I also tested again and realize we don't actually need to pre-resolve the symlink so I'll delete `rewriteDockerSocketBinds` in the next revision.

andusy: Updated the PR to patch the cert post `create` so this is no longer relevant.

msuozzo:
So I think we may have stumbled on an issue resulting from one of the Docker vuln patches that didn't get into the proper backport channel. We may be able to resolve the underlying issue here by addressing the issue in COS instead of changing the logic here.

Regardless, thanks for putting together this fix and I think we may still want to move the proxy in this direction regardless of whether it's used to work around this issue. 👍

Issue #1362, opened 2026-07-17
Cover prerelease versions in crates.io benchmark generation


MarlinDiary:
## Problem

The crates.io benchmark generator treats any version containing `-` as a prerelease:

```go
isPrerelease := strings.ContainsRune(v.Version, '-')
```

This excludes true prereleases from `cratesio_top_2000`, but it also misclassifies normal releases whose build metadata contains a hyphen. Under SemVer, `0.14.7+wasi-0.2.4` is not a prerelease. As of 2026-07-17, every non-yanked `wasi` version inside the generator's five-year window has this form, so regenerating the benchmark selects no version for that crate.

In a 2026-07-09 crates.io database dump, the filter affects:

- 204,700 non-yanked prerelease versions
- 1,304 non-yanked releases with a hyphen only in build metadata

`pkg/rebuild/cratesio.GetVersions` contains the same check, although it currently has no in-repo callers. The active behavior is in `tools/benchmark/generate`.

## Proposal

Use the existing SemVer parser to distinguish prerelease identifiers from build metadata.

For prerelease coverage, would you prefer an opt-in generator flag or a separate benchmark? A separate benchmark would avoid displacing stable versions from `cratesio_top_2000`.

msuozzo:
ah yeah this is sloppy implementation. let's use the semver machinery

MarlinDiary:
cc @jensdietrich @elliott-wen

PR #1365, opened 2026-07-17
Distinguish crates.io prereleases from build metadata

MarlinDiary:
Use the existing SemVer parser to distinguish prerelease identifiers from build metadata in the benchmark generator and GetVersions. This keeps stable versions such as `0.14.7+wasi-0.2.4` while continuing to omit true prereleases.

Invalid registry versions now fail explicitly.

Testing:
- `go run ./ci -v test`
- `go run ./ci -v lint`
- `go run ./ci -v fmt-check`

Related to #1362

### DIFF
--- a/pkg/rebuild/cratesio/rebuild.go
+++ b/pkg/rebuild/cratesio/rebuild.go
@@ -8,8 +8,8 @@ import (
 	"fmt"
 	"io"
 	"sort"
-	"strings"
 
+	"github.com/google/oss-rebuild/internal/semver"
 	"github.com/google/oss-rebuild/pkg/rebuild/rebuild"
 	reg "github.com/google/oss-rebuild/pkg/registry/cratesio"
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


pkg/rebuild/cratesio/rebuild.go
```diff
@@ -25,7 +25,11 @@ func GetVersions(ctx context.Context, pkg string, mux rebuild.RegistryMux) (vers
 	for _, v := range p.Versions {
 		// Omit pre-release versions.
 		// TODO: Support rebuilding pre-release versions.
```
msuozzo: nit: move comments down to just above the predicate skipping `Prerelease != ""`

pkg/rebuild/cratesio/rebuild_test.go
```diff
@@ -0,0 +1,54 @@
...
+			{Version: "0.9.0", Created: date(1)},
+		},
+	}}}
+
+	got, err := GetVersions(context.Background(), "example", mux)
+	if err != nil {
+		t.Fatalf("GetVersions() error = %v", err)
+	}
+	want := []string{"0.14.7+wasi-0.2.4", "1.0.0+build-1", "0.9.0"}
+	if !reflect.DeepEqual(got, want) {
```
msuozzo: we tend to prefer cmp.Diff for comparisons

msuozzo (approved):

pkg/rebuild/cratesio/rebuild.go
MarlinDiary: Done~

pkg/rebuild/cratesio/rebuild_test.go
MarlinDiary: Done~

PR #1366, opened 2026-07-18
Exclude non-crates.io packages from registry resolution

MarlinDiary:
Cargo.lock records root, path, Git, and alternate-registry packages alongside crates.io packages. The registry resolver currently treats every entry as if it came from crates.io. When a non-registry package has the same name and version as a published crate, it can move the selected index commit; its name is also included in git-index timewarp requests.

`agent-doc 0.31.9` reproduces this with its path package `tmux-router 0.3.9`: including that entry moves the selected registry commit.

Record each package's source and use only crates.io entries for registry resolution and `PackageNames`. Lockfiles with no crates.io packages now skip registry lookup.

Testing:
- `go test ./...`
- `go vet ./...`

### DIFF
--- a/pkg/registry/cratesio/cargolock/cargolock.go
+++ b/pkg/registry/cratesio/cargolock/cargolock.go
@@ -6,13 +6,18 @@ package cargolock
 import (
 	"bufio"
 	"fmt"
+	"slices"
+	"strconv"
 	"strings"
 )
 
-// Package represents a crate dependency from a Cargo.lock file.
+const cratesIOIndexSource = "registry+https://github.com/rust-lang/crates.io-index"
+
+// Package represents a package entry from a Cargo.lock file.
 type Package struct {
 	Name    string
 	Version string
+	Source  string
 }
 
 // Lockfile represents a parsed Cargo.lock file.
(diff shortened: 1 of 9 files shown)


pkg/registry/cratesio/cargolock/cargolock.go
```diff
@@ -6,13 +6,17 @@ package cargolock
...
 )
 
-// Package represents a crate dependency from a Cargo.lock file.
+const cratesIOIndexSource = "registry+https://github.com/rust-lang/crates.io-index"
+
+// Package represents a package entry from a Cargo.lock file.
 type Package struct {
 	Name    string
 	Version string
+	Source  string
```
msuozzo, line 20: since you're adding Source here, the necessary data to do this filtering will be present in the existing Parse() call. as a result, you could rewrite this as a slices.DeleteFunc call with that comparison of Source against cratesIOIndexSource. you'd probably want to wrap that in a free function here instead of requiring the slices call but you could avoid adding it as a part of the Lockfile interface.

MarlinDiary: Done~

msuozzo (approved):

pkg/registry/cratesio/cargolock/cargolock.go
```diff
@@ -23,35 +28,69 @@ type Lockfile struct {
 	Packages      []Package
 }
 
+// CratesIOPackages returns packages resolved from the crates.io registry.
+// The input slice is modified in place.
+func CratesIOPackages(packages []Package) []Package {
+	return slices.DeleteFunc(packages, func(pkg Package) bool {
```
msuozzo, line 34: ah wait this is an in-place filter but the signature returns a value and the name implies more of a pure function. Building a separate slice is probably the way to go OR change the name here to "FilterCratesIOPackages"

MarlinDiary: fixed!

msuozzo (approved):
nice!

PR #1373, opened 2026-07-23
Refine registry search after coarse scan

MarlinDiary:
When the coarse scan reaches the repository root without observing a drop in the match count, `FindRegistryResolution` returns the last sampled commit. Commits between that sample and the root are never checked, so the result can be later than the earliest matching state.

Track the oldest commit reached by the coarse scan and probe it before deciding whether a fine-grained scan is needed. If the root has the same match count as the baseline, return it directly; otherwise refine from the last complete coarse sample. The regression case covers multiple matching commits within one coarse interval.

Testing:
- `go test ./pkg/registry/cratesio/index`
- `go test ./...`
- `go vet ./...`

### DIFF
--- a/pkg/registry/cratesio/index/find.go
+++ b/pkg/registry/cratesio/index/find.go
@@ -177,28 +177,21 @@ func findCommitWithVersions(repo *git.Repository, packages []internalPackage, pu
 	// until we find a drop in the number of matches.
 	day := 24 * time.Hour
 	nextCheckTime := firstCommit.Committer.When.Add(-day)
-	foundDrop := false
 	for c, err := range iterx.ToSeq2(commitIter, io.EOF) {
 		if err != nil {
 			return nil, errors.Wrap(err, "iterating over daily commits")
 		}
 		if c.Committer.When.Before(nextCheckTime) {
 			if matchesFor(c) < maxFound {
-				foundDrop = true
 				break
 			}
 			upperBoundCommit = c
 			nextCheckTime = c.Committer.When.Add(-day)
 		}
 	}
-	if !foundDrop {
-		return &searchResult{
-			ResolutionCommit: upperBoundCommit,
-			ResolvableCrates: maxFound,
-			PriorCommit:      nil,
-		}, nil
-	}
-	// Scan backwards through that day's commits again to find the exact drop
+	// Scan backwards through the remaining commits to find the exact drop.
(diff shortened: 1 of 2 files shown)


pkg/registry/cratesio/index/find.go
```diff
@@ -177,28 +177,21 @@ func findCommitWithVersions(repo *git.Repository, packages []internalPackage, pu
 	// until we find a drop in the number of matches.
 	day := 24 * time.Hour
 	nextCheckTime := firstCommit.Committer.When.Add(-day)
```
msuozzo, line 179: what about instead always checking the root commit in the predicate?

```
lastCommit := firstCommit
foundDrop := false
for c, err := range iterx.ToSeq2(commitIter, io.EOF) {
	if err != nil {
		return nil, errors.Wrap(err, "iterating over daily commits")
	}
	lastCommit = c
	if c.Committer.When.Before(nextCheckTime) {
		if matchesFor(c) < maxFound {
			foundDrop = true
			break
		}
		upperBoundCommit = c
		nextCheckTime = c.Committer.When.Add(-day)
	}
}
// The coarse scan leaves up to a day of commits above the root unexamined;
// probing the root directly settles the all-matches case without a rescan.
if !foundDrop && matchesFor(lastCommit) == maxFound {
	return &searchResult{ResolutionCommit: lastCommit, ResolvableCrates: maxFound}, nil
}
```

pkg/registry/cratesio/index/find.go
```diff
@@ -191,14 +193,18 @@ func findCommitWithVersions(repo *git.Repository, packages []internalPackage, pu
...
+	if !foundDrop && matchesFor(oldestCommit) == maxFound {
 		return &searchResult{
-			ResolutionCommit: upperBoundCommit,
+			ResolutionCommit: oldestCommit,
 			ResolvableCrates: maxFound,
 			PriorCommit:      nil,
 		}, nil
 	}
-	// Scan backwards through that day's commits again to find the exact drop
+	// Scan backwards through the remaining commits to find the exact drop.
```
msuozzo: i think we could probably revert this docs change given the new impl.

msuozzo (approved):
nice find!

pkg/registry/cratesio/index/find.go
MarlinDiary: Okay~

PR #1374, opened 2026-07-23
Handle multiple versions sharing an index path

MarlinDiary:
A Cargo.lock can contain multiple versions of the same crate. These entries share one crates.io index path, but `findCommitWithVersions` caches presence only by path. The result for the first version is then reused for the others, which can overcount matches and select an older snapshot that does not contain every requested version.

Group lockfile entries by index path so each blob is still read once, while tracking presence separately for each version. The regression test covers two versions in the current index and only one in the older snapshot.

Testing:
- `go test ./pkg/registry/cratesio/index`
- `go test ./...`
- `go vet ./...`


msuozzo (approved):
nice fix!

PR #1375, opened 2026-07-23
Anchor registry search at the target version

MarlinDiary:
The crates.io API `created_at` can precede the target version's index update. Using it as a hard upper bound can make registry inference fail or select a state that is missing exact lockfile pins.

Pass the target package and version to the registry service, locate the target's first index-visible commit, and use that commit as the upper bound for the existing lock search. If the target is not in the initially selected snapshot, retry once with the next newer snapshot or current index. Requests without target information retain the existing timestamp-based behavior.

Testing:
- `go test ./...`
- `go vet ./...`
- Replayed 11 affected releases: all selected complete lock-relevant views matching the target-time control.
- Exercised 4 cross-snapshot and 5 same-day cases through `FindRegistryCommit`; all returned complete views, and all same-day commits matched direct lookup.
- Audited the target upper bound across 441 cases; no lockfile pins were missing.

### DIFF
--- a/internal/api/cratesregistryservice/findcommit.go
+++ b/internal/api/cratesregistryservice/findcommit.go
@@ -6,6 +6,7 @@ package cratesregistryservice
 import (
 	"context"
 	"encoding/base64"
+	"sort"
 	"time"
 
 	"github.com/go-git/go-git/v5"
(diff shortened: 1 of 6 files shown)


msuozzo:
Generally looks good but needs to integrate the changes/merges that have been made.

MarlinDiary:
Rebased onto current main. 🤗

PR #1379, opened 2026-07-23
Add scratch ctl surface

msuozzo:


wbxyz:
Wdyt of making a dedicated command for this outside of ctl? Could leave the implementation where it is and just mirror the command outside? I could send a PR if you're interested.

msuozzo:
> Wdyt of making a dedicated command for this outside of ctl? Could leave the implementation where it is and just mirror the command outside? I could send a PR if you're interested.

eh that feels like a thing you could keep locally as a helper (given it would only be like 25 lines of code). not sure it really fits with the pattern we've used for ctl.

wbxyz (approved):

PR #1380, opened 2026-07-23
Ensure native clone disables reftable storage

msuozzo:
This mode is not yet available in a go-git release.

### DIFF
--- a/internal/gitx/clone.go
+++ b/internal/gitx/clone.go
@@ -256,6 +256,10 @@ func NativeClone(ctx context.Context, s storage.Storer, fs billy.Filesystem, opt
 	args = append(args, opt.URL, targetDir)
 	// Execute the git command
 	cmd := exec.CommandContext(ctx, "git", args...)
+	// NOTE: go-git cannot read the reftable refstorage format so force "files".
+	// See https://github.com/go-git/go-git/issues/1827
+	// Unlike --ref-format, this var is a no-op on git versions predating reftable.
+	cmd.Env = append(os.Environ(), "GIT_DEFAULT_REF_FORMAT=files")
 	output, err := cmd.CombinedOutput()
 	if err != nil {
 		return nil, classifyCloneError(output, err)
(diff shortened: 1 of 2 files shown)


internal/gitx/clone.go
```diff
@@ -256,6 +256,10 @@ func NativeClone(ctx context.Context, s storage.Storer, fs billy.Filesystem, opt
 	args = append(args, opt.URL, targetDir)
 	// Execute the git command
 	cmd := exec.CommandContext(ctx, "git", args...)
+	// NOTE: go-git cannot read the reftable refstorage format so force "files".
+	// See https://github.com/go-git/go-git/issues/1827
+	// Unlike --ref-format, this var is a no-op on git versions predating reftable.
```
spencerschrock, line 261: the docs made it seem like both do the same thing?

internal/gitx/clone_test.go
```diff
@@ -216,6 +216,39 @@ commits:
 	}
 }
 
+func TestNativeClone_ReftableDefaultHost(t *testing.T) {
+	if !NativeGitAvailable() {
+		t.Skip("native git not available")
+	}
+	// Hosts may default new repos to the reftable format which go-git cannot
+	// read. NativeClone must override it for the cloned repo to be usable.
+	t.Setenv("GIT_DEFAULT_REF_FORMAT", "reftable")
```
spencerschrock, line 225: does this require any particular version of `git`? I assume we're fine with that dependency?

spencerschrock (approved):

internal/gitx/clone.go
msuozzo: right. but arg parsing on prior versions apparently rejects the --ref-format flag.

internal/gitx/clone_test.go
msuozzo: this test won't really mean anything on older git versions or those without the extension. this is more just the ability to test it at all for those platforms where it is relevant.

internal/gitx/clone.go
spencerschrock: SGTM

PR #1382, opened 2026-07-24
Record repo size metrics on inference clone

msuozzo:
Persist per-repository size measurements to a repo_metrics Firestore
collection keyed by canonical repo URI. The inference service measures
each successful clone (packed object store bytes, commit count, head)
in a detached goroutine, including when strategy inference itself fails
since oversized repos are the ones sizing decisions most need measured.

### DIFF
--- a/internal/api/inferenceservice/infer.go
+++ b/internal/api/inferenceservice/infer.go
@@ -6,8 +6,10 @@ package inferenceservice
 import (
 	"context"
 	"log"
+	"time"
 
 	"github.com/google/oss-rebuild/internal/api/cratesregistryservice"
+	"github.com/google/oss-rebuild/internal/db"
 	"github.com/google/oss-rebuild/internal/gitcache"
 	"github.com/google/oss-rebuild/internal/gitx"
 	"github.com/google/oss-rebuild/internal/httpx"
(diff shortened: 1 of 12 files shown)


internal/db/repometrics.go
```diff
@@ -0,0 +1,44 @@
...
+// map to a single document. Both the resource key and the value's URI field
+// must already be canonicalized with uri.CanonicalizeRepoURI.
+type RepoMetrics = Resource[schema.RepoMetrics, string]
+
+const repoMetricsCollection = "repo_metrics"
+
+// repoMetricsDocID hashes an already-canonical repo URI into its document ID.
+func repoMetricsDocID(canonicalURI string) string {
+	sum := sha256.Sum256([]byte(canonicalURI))
+	return hex.EncodeToString(sum[:])
```
wbxyz: Any desire to include something human-readable here? Perhaps the base URL and the first two path elements?

internal/api/inferenceservice/infer.go
```diff
@@ -86,7 +121,14 @@ func Infer(ctx context.Context, req schema.InferenceRequest, deps *InferDeps) (*
 	if !ok {
 		return nil, api.AsStatus(codes.InvalidArgument, errors.New("unsupported ecosystem"))
 	}
-	s, err := doInfer(ctx, rebuilder, t, mux, req.LocationHint(), repoOpt)
+	s, rcfg, err := doInfer(ctx, rebuilder, t, mux, req.LocationHint(), repoOpt)
+	// Record repo metrics even when inference failed: oversized repos that
+	// break inference are the ones sizing decisions most need measured.
```
wbxyz: Won't doInfer cause OOM and take the whole process down, preventing the code from making it here?

wbxyz (approved):
Nice, this is great data to keep track of and it'll make debugging and prioritizing packages much easier.

internal/db/repometrics.go
msuozzo: yeah i think we shouldn't make this opaque. it's not crucial since the URI is stored on the value but i'm unconvinced there's much of a benefit to hashing it.

internal/api/inferenceservice/infer.go
msuozzo: ah shoot this used to be at the top of the function (but in the background) and got moved to synchonous. should have gone above doInfer for exactly that reason.

msuozzo: hmm ah i remember why: Repository isn't present until after. going to leave this as a TODO and come back to it as it requires a little surgery.

internal/db/repometrics.go
msuozzo: k url.PathEscape should do it.

PR #1383, opened 2026-07-25
Write explicit lockfile to Cargo workspace root

MarlinDiary:
Explicit lockfiles are currently written at the checkout root. For packages in a subdirectory or nested workspace, Cargo reads the workspace-root lock instead, so the supplied lockfile can be ignored.

Install the selected toolchain before lockfile creation and use `cargo metadata` from the package directory to locate the workspace root. If the metadata does not expose one, preserve the existing checkout-root behavior.

Testing:
- `go test ./...`
- `go vet ./...`
- Verified root packages, standalone subdirectories, root workspace members, and nested workspace members across eight Rust versions: correct placement increased from 16/32 to 32/32. Rust 1.13 and 1.23 toolchains preserve the checkout-root fallback.

### DIFF
--- a/pkg/rebuild/cratesio/strategy.go
+++ b/pkg/rebuild/cratesio/strategy.go
@@ -40,15 +40,16 @@ func (b *CratesIOCargoPackage) ToWorkflow() *rebuild.WorkflowStrategy {
 		}},
 		Deps: []flow.Step{
 			{
-				Uses: "cargo/create-lockfile",
-				With: map[string]string{
-					"lockfile": lockfile,
-				},
-			},
-			{
 				Uses: "cargo/deps/toolchain",
 				With: map[string]string{
 					"rustVersion": b.RustVersion,
+				},
+			},
+			{
+				Uses: "cargo/create-lockfile",
+				With: map[string]string{
+					"dir":      b.Location.Dir,
+					"lockfile": lockfile,
 				},
 			},
 			{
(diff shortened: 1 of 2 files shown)


msuozzo:
Yeah this definitely looks like it would be necessary. Are you running into cases for which you need to provide an explicit lockfile? I believe one issue was that their size can get so large that various parts of the system (http header limits, db column limits, etc) ended up tripping and we weren't really able to continue on with this approach.

If you could provide specific crates, that would be useful to dig into.

MarlinDiary:
Yes. I found this while testing the existing ExplicitLockfile path on real packages.

For kwaai-compression 0.4.57, writing the lockfile at the checkout root is ignored and packaging fails; writing the same lockfile to core/Cargo.lock produces a stabilized match. I saw the same placement issue with deadpool-lapin, az-cvm-vtpm, embassy-usb-synopsys-otg, and spikes. Those four still package successfully, but the lockfile is silently ignored and the packaged lock comes out different.

The size concern is real. These lockfiles range from about 8 to 77 KiB, and forc-doc 0.70.1 has a 157 KiB lockfile.

For a general fallback, I don't think the lock bytes should live in the strategy. One option would be to extract the lock from the published .crate, store it as a content-addressed build input, and pass only a reference and digest to the builder for verification. That transport would need a separate design.

This PR only fixes where an existing ExplicitLockfile is written.

msuozzo:
Okay yeah i think we should instead deprecate the field officially and remove support for it. It was used prior to the introduction of the custom stabilizer ReplacePattern (and ExcludePath) which can be used to document any undesirable changes. #1426 is a fix kinda along these lines: it should fix the cases where the workspace dir impacts the output location which we're not currently following correctly.

MarlinDiary:
I'll close this PR. Thanks for the context. 🤗

PR #1384, opened 2026-07-25
Isolate historical Cargo builds from repository config

MarlinDiary:
Cargo loads `.cargo/config*` from the current working directory and its ancestors. During a historical rebuild, repository-local config can therefore override the timewarp registry configured at `/.cargo/config.toml`.

When supported, pass the timewarp crates.io replacement through Cargo's `--config` option. This gives the timewarp configuration command-line precedence while preserving repository settings needed for named registries and `rust-toolchain*` overrides. Older Cargo versions retain the isolated `/` fallback. Non-timewarp builds are unchanged.

Testing:
- `go test ./...`
- `go vet ./...`
- Checked Cargo 1.34-1.85 and 1.97 across root, subdirectory, and workspace layouts. The CLI-config path succeeded in all 72 supported cases; older Cargo versions selected the isolated fallback.
- Five published cases using `tuna` or `proton` registry aliases passed with CLI config and failed when repository config was hidden.

### DIFF
--- a/pkg/rebuild/cratesio/strategy.go
+++ b/pkg/rebuild/cratesio/strategy.go
@@ -125,8 +125,24 @@ var toolkit = []*flow.Tool{
 		Steps: []flow.Step{{
 			Runs: textwrap.Dedent(`
 				{{if and (ne .TimewarpHost "") (ne .With.registryCommit "") -}}
-				manifest="$PWD{{if and (ne .Location.Dir ".") (ne .Location.Dir "")}}/{{.With.dir}}{{end}}/Cargo.toml"
-				(cd / && /root/.cargo/bin/cargo package --manifest-path "$manifest" --no-verify)
+				package_dir="$PWD{{if and (ne .With.dir ".") (ne .With.dir "")}}/{{.With.dir}}{{end}}"
+				manifest="$package_dir/Cargo.toml"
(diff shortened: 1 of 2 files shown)


msuozzo:
Hmm we've generally tried to stay away from build script conditional so i think we'd want to address this sort of issue in a different way. One option could be requiring local build definitions be written for these packages outside of inference. Another could be adding an option to the strategy itself, to be set in the inference, which would enable this more involved build behavior only for the packages where it's necessary.

MarlinDiary:
I looked more closely at the five affected packages. They need more than repository config: cargo package also has to reach the tuna or proton registry. With crates.io held constant on a local registry, all five fail offline on a missing package from the alternate registry.

So local build definitions seem like a better fit than enabling this from inference. I can simplify this PR to always isolate repository Cargo config and remove the runtime conditional. Does that sound right?

msuozzo:
I think we might need something even more involved than just following rust-toolchain, if present. `stable`, `beta`, and `nightly` would all need separate handling with only `stable` being readily supportable. I think we may need to add toolchain resolution to inference and then do the `cd /` dance to avoid pulling in the user-defined toolchain. a bit larger scope than the simple script addition but should address the underlying issue.

MarlinDiary:
Thanks, that makes sense. I’ll close this PR for now~

PR #1386, opened 2026-07-25
Preserve absent Cargo.lock in rebuilt crates

MarlinDiary:
Cargo 1.84 and later include Cargo.lock in packaged libraries by default. When the published .crate has no root Cargo.lock, rebuilding with a newer Cargo can therefore add a file that was not published.

Record lockfile absence during inference and pass --exclude-lockfile when the selected Cargo supports it. Older Cargo versions retain the existing behavior.

Testing:
- go test ./...
- go vet ./...
- Rebuilt appwrite 0.10.0, pyo3-error 0.9.0, and cmd_lib_macros 1.9.6 from their Git revisions. Default packaging added only Cargo.lock; --exclude-lockfile matched all three published archives.
- Checked Cargo 1.83, 1.86, and 1.87: 1.83 omits the lockfile by default, 1.86 rejects --exclude-lockfile, and 1.87 honors it.

### DIFF
--- a/pkg/rebuild/cratesio/strategy.go
+++ b/pkg/rebuild/cratesio/strategy.go
@@ -22,6 +22,7 @@ type CratesIOCargoPackage struct {
 	rebuild.Location
 	RustVersion      string            `json:"rust_version" yaml:"rust_version,omitempty"`
 	ExplicitLockfile *ExplicitLockfile `json:"explicit_lockfile" yaml:"explicit_lockfile,omitempty"`
+	ExcludeLockfile  bool              `json:"exclude_lockfile,omitempty" yaml:"exclude_lockfile,omitempty"`
 	RegistryCommit   string            `json:"registry_commit,omitempty" yaml:"registry_commit,omitempty"`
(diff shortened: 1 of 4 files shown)


pkg/rebuild/cratesio/strategy.go
```diff
@@ -124,8 +126,17 @@ var toolkit = []*flow.Tool{
 		Name: "cargo/build/package",
 		Steps: []flow.Step{{
 			Runs: textwrap.Dedent(`
+				{{if eq .With.excludeLockfile "true" -}}
+				package_dir="$PWD{{if and (ne .With.dir ".") (ne .With.dir "")}}/{{.With.dir}}{{end}}"
+				lockfile_flag=
+				if (cd "$package_dir" && /root/.cargo/bin/cargo package --help | grep -q -- '--exclude-lockfile'); then
```
msuozzo: do you need this conditional? could we potentially crash when the flag is not supported? we could either detect that in validating the strategy or have it be a runtime error when the flag isn't found.

MarlinDiary: Updated~ 🥹

msuozzo (approved):

PR #1388, opened 2026-07-27
Copy GCB build logs into the build metadata store

msuozzo:
Enable log copy (previously deliberately omitted) into the per-build
metadata location. The builtin io.Copy detects sources conforming to the
WriteTo interface so gcsx.ObjectReader.WriteTo to
storage.ObjectHandle.CopierFrom is invoked and runs the copy server-side
via object rewrite, a metadata-only operation for same-location
same-class buckets, so no log bytes stream through the service.


wbxyz (approved):

wbxyz:
This is great, happy to see the data layout becoming simplified.

PR #1391, opened 2026-07-27
Fix gitcache GCS redirect escaping

msuozzo:
The previous behavior query-escaped the object path and set the escaped
string as both Path and RawPath. URL rendering only honors RawPath when
it decodes to Path, so the mismatched pair is discarded and Path escaped
again, double-escaping the slashes present in every cache path and
naming a nonexistent object. Query escaping also encoded spaces as plus
signs where the path form requires %20.

### DIFF
--- a/internal/gitcache/backend.go
+++ b/internal/gitcache/backend.go
@@ -64,13 +64,16 @@ func (g *gcsBackend) serve(rw http.ResponseWriter, req *http.Request, path strin
 		http.Error(rw, "Internal Error", 500)
 		return
 	}
-	rawPath := fmt.Sprintf("download/storage/v1/b/%s/o/%s", g.bucket, url.QueryEscape(path))
 	redirect := url.URL{
-		Scheme:   "https",
-		Host:     "storage.googleapis.com",
-		Path:     rawPath,
-		RawPath:  rawPath,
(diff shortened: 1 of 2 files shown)


internal/gitcache/backend.go
```diff
@@ -64,13 +64,16 @@ func (g *gcsBackend) serve(rw http.ResponseWriter, req *http.Request, path strin
...
-		Host:     "storage.googleapis.com",
-		Path:     rawPath,
-		RawPath:  rawPath,
-		RawQuery: fmt.Sprintf("generation=%d&alt=media", a.Generation),
+		Scheme: "https",
+		Host:   "storage.googleapis.com",
+		// The object is a single path segment in the JSON API: RawPath
+		// carries its escaped form and Path the decoded form, matching the
+		// GCS client's httpStorageClient.newRangeReaderXML. Escaping rules:
+		// https://cloud.google.com/storage/docs/request-endpoints#encoding
```
wbxyz, line 73: Is it possible to add a test for this?

wbxyz (approved):

internal/gitcache/backend.go
msuozzo: yeah good call there should be something here. tough to test the actual behavior against GCS but we should at least be able to catch the prior behavior's error.

PR #1396, opened 2026-07-28
Add scratch build implementation

msuozzo:


wbxyz (approved):

pkg/build/scratch/executor.go
```diff
@@ -0,0 +1,540 @@
...
+// command runs on until its worker-enforced timeout (build.CancelDetached
+// semantics). Every dispatched exec therefore carries a nonzero timeout.
+type Executor struct {
+	scratchID        string
+	stubs            Stubs
+	gcsClient        *gcs.Client
+	workDir          string
+	planner          build.Planner[*local.DockerRunPlan]
+	maxParallel      int
+	semaphore        chan struct{}
```
wbxyz, line 105: Can you add a docstring on what the semantics of this are?

pkg/build/scratch/executor.go
```diff
@@ -0,0 +1,540 @@
...
+// runBuild drives one build on the scratch VM. The container is stopped and,
+// unless RetainContainer is set, removed after the phases. The staging
+// directory is always retained for post-mortem inspection.
+func (e *Executor) runBuild(ctx context.Context, handle *scratchHandle, plan *local.DockerRunPlan, t rebuild.Target, opts build.Options, timeout time.Duration) error {
+	dir := path.Join(e.workDir, buildsSubdir, handle.id)
+	container := "rb-" + handle.id
+	prepareScript := strings.Join([]string{
+		"set -eu",
+		fmt.Sprintf("mkdir -p %q", dir+"/out"),
+		"docker ps -aq --filter name=^rb- | xargs -r docker rm -f",
```
wbxyz, line 265: AFAICT this would forcefull kill any already running build? So not only is the default 1, but I think the max concurrent build is also 1?

Might be worth adding `--filter status=exited` to the query.

wbxyz (approved):

pkg/build/scratch/executor.go
msuozzo: the idea here is to be able to reap previous runs that may have failed or been abandoned (given there are cases when we can't guarantee cleanup). i think we're generally going to be running this single-threaded and any co-hosting we'd do on VMs would be in separate vmlets so this feels like a tolerable default. there's definitely a better tracking/scheduling scheme we could use to, for instance, track prior attempts that weren't cleaned up and isolate this step to those but i think this fits our current level of maturity.

pkg/build/scratch/executor.go
msuozzo: This is actually the same structure used in the local executor variants which also lack a good explanation. let me do a larger docs pass in a follow-on.

msuozzo: #1411

pkg/build/scratch/executor.go
msuozzo: also less of an issue in the docker build variant forthcoming in the chain since there won't be a persistent container active after the build.

wbxyz: Given the reality that there can only ever be 1 concurrent build, should we throw an error if NewExecutor is called with a value besides 0 or 1?

PR #1397, opened 2026-07-29
Populate phase timings in docker build variants

msuozzo:
Docker history layer deltas supply the Setup, Source, and Deps phase
durations, one layer per phase under the split Dockerfile structure.
Container state clocks from a post-build inspect supply the Build phase
on both backends. No log parsing is involved. Extraction is
all-or-nothing: a record exists iff every phase was determined,
otherwise the result carries nil rather than a partial or estimated
figure.

NOTE: We omit timing support in the 'docker run' builder for now.

### DIFF
--- a/pkg/build/local/build_planner.go
+++ b/pkg/build/local/build_planner.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/google/oss-rebuild/internal/textwrap"
 	"github.com/google/oss-rebuild/pkg/build"
+	"github.com/google/oss-rebuild/pkg/build/timing"
 	"github.com/google/oss-rebuild/pkg/rebuild/rebuild"
 	"github.com/pkg/errors"
 )
@@ -28,6 +29,7 @@ type dockerBuildContainerArgs struct {
 }
 
 // dockerBuildDockerfileTpl generates Dockerfiles for local Docker builds
+// NOTE: Layer mappings must be kept in sync with dockerfileLayers.
 var dockerBuildDockerfileTpl = template.Must(
 	template.New("docker build dockerfile").Funcs(template.FuncMap{
 		"indent": func(s string) string { return strings.ReplaceAll(s, "\n", "\n ") },
@@ -79,6 +81,16 @@ var dockerBuildDockerfileTpl = template.Must(
 			`)[1:], // remove leading newline
 	))
 
+// dockerfileLayers declares dockerBuildDockerfileTpl's instruction shape.
+// NOTE: Template conditionals vary only script contents, never the sequence.
+func dockerfileLayers(hasDeps bool) timing.Layers {
+	l := timing.Layers{Appended: 5, Setup: 0, Source: 1, Deps: -1}
+	if hasDeps {
+		l.Appended, l.Deps = 6, 2
+	}
+	return l
+}
+
 // DockerBuildPlanner generates Docker build execution plans
 type DockerBuildPlanner struct{}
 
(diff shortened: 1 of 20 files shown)


pkg/rebuild/rebuild/models.go
```diff
@@ -83,6 +83,14 @@ type Input struct {
 	Strategy Strategy
 }
 
+// BuildTimings are the measured durations of each build phase.
+type BuildTimings struct {
+	Setup  time.Duration
+	Source time.Duration
+	Deps   time.Duration // legitimately zero when the plan had no deps layer
```
wbxyz, line 90: Is deps the only one that could ever be empty?

pkg/build/timing/history.go
```diff
@@ -0,0 +1,88 @@
...
+	layers Layers
+}
+
+// ParseHistory extracts this build's layer completion times from
+// `docker history --human=false --format '{{json .}}'` output, taking the
+// newest Appended rows by count.
+// NOTE: History clocks are second-granularity, so durations derived from
+// their deltas carry up to a second of error per boundary.
+// NOTE: Non-JSON lines are skipped so the same parser serves raw command
+// output and a step log interleaving set -x traces and inspect output.
```
wbxyz, line 27: Is this common? Maybe the docker commands should be run in a subshell piped to a var, then the var can be logged without interleaved output?

Generally this feels like it's a brittle way to capture the details needed.

pkg/build/local/build_planner.go
```diff
@@ -79,6 +80,16 @@ var dockerBuildDockerfileTpl = template.Must(
...
 
+// dockerfileLayers declares dockerBuildDockerfileTpl's instruction shape.
+// NOTE: Template conditionals vary only script contents, never the sequence.
+func dockerfileLayers(hasDeps bool) timing.Layers {
+	l := timing.Layers{Appended: 5, Setup: 0, Source: 1, Deps: -1}
+	if hasDeps {
+		l.Appended, l.Deps = 6, 2
+	}
+	return l
+}
```
wbxyz, line 92: Oof, this feels like it could accidentally be broken in the future. An inadvertent change to the docker template could invalidate these numbers which would fail in a non-noisy way IIUC.

It's at least worth a comment above the dockerfile to warn future editors.

wbxyz:
You mentioned an alternative way of implementing this would be to run a server during the build and make network requests during each of the phases.

Want to mention why the docker history inspection is the preferred implementation?

wbxyz (approved):

pkg/rebuild/rebuild/models.go
msuozzo: in general, no. in practice, yes. we don't yet have any cases without a source extraction or a build so this is functionally the only nullable state.

pkg/build/timing/history.go
msuozzo: not common and not likely in any of our use-cases. i believe the reason i included this is we still have some log spam (although i'd like to go through and remove it) related to sysgraph trace collection although that's really non-critical here.

pkg/build/local/build_planner.go
msuozzo: eh this would require a pretty substantial change to the planner and it's co-located with that logic. i'll add another comment by the dockerfile predicate to be sure.

msuozzo:
> You mentioned an alternative way of implementing this would be to run a server during the build and make network requests during each of the phases.
> 
> Want to mention why the docker history inspection is the preferred implementation?

Yeah I tried that in addition to pure logs parsing (e.g. parsing the build logs for docker build phase completions) but neither was particularly attractive. The parsing logic for the latter was far more complex (comparing to this structured form) and the former was difficult to integrate with docker build (each step is independent so a server needs to be brought up in each or mapped into the environment. The server-based option also requires GCB either a) emit timing info into the build step logs or b) write out a file which is separately uploaded. The latter is fine but, if we want to support uploading when build phases fail, we'd need some way of actively recovering from and then recording those failures.

With the docker history, the record and integrity of the timing is implicit as is the progressive record. Provided we keep the layer tracking in-sync, we get a portable, consistent record of each phase without any changes to the build process itself.

PR #1400, opened 2026-07-29
Record measured costs on rebuild attempts

msuozzo:
Add AttemptCosts to the attempt record and assemble it in the finish
path from data already loaded there: the verdict's phase timings and
the BuildInfo GCB wall-clock, tagged with the pool the size hint routes
to. Nil when nothing measurable was recorded.


wbxyz (approved):
Love it. Beyond measuring individual attempts I'm excited to do an aggregate measurement across all the versions in a time window. That way we can capture the cost of frequent releases when attempting rebuilds.

PR #1402, opened 2026-07-29
Populate rebuild timings

msuozzo:


internal/api/apiservice/rebuild.go
```diff
@@ -236,6 +240,7 @@ func buildAndAttest(ctx context.Context, deps *RebuildPackageDeps, mux rebuild.R
 		UseNetworkProxy:    useProxy,
 		UseSyscallMonitor:  useSyscallMonitor,
 		SaveContainerImage: true,
+		RecordTimings:      true,
```
wbxyz, line 243: Can we make this configurable in flags to the api service?

internal/api/apiservice/rebuild_lro.go
```diff
@@ -63,6 +64,17 @@ func ProjectRebuildAttempt(a schema.RebuildAttempt) longrunning.Operation[schema
 	return op
 }
 
+// verdictTimings reassembles the verdict's aggregate from its stored parts:
+// build phases from the attempt record, infer duration from costs.
+func verdictTimings(a schema.RebuildAttempt) rebuild.Timings {
```
wbxyz: Should this be a receiver method on schema.RebuildAttempt called "Timings"? That way it's just `a.Timings()`.

Either way, I don't think "verdict" is the right word, I'd expect that to be referring to schema.Verdict.

pkg/rebuild/rebuild/models.go
```diff
@@ -113,11 +110,3 @@ type PrebuildConfig struct {
...
 	Auth   bool   `json:"auth,omitempty"`
 }
-
-// Verdict is the result of a single rebuild attempt.
-type Verdict struct {
-	Target   Target
-	Message  string
-	Strategy Strategy
-	Timings  Timings
-}
```
wbxyz, line 123: Is this an unrelated dead code cleanup?

I assume schema.Verdict is still being kept around.

wbxyz (approved):

pkg/rebuild/rebuild/models.go
msuozzo: yeah i couldn't figure out why this existed since it appears unused

internal/api/apiservice/rebuild_lro.go
msuozzo: yep done.

internal/api/apiservice/rebuild.go
msuozzo: i can do this in a follow-on. we could potentially expose the container image logic as well. it needs to be threaded through the tf, server, deps, and handler and even then we wouldn't expect to be flipping the flags especially often.

PR #1404, opened 2026-07-30
Populate phase timings in docker run-like executors

msuozzo:


pkg/build/local/run_executor.go
```diff
@@ -233,17 +234,31 @@ func (e *DockerRunExecutor) executeBuild(ctx context.Context, handle *localHandl
 	// Start the container idling, then execute each phase in it. The start
 	// output is the container ID, not build output: keep it out of the logs.
 	idBuf := &bytes.Buffer{}
+	// Setup's span opens before the container start so it absorbs the image
+	// pull and container create, mirroring the build executor's buildx clock.
+	phaseStart := time.Now()
 	buildErr := e.cmdExecutor.Execute(ctx, CommandOptions{Output: idBuf}, e.dockerCmd, ComposeDockerStartArgs(plan, argOpts)...)
+	in := rebuild.BuildTimings{}
```
wbxyz: nit: Why is this called `in`?

wbxyz (approved):

pkg/build/local/run_executor.go
msuozzo: ah yeah this is from a previous name for the struct. fixed

PR #1405, opened 2026-07-30
Add scratch build backend to agent

msuozzo:


internal/api/apiservice/agent.go
```diff
@@ -62,6 +68,13 @@ func executionFromOp(op *run.GoogleLongrunningOperation) (string, error) {
 }
 
 func AgentCreate(ctx context.Context, req schema.AgentCreateRequest, deps *AgentCreateDeps) (*schema.AgentCreateResponse, error) {
+	executionMode := req.ExecutionMode
+	if executionMode == "" {
+		executionMode = schema.AgentExecutionModeGCB
+	}
+	if executionMode == schema.AgentExecutionModeScratch && deps.ScratchCreateStub == nil {
+		return nil, api.AsStatus(codes.FailedPrecondition, errors.New("scratch execution mode is not configured"))
+	}
```
wbxyz, line 77: nit: might be worth just unconditionally doing scratch execution for agents

wbxyz (approved):

internal/api/apiservice/agent.go
msuozzo: it's a thought. don't need to decide yet (especially as this code is still settling) but I think it might be a good default for the future, whether we remove the gcb-only mode or not.

PR #1408, opened 2026-07-30
Add agent task mode plumbing

msuozzo:


wbxyz:
What about using different sub-commands on the agent binary for different tasks (rather than plumbing the value all the way into session.go)? 

Notably, the driving loop and evaluation at the end can be defined at a high level, re-using tools and utilities across different agents.

msuozzo:
yeah actually i think we can make do for this case without introducing a new task type so i'm going to punt on it for a moment. i think there are a still a few interesting cases where we'll want a fundamentally different loop so we'll hold off until those get implemented.

PR #1409, opened 2026-07-30
Add a docker build variant for pkg/build/scratch

msuozzo:


pkg/build/scratch/build_executor.go
```diff
@@ -0,0 +1,242 @@
...
+	dir := path.Join(e.workDir, buildsSubdir, handle.id)
+	container := "rb-" + handle.id
+	// Image references must be lowercase, unlike container names.
+	imageTag := "rb-" + strings.ToLower(handle.id)
+	// The planner declares layers only under Options.RecordTimings, so the
+	// declaration is the observation gate. The container is retained when
+	// observed so its state clocks survive for inspection.
+	recordTimings := plan.Layers.Appended > 0
+	autoRemove := !e.retainContainer && !recordTimings
+	if err := e.prepareBuild(ctx, dir, "docker images -q --filter 'reference=rb-*' | xargs -r docker rmi -f"); err != nil {
```
wbxyz, line 92: I think prepareBuild already has this line, so maybe it should be removed from executor.go?

And also did we want to drop the hard kill of existing containers? Or are we keeping the "at most one" concurrent invariant?

wbxyz (approved):

pkg/build/scratch/build_executor.go
msuozzo: this one's cleaning up past images while the unconditional one removes past containers (which both will have). removing them shouldn't impact caching so i think it's safe to keep.

PR #1410, opened 2026-07-30
Support partial and failure-marked build timings

msuozzo:


pkg/build/local/run_executor.go
```diff
@@ -242,22 +242,28 @@ func (e *DockerRunExecutor) executeBuild(ctx context.Context, handle *localHandl
 	if buildErr != nil {
 		buildErr = errors.Wrap(buildErr, "starting build container")
 	} else {
-		spans := map[string]*time.Duration{"setup": &elapsed.Setup, "source": &elapsed.Source, "deps": &elapsed.Deps, "build": &elapsed.Build}
+		slots := map[string]**time.Duration{"setup": &elapsed.Setup, "source": &elapsed.Source, "deps": &elapsed.Deps, "build": &elapsed.Build}
```
wbxyz: Should this be using the BuildPhase types instead of string constants?

wbxyz (approved):

pkg/build/local/run_executor.go
msuozzo: yep good call

PR #1412, opened 2026-07-31
Extract step logs from the merged GCB log

msuozzo:
GCB writes no per-step log objects, so ReadStepLogs was misconfigured to
attempt reads from per-step objects that didn't exist. Filter the step
section out of the merged log by its "Step #<n>" line prefix instead,
stripping the prefix so the section parsers see raw step output. Drop
the never-called ListStepLogs probe.

### DIFF
--- a/pkg/gcb/gcb.go
+++ b/pkg/gcb/gcb.go
@@ -4,12 +4,15 @@
 package gcb
 
 import (
+	"bufio"
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
 	"io"
 	"log"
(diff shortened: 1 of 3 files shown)


wbxyz (approved):
Would it be worth switching over to per-step logs? Then the thing that copies logs to the debug bucket can merge them if we want.

msuozzo:
> Would it be worth switching over to per-step logs? Then the thing that copies logs to the debug bucket can merge them if we want.

they actually come pre-merged in the GCS_ONLY logs setting so i think this is the best we can do natively.

PR #1414, opened 2026-08-01
Support GitLab subgroup repository URLs

MarlinDiary:
GitLab repositories can be nested under subgroups. The current pattern stops after two path segments, so nested project URLs resolve to the wrong repository.

Preserve the full GitLab project path while continuing to strip modern /-/ routes and legacy two-segment /tree/ routes.

Testing:
- `go test ./...`
- `go vet ./...`

### DIFF
--- a/internal/uri/uri.go
+++ b/internal/uri/uri.go
@@ -14,7 +14,7 @@ import (
 var (
 	// NOTE: This is non-exhaustive and should be expanded as necessary.
 	githubRE    = re.MustCompile(`(?i)\bgithub(\.com)?[:/]([\w-]+/[\w-\.]+)`)
-	gitlabRE    = re.MustCompile(`(?i)\bgitlab(\.com)?[:/]([\w-]+/[\w-\.]+)`)
+	gitlabRE    = re.MustCompile(`(?i)\bgitlab(\.com)?[:/]([\w-]+(?:/[\w.-]+)+)`)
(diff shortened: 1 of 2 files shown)


msuozzo (approved):

internal/uri/uri.go
```diff
@@ -25,6 +25,19 @@ var (
 
 var errUnsupportedRepo = errors.Errorf("unsupported repo type")
 
+func trimGitLabRoute(repo string) string {
+	if repo, _, found := strings.Cut(repo, "/-/"); found {
+		return repo
+	}
+	if repoRoot, _, found := strings.Cut(repo, "/tree/"); found {
```
msuozzo, line 32: from some brief research, it appears there were some more historical path segments permitted here (e.g. /raw/) although they would likely only show up via FindCommonRepo (pure text from READMEs). i think that's tolerable for now but i might add a follow-on here to expand out this case

internal/uri/uri.go
```diff
@@ -25,6 +25,19 @@ var (
...
+		pathStart := strings.IndexAny(repoRoot, ":/")
+		if pathStart >= 0 && strings.Count(repoRoot[pathStart+1:], "/") == 1 {
```
msuozzo, line 34: hmm this also looks like it wouldn't strip /tree/ from the group/subgroup/repo case you're interested in here, right? just the group/repo case?

this is still the common case but i think we'll need a generalization here, as well.

PR #1416, opened 2026-08-01
Try all matching tags during crates.io inference

MarlinDiary:
> Parents: [#1424](https://redirect.github.com/google/oss-rebuild/pull/1424)

#1424 ranks matching tags, but crates.io inference still validates only the first candidate. When that tag belongs to another package in the same repository, a later candidate can be valid.

Use #1424's ranking, then validate each ranked candidate until one contains the requested package and version. Existing callers of `FindTagMatch` keep the current first-match behavior.

Testing:
- `go test ./...`
- `go vet ./...`
- Replayed 1,605 affected crates.io versions: #1424's top-ranked candidate recovered 1,257/1,605; validating later candidates recovered 1,605/1,605.

### DIFF
--- a/pkg/rebuild/rebuild/git.go
+++ b/pkg/rebuild/rebuild/git.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"log"
 	"regexp"
+	"slices"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


pkg/rebuild/rebuild/git.go
```diff
@@ -65,18 +65,32 @@ func FindTagMatch(pkg, version string, repo *git.Repository) (commit string, err
 		if len(matches) > 1 {
 			log.Printf("Multiple tag matches [pkg=%s,ver=%s,matches=%v]\n", pkg, version, matches)
 		}
-		ref, err := repo.Tag(matches[0])
+	}
+	for _, match := range matches {
+		ref, err := repo.Tag(match)
```
msuozzo: ah so one tweak I'd make here is this new logic resolves, and fails on err, for each matched tag now for validity where it previously didn't. it's not a large change in behavior but it does change the error handling behavior for all other ecosystems in addition to rust.

pkg/rebuild/rebuild/git.go
```diff
@@ -42,12 +42,12 @@ func MatchTag(tag, pkg, version string) (strict bool, approx bool) {
 	return
 }
 
-// FindTagMatch searches a repositories tags for a possible version match and returns the commit hash.
-func FindTagMatch(pkg, version string, repo *git.Repository) (commit string, err error) {
+// FindTagMatches searches a repository's tags for possible version matches and returns their commit hashes.
+func FindTagMatches(pkg, version string, repo *git.Repository) (commits []string, err error) {
```
msuozzo: so i'm not opposed to this change but i played around with our match selection approach and got #1424 to ~95% accuracy (>98% for rust) so I think we could potentially opt for that approach versus this more structural change to inference.

what do you think?

MarlinDiary: I replayed #1424 against the same 1,605 crates.io cases used here. Its top-ranked tag recovers 1,257 (78%); validating the remaining candidates recovers all 1,605. The remaining misses include rank ties such as 0.2.1 versus v0.2.1, and cases where the correct tag has no clear package-name signal.

I may be missing a tradeoff, but would it make sense to use #1424 for ordering and keep the additional validation as a crates.io-only fallback?

msuozzo: ah that's fair. i had observed better results on my crates corpus but if it is sub-80%, i think it makes sense to claw that back if we can. i'd be fine with this change then but do you want to rebase this onto 1424 so we can land it after that?

pkg/rebuild/rebuild/git.go
MarlinDiary: Updated.

pkg/rebuild/rebuild/git.go
MarlinDiary: Done.

PR #1417, opened 2026-08-01
Use path_in_vcs during crates.io inference

MarlinDiary:
`.cargo_vcs_info.json` records both the commit and the package's repository-relative path. Inference currently uses the commit but ignores the path. If a commit contains more than one manifest with the requested package name and version, it can select the wrong directory.

Use `path_in_vcs` as the first directory checked for the recorded commit. Artifacts that do not contain this field keep the existing directory, and the current manifest search remains the fallback.

Testing:
- `go test ./...`
- `go vet ./...`
- Replayed 172 affected published versions: the packaged path was selected in 0/172 before and 172/172 after
- Replayed 5,000 unaffected published versions: 5,000/5,000 remained correct

### DIFF
--- a/pkg/rebuild/cratesio/infer.go
+++ b/pkg/rebuild/cratesio/infer.go
@@ -129,13 +129,17 @@ func inferRefAndDir(t rebuild.Target, vmeta *reg.CrateVersion, crateBytes []byte
 		cargoVCSGuess = info.GitInfo.SHA1
 	}
 	dir = rcfg.Dir
+	vcsDir := dir
+	if info.Dir != nil {
+		vcsDir = *info.Dir
+	}
 	var c *object.Commit
 	switch {
 	// Ensure the package config has the expected name and version.
 	case cargoVCSGuess != "":
 		c, err = rcfg.Repository.CommitObject(plumbing.NewHash(cargoVCSGuess))
 		if err == nil {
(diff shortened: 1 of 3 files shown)


pkg/rebuild/cratesio/infer.go
```diff
@@ -129,13 +129,17 @@ func inferRefAndDir(t rebuild.Target, vmeta *reg.CrateVersion, crateBytes []byte
 		cargoVCSGuess = info.GitInfo.SHA1
 	}
 	dir = rcfg.Dir
+	vcsDir := dir
```
msuozzo, line 132: Should we declare a new car here are reuse dir? I suspect this metadata is a reliable indicator for us to use generally, no?

msuozzo (approved):

pkg/rebuild/cratesio/infer.go
MarlinDiary: Yes. #1418 extends path_in_vcs to the tag and Cargo.toml history fallbacks. vcsDir only keeps this PR scoped to the VCS case. I’ll fold it into dir when rebasing #1418. findAndValidateCargoTOML still validates the path and falls back to a repository-wide search.

msuozzo: ah my mistake. i'd missed the newPath -> dir assignment in the `else` branch that does pick up the vcsDir assignment. i guess i was more just wondering why we wouldn't just assign it to `dir` directly but it's moot.

PR #1424, opened 2026-08-03
Add ranking to prioritize best tag match

msuozzo:
FindTagMatch sorts candidate tags lexicographically and takes the first.
This has many failure modes especially in monorepos where package
prefixes may result in an arbitrary and poor match. For example, grep
0.2.0 in BurntSushi/ripgrep matches tags 0.2.0, grep-0.2.0 and
grep-printer-0.2.0, and ripgrep's own 0.2.0 sorts first.

This new ranking prefers _more specific_ tags (e.g. grep-0.2.0) where
the non-version part still pertains to the package in question. It then
ranks partial package matches, then no package matches in that order.

Measured against known groundtruth commits across ecosystems (N=~2k),
this improves our performance for multi-tag matches from ~57% to ~95%.

### DIFF
--- a/pkg/rebuild/rebuild/git.go
+++ b/pkg/rebuild/rebuild/git.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"log"
 	"regexp"
+	"slices"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


pkg/rebuild/rebuild/git.go
```diff
@@ -43,6 +44,54 @@ func MatchTag(tag, pkg, version string) (strict bool, approx bool) {
...
 
+// tagAliases provides probable tag names for pkg accounting for e.g. npm @scope/, Maven group:artifact.
+func tagAliases(pkg string) []string {
+	aliases := []string{strings.ToLower(pkg)}
+	if i := strings.LastIndexAny(pkg, "/:"); i >= 0 && i+1 < len(pkg) {
+		aliases = append(aliases, strings.ToLower(pkg[i+1:]))
+	}
+	if t, ok := strings.CutPrefix(pkg, "@"); ok {
+		aliases = append(aliases, strings.ToLower(t))
+	}
```
wbxyz, line 55: This feels like it could be pushed into per-ecosystem logic, although that would be difficult to implement with recursive resolution.

pkg/rebuild/rebuild/git.go
```diff
@@ -43,6 +44,54 @@ func MatchTag(tag, pkg, version string) (strict bool, approx bool) {
...
+	r := strings.ToLower(strings.Trim(before+after, "-_/.@"))
+	return strings.Trim(strings.TrimSuffix(r, "v"), "-_/.@")
+}
+
+// tagRank measures how strongly a tag appears to relate to this package's release.
+func tagRank(tag, version string, aliases []string) int {
+	residue := tagResidue(tag, version)
+	if slices.Contains(aliases, residue) {
+		return 0 // contains package name alias AND version
```
wbxyz, line 73: It does not seem clear to me that this comment is correct.

If the version is missing from the tag, then tagResidue still returns the full tag. At which point slices.Containse(aliases, residue) just means the full tag is one of the aliases (only a match on the package, not the version).

pkg/rebuild/rebuild/git.go
wbxyz, line 103: How does MatchTag() fit with this new paradigm? It seems like maybe we could replace MatchTag with the logic in tagRank?

wbxyz (approved):

pkg/rebuild/rebuild/git.go
msuozzo: Yeah this simple thing worked surprisingly well across ecosystems tho. I agree the at sign is a bit to npm but it's slight enough in the scheme of this feature that I feel it's tolerable.

pkg/rebuild/rebuild/git.go
msuozzo: It's indirect but the input set of matches to this function already is filtered to those with the version present.

pkg/rebuild/rebuild/git.go
msuozzo: I think it ends up being equivalent, no? If there's one match, ranking won't change anything. If many, this logic kicks in.

PR #1428, opened 2026-08-04
Trim GitLab project routes by path segment

msuozzo:
The /tree/ special case only fired when exactly two segments preceded it, so
routes on subgroup projects and every other legacy route stayed in the path.
Match route names by segment position instead. Only segments past the mandatory
group/project prefix are considered, so a subgroup named for a route is
unaffected. The listed routes are the ones observed in published package
metadata; the wider set of GitLab routes never occurred.

Across 4,605 GitLab references in the crates.io corpus this trims 310 results,
most of them CI badge URLs, and changes none of the 3,522 declared repository
fields.

### DIFF
--- a/internal/uri/uri.go
+++ b/internal/uri/uri.go
@@ -6,6 +6,7 @@ package uri
 import (
 	"net/url"
 	re "regexp"
+	"slices"
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


internal/uri/uri.go
```diff
@@ -21,18 +22,30 @@ var (
...
+	}
+	pathStart := strings.IndexAny(repo, ":/")
+	if pathStart < 0 {
 		return repo
 	}
-	if repoRoot, _, found := strings.Cut(repo, "/tree/"); found {
-		pathStart := strings.IndexAny(repoRoot, ":/")
-		if pathStart >= 0 && strings.Count(repoRoot[pathStart+1:], "/") == 1 {
-			return repoRoot
+	segments := strings.Split(repo[pathStart+1:], "/")
```
wbxyz, line 45: Can this be a url parse and path.Split?

wbxyz (approved):

internal/uri/uri.go
msuozzo: path split is just the last element and these aren't (quite) urls so this is the best we can do, i think

PR #1432, opened 2026-08-04
Return untyped nil for successful operations

msuozzo:
Returning Operation.Error directly wraps a nil *OperationError in a
non-nil error interface, so successful ops read as failures. Guard
explicitly and return a literal nil.


wbxyz (approved):
Booo golang

PR #1435, opened 2026-08-06
Parse setup.py files to extract static metadata

msuozzo:
Ports #1011 to use bonsai to avoid cgo and rebases it to the current
mainline.

Adds a tree-sitter based setup.py analyzer that resolves the arguments of
setup() calls, and wires setup.py into build-dir discovery and requirement
extraction alongside pyproject.toml and setup.cfg.

The supported static data passed to setup() includes: string literals
(including raw, triple-quoted, and implicitly concatenated),
list/tuple/set displays, and module-level names bound to either.
Everything else resolves to the pyUnresolved sentinel that callers had
to string-match to detect.


msuozzo:
@gtystahl

pkg/rebuild/pypi/parsing/setup_parser.go
wbxyz, line 1: nit: The other config files have the full verify+extract implementations in a single file. Those other ones are very basic parsing, so there's not much beside verify+extract but it might be worth keeping the pattern of single file implementations for each type?

pkg/rebuild/pypi/parsing/setup_parser.go
```diff
@@ -0,0 +1,188 @@
...
+	src   []byte
+	vars  map[string]pyValue
+	calls []map[string]pyValue
+}
+
+// walk visits nodes in source order so that a name is bound before the setup()
+// calls that reference it.
+func (a *analyzer) walk(n *bonsai.Node) {
+	switch n.Type {
+	case "assignment":
```
wbxyz: Are there constants for these values?

wbxyz (approved):

pkg/rebuild/pypi/parsing/setup_parser.go
gtystahl: I don't think there was a strong reason to keep them separate, so I am good with combining them into one.

pkg/rebuild/pypi/parsing/setup_parser.go
gtystahl: There is definitely constants on the key side, but I did not check for specific values tied to them. Looking into the documentation it seems that all of them are general types like string, list, etc and not constant values that you can choose from. Here is the [documentation](https://setuptools.pypa.io/en/latest/references/keywords.html) related to that.

msuozzo: there aren't yet constants for node types. looking into adding them but, for now, they need to be addressed as raw strings.

pkg/rebuild/pypi/parsing/setup_parser.go
msuozzo: yeah i'm fine with combining them

gtystahl:
I did a small ablation study for this on a sample of 100 packages that I know use setup.py with and without these new features. These are the results:
### Without Setup.py Changes:
* 68 - build failed: executing build phase: exit status 1
* 16 - rebuild does not match upstream artifact
* **11 - success**
* 1 - clone failed
* 4 - timeout (> 30 minutes)
### With Setup.py Changes:
* **40 - success**
* 40 - rebuild does not match upstream artifact
* 18 - build failed: executing build phase: exit status 1
* 2 - timeout (> 30 minutes)

Overall, reduction of build errors from 68 to 18 (40%), boosted successful rebuilds from 11 to 40 (29%), and completed a build but not reproducible 16 to 40 (24%). 

This looks great should be merged and the old [PR](https://github.com/google/oss-rebuild/pull/1011) for setup.py should be retired.

gtystahl (approved):

PR #1453, opened 2026-08-07
Reject timewarp upstream server errors

MarlinDiary:
Cargo timewarp currently treats every unsuccessful upstream index response as a missing package. A server error can therefore produce an incomplete index archive instead of failing the request.

Keep HTTP 404 as package absence and return an error for other unsuccessful statuses.

Testing:
- `go test ./internal/timewarp`
- `go test ./...`
- `go vet ./...`

### DIFF
--- a/internal/timewarp/cargo_status_test.go
+++ b/internal/timewarp/cargo_status_test.go
@@ -1,0 +1,45 @@
+// Copyright 2026 Google LLC
+// SPDX-License-Identifier: Apache-2.0
+
+package timewarp
+
+import (
+	"bytes"
+	"io"
+	"net/http"
+	"testing"
+
+	"github.com/go-git/go-billy/v5/memfs"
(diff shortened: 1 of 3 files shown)


msuozzo (approved):

internal/timewarp/cargo_status_test.go
```diff
@@ -0,0 +1,45 @@
...
+	"bytes"
+	"io"
+	"net/http"
+	"testing"
+
+	"github.com/go-git/go-billy/v5/memfs"
+	"github.com/google/oss-rebuild/internal/httpx/httpxtest"
+)
+
+func TestAddPackageToIndexRejectsUpstreamServerError(t *testing.T) {
```
msuozzo: hmm shouldn't we be able to test this with the other timewarp fixtures?

MarlinDiary: Done — moved the 404 and 500 cases into `TestHandler_ServeHTTP`.

PR #1463, opened 2026-08-12
Add sessions to dashboard package view

msuozzo:


internal/api/dashboard/package.gohtml
```diff
@@ -7,18 +7,41 @@
...
-                <th>Attempt</th>
+                <th>Reference</th>
                 <th>Completed</th>
             </tr>
         </thead>
         <tbody>
-            {{range .PackageRebuilds}}
+            {{range .Events}}
+            {{if .Session}}
+            {{with .Session}}
```
wbxyz, line 22: Should these be a separate table or separate tab?

msuozzo: discussed and we'll break out separate tabs in a later change.

wbxyz (approved):

PR #1466, opened 2026-08-12
Add per-version drill-down page

msuozzo:
Adds /package/{ecosystem}/{package}/version/{version}: every rebuild attempt
for a single version alongside that version's agent sessions and their
iterations. The package view links to it in a follow-up change.

### DIFF
--- a/cmd/dashboard/main.go
+++ b/cmd/dashboard/main.go
@@ -112,6 +112,14 @@ func main() {
 			Package:   t.Package,
 		}, nil
 	}, api.HTMLHandler(DashboardInit, api.WithTimeout(30*time.Second, dashboard.Package), dashboard.PackageTmpl)))
+	http.HandleFunc("/package/{ecosystem}/{package}/version/{version}", api.Translate(func(r *http.Request) (dashboard.VersionRequest, error) {
+		t := encoding.New(rebuild.Ecosystem(r.PathValue("ecosystem")), r.PathValue("package"), r.PathValue("version"), "").Decode()
+		return dashboard.VersionRequest{
+			Ecosystem: string(t.Ecosystem),
+			Package:   t.Package,
+			Version:   t.Version,
+		}, nil
+	}, api.HTMLHandler(DashboardInit, api.WithTimeout(30*time.Second, dashboard.Version), dashboard.VersionTmpl)))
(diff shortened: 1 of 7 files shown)


cmd/dashboard/main.go
```diff
@@ -110,6 +112,14 @@ func main() {
 			Package:   t.Package,
 		}, nil
 	}, api.HTMLHandler(DashboardInit, api.WithTimeout(30*time.Second, dashboard.Package), dashboard.PackageTmpl)))
+	http.HandleFunc("/package/{ecosystem}/{package}/version/{version}", api.Translate(func(r *http.Request) (dashboard.VersionRequest, error) {
```
wbxyz, line 115: Shouldn't this just be "/package/{ecosystem}/{package}/{version}"?

We don't have mid-path constants anywhere else, just top level paths for different handler's routing (in which case this could maybe be /version/... instead of /package/...) or suffixes like `/attempt/...../build-logs/`.

internal/api/dashboard/dashboard.css
```diff
@@ -79,3 +79,13 @@ pre {
 .bg-green { background-color: var(--accent); color: #fff; }
 .bg-red { background-color: var(--mismatch); color: #fff; }
 .bg-gray { background-color: var(--fill); color: var(--muted); }
+
+/* Per-version drill-down: sessions and running status. */
```
wbxyz: If these are generic then we should remove this comment, if they're specific we should move this CSS to the html template for the version page.

internal/api/dashboard/version.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func (VersionRequest) Validate() error { return nil }
+
+// VersionSessionView pairs an agent session with its iterations for the
+// per-version drill-down.
+type VersionSessionView struct {
+	schema.AgentSession
+	Iterations []schema.AgentIteration
+}
```
wbxyz: This should probably be shared between the index, package, and version views.

internal/api/dashboard/dashboard.css
msuozzo: eh i think keeping the styles separate is tolerable

cmd/dashboard/main.go
msuozzo: yep fixed.

wbxyz (approved):

internal/api/dashboard/version.go
msuozzo: yep that's factored out later in the chain but i'll bring that definition here.

PR #1469, opened 2026-08-12
Include deleting and starting in reaper scope

wbxyz:


msuozzo (approved):

internal/db/scratch.go
```diff
@@ -82,27 +82,50 @@ func (f *firestoreScratch) UpdateLastUsed(ctx context.Context, scratchID string,
...
+
+	qReady := f.client.Collection(scratchCollection).
 		Where("state", "==", string(schema.ScratchReady)).
 		Where("last_used", "<", t.UTC())
+	if err := f.appendQuery(ctx, qReady, &out); err != nil {
+		return nil, err
+	}
+
+	qStarting := f.client.Collection(scratchCollection).
+		Where("state", "==", string(schema.ScratchStarting)).
```
msuozzo: i think you should be able to combine these as one query using "in" instead of "=="

you might even be able to combine all 3 into a single query

internal/db/scratch.go
```diff
@@ -82,27 +82,50 @@ func (f *firestoreScratch) UpdateLastUsed(ctx context.Context, scratchID string,
 }
 
 func (f *firestoreScratch) ListIdleSince(ctx context.Context, t time.Time) ([]schema.Scratch, error) {
```
msuozzo, line 85: you should also update memoryScratch to use this behavior

internal/db/scratch.go
```diff
@@ -82,27 +82,50 @@ func (f *firestoreScratch) UpdateLastUsed(ctx context.Context, scratchID string,
...
+
+	qStarting := f.client.Collection(scratchCollection).
+		Where("state", "==", string(schema.ScratchStarting)).
+		Where("updated", "<", t.UTC())
+	if err := f.appendQuery(ctx, qStarting, &out); err != nil {
+		return nil, err
+	}
+
+	qDeleting := f.client.Collection(scratchCollection).
+		Where("state", "==", string(schema.ScratchDeleting)).
```
msuozzo: this requires a new composite updates not added to the tf

internal/db/scratch.go
msuozzo: maybe we break out the stuck instances into a separate endpoint? "ListStuckSince"?

internal/api/agentapiservice/scratch_reap.go
```diff
@@ -112,10 +112,21 @@ func ScratchReap(ctx context.Context, _ ScratchReapRequest, deps *ScratchReapDep
...
 			log.Printf("reap re-check scratch %s: %v", scratch.ID, err)
 			continue
-		} else if cur.State != schema.ScratchReady || !cur.LastUsed.Before(idleCutoff) {
+		}
+		isIdle := false
+		switch cur.State {
+		case schema.ScratchReady:
+			isIdle = cur.LastUsed.Before(idleCutoff)
+		case schema.ScratchStarting:
+			isIdle = cur.Created.Before(idleCutoff) || cur.Updated.Before(idleCutoff)
```
msuozzo: created will always be before updated, right? (i think updated should always be set...?) i think we can drop the created check then.

msuozzo (approved):

internal/db/scratch.go
wbxyz: Updated

internal/db/scratch.go
wbxyz: Added

internal/api/agentapiservice/scratch_reap.go
wbxyz: Yeah I think updated is the right check.

internal/db/scratch.go
wbxyz: Do you want to converge them all on checking `updated` rather than `last_used`? I think maybe they should stay checking the separate values (starting and deleting aren't checking time since use, but time since the operation began).

wbxyz: I combined starting and deleted with "in" and added the composite index in the meantime.

PR #1470, opened 2026-08-12
Widen the package view's attempt query

msuozzo:
RecentPackageRebuilds returns the 100 most recent completed attempts,
which is too narrow to describe a package fully. We want older versions
and pending attempst that won't be represented in the query. This
changes the query to fetch all entries related to a package.

### DIFF
--- a/internal/api/dashboard/package.go
+++ b/internal/api/dashboard/package.go
@@ -17,6 +17,9 @@ import (
 
 var _ api.HandlerFn[PackageRequest, PackageData, *Deps] = Package
 
+// eventsWindow caps how many recent events the timeline renders.
+const eventsWindow = 100
+
 type PackageRequest struct {
 	Ecosystem string
 	Package   string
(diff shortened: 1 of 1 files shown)


internal/api/dashboard/package.go
```diff
@@ -40,8 +43,16 @@ type PackageData struct {
...
+	eco := rebuild.Ecosystem(req.Ecosystem)
+
+	// Fetch attempts across ALL versions of this package (include in-flight so
+	// running rebuilds are visible). FetchRebuilds with a version-less target is a
+	// collection-group query filtered by ecosystem+package.
+	rebuilds, err := deps.Rundex.FetchRebuilds(ctx, &rundex.FetchRebuildRequest{
+		Target: &rebuild.Target{Ecosystem: eco, Package: req.Package},
+		Opts:   rundex.FetchRebuildOpts{IncludePending: true},
+		Limit:  1000,
+	})
```
wbxyz: I think fetchRebuildsQuery should probably just be updated to include pending, RecentPackageRebuilds is only used for dashboard views so it's ok to push dashboard expecations in there, and a few other flavors of RecentX use fetchRebuildsQuery so it'd be good to standardize them

wbxyz (approved):

internal/api/dashboard/package.go
msuozzo: what if i do a treatment of this query code in a follow-on? i think there is value in keeping this on similar footing with other pages but there's enough blast radius in the underlying impl to be worth a separate change.

PR #1472, opened 2026-08-12
Add the version status strip to the package view

msuozzo:
Renders the per-version statuses as a grid of one cell per version, so a
package's shape is visible at a glance rather than by reading a table. The
strip shows one row by default and can be expanded to a bounded matrix, with
paging for packages whose history exceeds the window.

Cells link to the per-version drill-down.

### DIFF
--- a/cmd/dashboard/main.go
+++ b/cmd/dashboard/main.go
@@ -12,6 +12,7 @@ import (
 	"net/http"
 	"path/filepath"
 	"regexp"
+	"strconv"
 	"time"
 
 	"cloud.google.com/go/storage"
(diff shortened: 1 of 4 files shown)


wbxyz (approved):

cmd/dashboard/main.go
```diff
@@ -107,9 +108,12 @@ func main() {
 	http.HandleFunc("/", api.HTMLHandler(DashboardInit, api.WithTimeout(30*time.Second, dashboard.Index), dashboard.IndexTmpl))
 	http.HandleFunc("/package/{ecosystem}/{package}", api.Translate(func(r *http.Request) (dashboard.PackageRequest, error) {
 		t := encoding.New(rebuild.Ecosystem(r.PathValue("ecosystem")), r.PathValue("package"), "", "").Decode()
+		offset, _ := strconv.Atoi(r.URL.Query().Get("offset"))
```
wbxyz, line 112: nit: I think the query param should be more specific to the version table. I reasonable extension would be for the table of "recent" runs to be paginated as well, in which case "offset" would be ambiguous.

Maybe "vstrip_offset" and "vstrip_expanded" for the params?

wbxyz (approved):

cmd/dashboard/main.go
msuozzo: yeah i think this should really also not do a server round-trip, too. going to need some cleanup separately. added TODOs

PR #1480, opened 2026-08-18
Convert scratch VMs from Ubuntu to COS

msuozzo:
The Ubuntu scratch template apt-installed docker at boot which resulted
in a substantial delay to provisioning. Booting from a container on COS
drops this installation step by using the host COS's own docker daemon,
and a cloud-init-written systemd unit to run the worker. The worker
binary is fetched from the bootstrap-tools bucket the same as before.

COS mounts its writable paths noexec and docker binds inherit the source
mount's flags, so the worker binary is written to /var/lib/toolbox (the
exec-mounted exception) and /home/builder is remounted exec for the
build containers executing files from bind-mounts.

### DIFF
--- a/deployments/services.tf
+++ b/deployments/services.tf
@@ -52,9 +52,9 @@ data "google_compute_zones" "scratch" {
 }
 
 # Instance template for scratch VMs. The startup script fetches and runs the
-# scratch-worker binary; agent-api drives the worker over private-IP HTTP with
-# an ID token. Conditionally attach a service account for private instances to
-# access (the bootstrap-tools bucket is public-readable).
+# scratch-worker binary as a systemd service. agent-api drives the worker over
+# private-IP HTTP with an ID token. Conditionally attach a service account for
+# private instances to access bootstrap tools.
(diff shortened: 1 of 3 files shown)


deployments/services.tf
```diff
@@ -88,10 +88,12 @@ resource "google_compute_instance_template" "scratch-standard" {
   labels = { purpose = "scratch" }
 
   metadata = {
+    user-data = templatefile("${path.module}/scratch_cloudinit.yaml", {
+      cli_image = "docker:29.5-cli@sha256:11e1133c30f3ceb73c6bdc7dfb78b3f9ed8e8e0d1d0400e91c5ec2eb240bf2ff"
```
wbxyz, line 93: Is this an alpine base image? This will be a change in the "native" environment visible on the scratch client.

wbxyz (approved):

deployments/services.tf
msuozzo: added a note. we may want to reconsider this choice in the future so it's best to call it out explicitly.

PR #1493, opened 2026-08-20
Timewarp npm11 fix

gtystahl:
I noticed when running more recent packages that I was getting a lot more errors like this:
```
2026/08/18 20:26:15 error: unsupported platform  [/@babel/helper-simple-access/-/helper-simple-access-7.24.7.tgz]
2026/08/18 20:26:15 error: unsupported platform  [/@babel/helper-plugin-utils/-/helper-plugin-utils-7.24.7.tgz]
2026/08/18 20:26:15 error: unsupported platform  [/@babel/helper-module-imports/-/helper-module-imports-7.24.7.tgz]
2026/08/18 20:26:15 error: unsupported platform  [/@babel/helper-function-name/-/helper-function-name-7.24.7.tgz]
2026/08/18 20:26:15 error: unsupported platform  [/@babel/helper-hoist-variables/-/helper-hoist-variables-7.24.7.tgz]
2026/08/18 20:26:15 error: unsupported platform  [/@babel/generator/-/generator-7.24.7.tgz]
2026/08/18 20:26:15 error: unsupported platform  [/semver/-/semver-6.3.1.tgz]
2026/08/18 20:26:15 error: unsupported platform  [/@babel/core/-/core-7.12.9.tgz]
2026/08/18 20:26:15 error: unsupported platform  [/@babel/code-frame/-/code-frame-7.24.7.tgz]
2026/08/18 20:26:15 error: unsupported platform  [/@ampproject/remapping/-/remapping-2.3.0.tgz]
```

After talking with @wbxyz and @msuozzo we surmised it may have to do with timewarp since it seemed like this error matched the error output within it. 

I went and confirmed this, and through doing some testing found out it was because of from npm 10 to npm 11 there was a change (not seemed to be directly listed in the changelog from what I found) to how the authorization header is filtered when doing a request for a package during install time.

This patch fixes this by appending `--replace-registry-host=true` for installations using npm 11 and beyond. I tested it on the dataset I noticed it in and out of the 48 I was getting errors for, none of them are getting errors for this problem anymore. I will put the new breakdown with these changes below and include the sample benchmark I used if you want to check it out:
* 23 - Success
* 12 - Build failed: Docker run failed: exit status 1
* 5 - Rebuild does not match upstream artifact
* 4 - Build failed: docker run failed: exit status 127
* 4 - build failed: docker run failed: exit status 254

[npm_11_install_fails.json](https://github.com/user-attachments/files/31276647/npm_11_install_fails.json)


pkg/rebuild/npm/infer.go
```diff
@@ -301,6 +301,8 @@ func (Rebuilder) InferStrategy(ctx context.Context, t rebuild.Target, mux rebuil
 			}
 			if v, _ := semver.New(npmv); v.Major <= 6 { // NOTE: PickNPMVersion guarantees a valid semver
 				b.KeepRoot = true
+			} else if v.Major > 10 {
+				b.ReplaceRegistryHost = true
```
msuozzo, line 305: so one option would be to condition on the major version in the template itself instead of introducing a new top-level switch. i think we already have a compare function available in the template so we should be able to compare against a string with NPMVersion

msuozzo: added a commit!

msuozzo (approved):

PR #1498, opened 2026-08-24
Add dependency graph priority signal

msuozzo:
For onboarding, we want to define an order correlated with the security
relevance of a package to its ecosystem. The number of recursive
dependent packages is a good proxy for this and can be calculated from
deps.dev's dependency graph tables. We define this count as a package's
"criticality" to its ecosystem. Notably, while download counts and
repository statistics could serve a similar role, the dependency graph
is the only data uniformly available across all ecosystems we support.

It is relevant to calculate at two granularities: 1) A package in
aggregate and 2) a package at a particular version. The aggregate may be
used to judge a package's worth to all ecosystem participants and so
whether we might consider it worthwhile to track on an ongoing basis.
The per-version value is a useful prioritization within a given
package's history.

Scores are stored as per-ecosystem quantiles rather than raw counts, since
dependent counts are tail heavy (i.e. many with few dependents) and
values are not comparable across registries (e.g. npm has many more
total packages and thus higher dependents per package).

The ranked result is written as a JSONL export to a local path or a GCS
object rather than loaded into Firestore. The export is recomputed
wholesale on each run, so a bulk artifact matches the write pattern:
refreshing it is one atomic object rewrite, and consumers combine the
exported quantiles at read time.

### DIFF
--- a/tools/ctl/command/onboard/README.md
+++ b/tools/ctl/command/onboard/README.md
@@ -1,0 +1,36 @@
+# Onboarding
+
+Bringing a package into oss-rebuild coverage means two things: deciding it is
+worth the capacity, and then spending capacity until it reproduces.
+`ctl onboard priority` answers the first.
+
+## Priority
+
+There are far more packages than rebuild capacity, so something has to say
+which ones matter. Priority is a per-package score combining independent
+signals of importance, each computed by its own offline job and each
+normalized per ecosystem into (0,1] so heavy tails and incomparable registry
+scales stay out of the arithmetic.
+
+| Signal     | What it measures                                            | Source                    |
+| ---------- | ----------------------------------------------------------- | ------------------------- |
+| prevalence | distinct packages depending on this one, transitively too   | deps.dev dependency graph |
+
+Prevalence is computed at two granularities. The package number measures total
+use and is a coarse estimation of overall value. The per-version number allows
+us to better assess priority within a package's publications. Both are the
+dependent count on a log scale relative to the ecosystem's most depended-on
+package, so a version never outscores its package and the gap between them is
+the share of the package's use that version carries: lodash sits near the top
+of npm, lodash@4.17.21 just under it, and lodash@3.10.1 well below. The
+package number says lodash is worth tracking at all. The version numbers say
+which of its releases to rebuild first.
+
+Each job writes its ranked signal as a JSONL export, to a local path or a
+gs:// object:
+
+```sh
+# Read-only against public BigQuery data.
+ctl onboard priority prevalence --project my-project --out prevalence.jsonl
+ctl onboard priority prevalence --project my-project --out gs://my-analytics/priority/prevalence.jsonl
+```
--- a/internal/signals/doc.go
+++ b/internal/signals/doc.go
@@ -1,0 +1,16 @@
+// Copyright 2026 Google LLC
+// SPDX-License-Identifier: Apache-2.0
+
+// Package signals defines the priority signals that rank packages for
+// onboarding: per-package measures of importance, each normalized within its
+// ecosystem into (0,1] so heavy tails and incomparable registry scales stay
+// out of the arithmetic, and combined into one priority score at read time.
+//
+// "Prevalence" is the signal derived from the dependency graphs of each
+// ecosystem. We define it as "how many other packages depend on this one,
+// directly or transitively," measured both per package and per version.
+//
+// This package holds the record types, the scoring, and the exports' storage.
+// The jobs that produce and consume the exports live in
+// tools/ctl/command/onboard.
+package signals
--- a/internal/signals/prevalence.go
+++ b/internal/signals/prevalence.go
@@ -1,0 +1,39 @@
+// Copyright 2026 Google LLC
+// SPDX-License-Identifier: Apache-2.0
+
+package signals
+
+import (
+	"github.com/google/oss-rebuild/pkg/rebuild/rebuild"
+)
+
+// EcosystemSystem maps a local Ecosystem to its deps.dev `System`.
+// This allows querying against the enum in bigquery-public-data.deps_dev_v1.
+var EcosystemSystem = map[rebuild.Ecosystem]string{
+	rebuild.NPM:      "NPM",
+	rebuild.PyPI:     "PYPI",
+	rebuild.CratesIO: "CARGO",
+	rebuild.RubyGems: "RUBYGEMS",
+	rebuild.Maven:    "MAVEN",
+}
+
+// PrevalenceRecord contains the number of distinct packages that depend on a
+// package, directly or transitively, resolved against the latest deps.dev
+// dependency-graph snapshot, and that count scored against its ecosystem.
+// Version is omitted for package-level rows and set for version-level ones.
+//
+// Version granularity is pertinent because the most-depended-upon version is
+// rarely the newest and old versions may retain use (and, thus, value) well
+// past their publication. Ranking a package's versions by its
+// package-level prevalence would leave recency as the only differentiator,
+// which may not correlate strongly with blast radius.
+type PrevalenceRecord struct {
+	Ecosystem  string `json:"ecosystem"`
+	Package    string `json:"package"`
+	Version    string `json:"version,omitempty"`
+	Dependents int64  `json:"dependents"`
+	// Prevalence is Dependents on a log scale relative to the ecosystem's most
+	// depended-on package, in (0,1]. Counts are tail heavy enough that a rank
+	// quantile over the ecosystem would crowd every exported row above 0.99.
+	Prevalence float64 `json:"prevalence"`
+}
--- a/internal/signals/score.go
+++ b/internal/signals/score.go
@@ -1,0 +1,19 @@
+// Copyright 2026 Google LLC
+// SPDX-License-Identifier: Apache-2.0
+
+package signals
+
+import "math"
+
+// LogScale maps a dependent count onto (0,1] by its order of magnitude
+// relative to max, the ecosystem's most depended-on package:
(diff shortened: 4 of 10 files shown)


tools/ctl/command/onboard/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+There are far more packages than rebuild capacity, so something has to say
+which ones matter. Priority is a per-package score combining independent
+signals of importance, each computed by its own offline job and each exported
+as a per-ecosystem quantile so heavy tails and incomparable registry scales
+stay out of the arithmetic.
+
+| Signal      | What it measures                          | Source                    |
+| ----------- | ----------------------------------------- | ------------------------- |
+| criticality | distinct packages that depend on this one | deps.dev dependency graph |
```
pohlarized: I'm not sure if calling this criticality is ideal, since the [OSSF has a criticality score for packages](https://openssf.org/projects/criticality-score/) which I think might be a known concept for consumers of OSS Rebuild and its documentation, but our package criticality is measuring something different.
Potentially something like "prevalence" would be a less loaded term?

tools/ctl/command/onboard/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+Criticality is computed at two granularities. The package number measures total
+use and is a coarse estimation of overall value. The per-version number allows
+us to better assess priority within a package's publications.
```
pohlarized: I am not sure whether I really understand what these two numbers are and what it means that they both exist just from this explanation.
Perhaps worth adding an explained example e.g. from the tests here to explain that a bit more.

tools/ctl/command/onboard/criticality.go
```diff
@@ -0,0 +1,278 @@
...
+		return nil, errors.Wrap(err, "query failed")
+	}
+	it, err := job.Read(ctx)
+	return it, errors.Wrap(err, "reading results")
+}
+
+func criticalityCommand() *cobra.Command {
+	cfg := criticalityConfig{}
+	cmd := &cobra.Command{
+		Use:   "criticality --project <project> --out <uri>",
```
pohlarized: This is missing at least that `--ecosystems` exists and has to contain at least one ecosystem, and i would say it also makes sense to document the optional flags here.

Also, at least to me it was a little surprising that the deps.dev data is apparently accessed via BigQuery, and that the `--project` is required because it references a BigQuery project that is used to perform the queries on that data, but perhaps that's because I'm not too familiar with the deps.dev data.

tools/ctl/command/onboard/criticality.go
```diff
@@ -0,0 +1,278 @@
...
+	"google.golang.org/api/option"
+)
+
+type criticalityConfig struct {
+	Project     string
+	Ecosystems  string
+	Top         int
+	TopVersions int
+	Out         string
+}
```
pohlarized: I know I am just a big fan of documentation so it's fine if we establish that you don't want that much documentation here, but I think the names themselves (like `Top` or `TopVersions`) are not entirely self-explanatory, so I would propose to either give them longer but self-explanatory names (e.g. `MaxPackagesPerEcosystem` and `MaxPackageVersionsPerEcosystem` or to at least add doc comments to the fields so that it's easy for devs to find out what the fields mean.

tools/ctl/command/onboard/criticality_test.go
```diff
@@ -0,0 +1,67 @@
...
+		{Ecosystem: "", Package: "dropped-no-ecosystem", Dependents: 5},
+		{Ecosystem: "npm", Package: "", Dependents: 5},
+	})
+
+	want := []scheduler.CriticalityRecord{
+		{Ecosystem: "npm", Package: "lodash", Dependents: 9000, QCrit: 1.0},
+		{Ecosystem: "npm", Package: "rarely-used", Dependents: 1, QCrit: 0.5},
+		{Ecosystem: "rubygems", Package: "rack", Dependents: 50, QCrit: 1.0},
+		{Ecosystem: "npm", Package: "lodash", Version: "4.17.21", Dependents: 8000, QCrit: 1.0},
+		{Ecosystem: "npm", Package: "lodash", Version: "3.10.1", Dependents: 400, QCrit: 0.5},
```
pohlarized: Does the `QCrit` value of the versioned case only compare against other versioned rows in the input dataset?
As in, regarding QCrit calculation, entries with versions and entries without versions do not influence each other at all, only within the groups?

If so, for the versioned case we only have an example where all entries belong to the same package.
Say we had another row
```go
{Ecosystem: "npm", Package: "rarely-used", Version: "1.0.0" Dependents: 1},
```
as input, I guess we would now expect values `1` (lodash@4.17.21), `0.66` (lodash@3.10.1) and `0.33` (rarely-used@1.0.0) (roughly), right?

tools/ctl/command/onboard/criticality.go
```diff
@@ -0,0 +1,278 @@
...
+	if row.Time.IsZero() {
+		return time.Time{}, errors.New("no deps.dev snapshot found")
+	}
+	return row.Time, nil
+}
+
+// packageCriticality counts, for each package, the distinct packages that
+// depend on it at the given snapshot. Self-edges are excluded, as are
+// pseudo-packages (e.g. npm ">"-prefix).
+func packageCriticality(ctx context.Context, client *bigquery.Client, ecosystem, system string, snap time.Time, top int) ([]scheduler.CriticalityRecord, error) {
```
pohlarized: At least to me the difference between `ecosystem` and `system` was not quite clear from the doc-comment and signature here.

tools/ctl/command/onboard/criticality.go
```diff
@@ -0,0 +1,278 @@
...
+	}
+	if row.Time.IsZero() {
+		return time.Time{}, errors.New("no deps.dev snapshot found")
+	}
+	return row.Time, nil
+}
+
+// packageCriticality counts, for each package, the distinct packages that
+// depend on it at the given snapshot. Self-edges are excluded, as are
+// pseudo-packages (e.g. npm ">"-prefix).
```
pohlarized: What are these "pseudo-packages"? Are they something deps.dev specific that they are inserting, is this a general npm thing?

tools/ctl/command/onboard/criticality.go
```diff
@@ -0,0 +1,278 @@
...
+	out := make([]string, 0, len(m))
+	for k := range m {
+		out = append(out, k)
+	}
+	sort.Strings(out)
+	return out
+}
+
+func byDependents(recs []scheduler.CriticalityRecord) []scheduler.CriticalityRecord {
+	sort.SliceStable(recs, func(i, j int) bool { return recs[i].Dependents > recs[j].Dependents })
```
pohlarized: Do we need to sort stable here?

Also the docs have this note:
> Note: in many situations, the newer [slices.SortStableFunc](https://pkg.go.dev/slices#SortStableFunc) function is more ergonomic and runs faster

pohlarized (commented):
[We found](https://arxiv.org/pdf/2505.05897) that >50% of the top 1000 most depended upon npm packages are just "sybil packages" that inflate their own dependents.
I still think it is worth it to implement this as a signal, just wanted to raise awareness that this is kind of imprecise and already being abused a lot in the wild.

pohlarized:
This looks really good!
Mainly have some code style comments and understanding questions.

msuozzo:
> [We found](https://arxiv.org/pdf/2505.05897) that >50% of the top 1000 most depended upon npm packages are just "sybil packages" that inflate their own dependents. I still think it is worth it to implement this as a signal, just wanted to raise awareness that this is kind of imprecise and already being abused a lot in the wild.

very interesting. that finding appears to relate only to direct dependent counts while this is the transitive count which the paper finds is not as gamed. still a good thing to keep in mind.

tools/ctl/command/onboard/README.md
msuozzo: yeah that's an option. i did come across OSSF's criticality metric but felt it was distinct enough (esp. given this is an implementation detail of our tracking) but i'd consider changing to prevalence.

tools/ctl/command/onboard/criticality.go
msuozzo: eh bigquery use is always billed to a project so i think that aspect is unavoidable and is just gcp minutia.

and yeah this is missing the other flags.

tools/ctl/command/onboard/criticality.go
msuozzo: longer names, no. doc comments, yeah that would be useful (for reference, TopVersions was the last thing added here so it was simpler/clearer originally).

tools/ctl/command/onboard/criticality.go
msuozzo: yeah that's deps.dev's name for ecosystem. avoidable if we move the resolution into the function (although the awkwardness of two distinct names is relatively unavoidable).

tools/ctl/command/onboard/criticality.go
msuozzo: it's just package version ranges instead of packages themselves. not sure why the decision was made for this column but it's easy enough to ignore.

tools/ctl/command/onboard/README.md
msuozzo: done

tools/ctl/command/onboard/criticality.go
msuozzo: fixed the stability concern but we actually _can't_ guarantee input stability so we need to do so on the comparison itself. we do so by including the package/version values, in addition.

tools/ctl/command/onboard/criticality_test.go
msuozzo: done. yeah that's accurate

pohlarized (approved):
Looks good now 🍪

PR #1500, opened 2026-08-24
Add docdb, indexed SQLite views of document streams

msuozzo:
In order to provide better support and interoperability with local-first
workflows, this SQLite data layer provides flexible downstream query and
storage without requiring changes to the existing firestore schema and
usage. Updates can be streamed in periodically to ensure freshness and
full snapshots are made ~daily which ensures accuracy.


internal/docdb/cache.go
```diff
@@ -0,0 +1,284 @@
...
+	c.stop = cancel
+	go c.refreshLoop(bg)
+	return c, nil
+}
+
+// Query runs f with the current database, serialized against refreshes.
+func (c *Cache) Query(f func(*sqlite3.Conn) error) error {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return f(c.db)
```
wbxyz, line 88: Might be worth checking if c.db is nil before calling this?

internal/docdb/io.go
```diff
@@ -0,0 +1,256 @@
...
+	name := path.Join(r.prefix, relPath)
+	// ReadCompressed keeps gzip-encoded objects compressed on the wire (a
+	// no-op for plain objects). Consumers that need the decoded bytes
+	// detect the encoding themselves.
+	return r.client.Bucket(r.bucket).Object(name).ReadCompressed(true).NewReader(ctx)
+}
+
+func (r *gcsReader) List(ctx context.Context, prefix, start string) ([]string, error) {
+	full := path.Join(r.prefix, prefix)
+	q := &gcs.Query{Prefix: full + "/"}
```
wbxyz: If `r.prefix` and `prefix` are both the root or empty, this will create the invalid prefix of "/".

```
var q *gcs.Query
if full == "" {
	q = &gcs.Query{}
} else {
	q = &gcs.Query{Prefix: full + "/"}
}
```

internal/docdb/io.go
```diff
@@ -0,0 +1,256 @@
...
+	if start != "" {
+		q.StartOffset = path.Join(r.prefix, start)
+	}
+	it := r.client.Bucket(r.bucket).Objects(ctx, q)
+	var names []string // object listings are lexicographic by name
+	for {
+		attrs, err := it.Next()
+		if err == iterator.Done {
```
wbxyz: Does this need to skip attrs.Name that ends with "/"? Or maybe attrs.Prefix that's non-empty?

internal/docdb/segments.go
```diff
@@ -0,0 +1,187 @@
...
+	var recs []deltaRecord
+	sc := bufio.NewScanner(gz)
+	sc.Buffer(make([]byte, 1024*1024), 64*1024*1024)
+	for sc.Scan() {
+		line := strings.TrimSpace(sc.Text())
+		if line == "" {
+			continue
+		}
+		var rec deltaRecord
+		if err := json.Unmarshal([]byte(line), &rec); err != nil {
```
wbxyz: Leave `line` as bytes with `sc.Bytes()` then avoid casting it to []byte(line) in the unmarshal?

internal/docdb/meta.go
```diff
@@ -0,0 +1,85 @@
...
+	BuiltAt       time.Time // when the build began
+	Watermark     time.Time // every source write before this is captured, so replay resumes here
+	SchemaVersion int       // table layout generation, older readers refuse newer databases
+	SourceProject string    // the scanned source project
+	ToolVersion   string    // builder binary version
+}
+
+// WriteMeta replaces the database's meta row. Zero times store NULL so they
+// round-trip as zero.
+func WriteMeta(db *sqlite3.Conn, m Meta) error {
```
wbxyz: maybe wrap this whole thing in a sql transaction ensure this operation completes atomically (rather than deleting the snapshot_meta then aborting before replacing it.

wbxyz (approved):

internal/docdb/cache.go
msuozzo: meh it'd only be nil after Close but i suppose it couldn't hurt

internal/docdb/io.go
msuozzo: mooted by io refactor

internal/docdb/io.go
msuozzo: mooted by io refactor

internal/docdb/segments.go
msuozzo: yeah good call

internal/docdb/meta.go
msuozzo: this got refactored quite a bit so this should be addressed separately. the watermark update should be tracked in isolation now

PR #1502, opened 2026-08-25
Add a full database docdb SQLite database

msuozzo:
Scan Firestore into docdb doc tables (attempts, runs, agent sessions
and iterations, scratch VMs and execs, repo metrics) and write the
result as a single rebuild-v<N>.db, gzip-encoded where the destination
supports it. The registry in tables.go is the entire snapshot schema:
entity columns are extractions over the source documents (the strategy
oneof labels itself via json_each) so no row structs or Go
projection code exist anywhere. The database meta watermark records the
scan start so incremental replay can resume from it. It will run as a
scheduled job periodically (daily?).

The object name carries the schema version, so a version bump ensures
new consumers will require deployment instead of getting unexpected
data. And the schema safety is enforced with a golden test which pins
the materialized column shapes (generated columns included) against a
committed golden per SchemaVersion. Additive changes may regenerate it
in-place with -update, while removals, renames, retypes, and key changes
fail until the version is bumped, and -update itself refuses to rewrite
an incompatible golden.

In terms of mapping safety, a saturation test stores a fully populated
instance of every source type and asserts every column reads non-NULL
with its declared storage type, so an extraction path whose identity
departs from its struct (a renamed field, a newly added json tag) fails
rather than silently reading as empty data.


cmd/api/main.go
```diff
@@ -64,6 +68,7 @@ var (
 	agentLogsBucket       = flag.String("agent-logs-bucket", "", "GCS bucket for agent build logs")
 	agentTimeoutSeconds   = flag.Int("agent-timeout-seconds", 3600, "Seconds to allow agent to run")
 	rebuildJobName        = flag.String("rebuild-job-name", "", "Name of the pre-created Cloud Run Job for rebuilds")
+	analyticsURI          = flag.String("analytics-uri", "", "URI for snapshots (supported schemea: gs, file). Empty disables /snapshot/rollup")
```
wbxyz: Does this need to be set in terraform? And does the orchestrator need to be granted access?

internal/api/snapshotservice/rollup.go
```diff
@@ -0,0 +1,58 @@
...
+type RollupDeps struct {
+	Source snapshot.Source  // raw records comprising the snapshot
+	Dest   billy.Filesystem // snapshot destination. when nil, endpoint reports error
+	Opts   snapshot.Options
+}
+
+// Rollup scans the source and writes the snapshot database to the configured
+// analytics destination, replacing the published object wholesale.
+func Rollup(ctx context.Context, _ RollupRequest, deps *RollupDeps) (*RollupResponse, error) {
+	if deps.Dest == nil {
```
wbxyz: Also check for deps.Source being nil?

wbxyz (approved):

wbxyz (approved):

cmd/api/main.go
msuozzo: yep the terraform stuff gets set in #1506

internal/api/snapshotservice/rollup.go
msuozzo: eh this should be the same as Dest. could add as separate check here

PR #1504, opened 2026-08-26
Implement rundex for sqlite snapshot database

msuozzo:
Implement Reader and SessionReader over the snapshot. Notably, we reuse
the shared client-side pipeline (pending filter, verdict cleaning,
latest-per-package, limits) so semantics match the Firestore reader.
Access is serialized with Querier so the cache can swap the database
between calls.

### DIFF
--- a/internal/rundex/sqlite.go
+++ b/internal/rundex/sqlite.go
@@ -1,0 +1,300 @@
+// Copyright 2025 Google LLC
+// SPDX-License-Identifier: Apache-2.0
+
+package rundex
+
+import (
+	"context"
+	"encoding/json"
+	"slices"
+	"strconv"
+	"strings"
+	"sync"
+
+	"github.com/google/oss-rebuild/internal/sqlitex"
+	"github.com/google/oss-rebuild/pkg/feed"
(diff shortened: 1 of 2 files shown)


internal/rundex/sqlite.go
```diff
@@ -0,0 +1,300 @@
...
+func (c *cond) in(col string, vals []string) {
+	c.exprs = append(c.exprs, col+" IN (?"+strings.Repeat(", ?", len(vals)-1)+")")
+	c.args = append(c.args, vals...)
+}
+
+func (c *cond) clause() string {
+	if len(c.exprs) == 0 {
+		return ""
+	}
+	return " WHERE " + strings.Join(c.exprs, " AND ")
```
wbxyz, line 74: Sql injection?

wbxyz (approved):

internal/rundex/sqlite.go
msuozzo: all of these are assembled from string literals in the file so, as long as this invariant is maintained, this won't be much of an issue.

PR #1509, opened 2026-08-26
Support local snapshot database in benchmark runs

msuozzo:
run-bench -local -db <PATH> writes its run and verdicts into a snapshot
database on disk instead of the per-attempt file tree, creating the file
on first use, and rederives the trend tables when the run finishes. The
result is the same artifact the rollup publishes, so every snapshot
consumer (the dashboard sqlite rundex, ctl snapshot) reads it unchanged,
and because writes go through the guarded upserts, re-runs and replayed
upstream delta segments compose with local results safely. A database
from another schema era is refused: rebuilding from raw documents is the
only migration. Cross-process writers coordinate through SQLite file
locking.

### DIFF
--- a/internal/docdb/docdb_test.go
+++ b/internal/docdb/docdb_test.go
@@ -153,6 +153,18 @@ func TestQueryTable(t *testing.T) {
 	}
 }
 
+func TestEnsureDocTables(t *testing.T) {
+	db := newDB(t)
+	defs := []TableDef{probeTable, {Name: "message_counts", Query: "SELECT message, count(*) AS n FROM gen_probe GROUP BY message"}}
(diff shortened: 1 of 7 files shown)


wbxyz (approved):

wbxyz:
Should this abandon the existing rundex method? Might be worth asking existing users if they depend on the rundex output. Could also make a translation tool that exports the data from the sql database into the old format.

msuozzo:
> Should this abandon the existing rundex method? Might be worth asking existing users if they depend on the rundex output. Could also make a translation tool that exports the data from the sql database into the old format.

Yeah i think we should move towards this being the only option locally.

PR #1511, opened 2026-08-27
Add JSON lines helper

msuozzo:
jsonl encodes and decodes JSON Lines streams, one JSON object per
line, format-only: callers own the streams.

### DIFF
--- a/internal/jsonl/jsonl.go
+++ b/internal/jsonl/jsonl.go
@@ -1,0 +1,56 @@
+// Copyright 2026 Google LLC
+// SPDX-License-Identifier: Apache-2.0
+
+// Package jsonl decodes and encodes JSON Lines streams (one JSON value per line).
+package jsonl
+
+import (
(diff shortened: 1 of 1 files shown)


internal/jsonl/jsonl.go
```diff
@@ -0,0 +1,56 @@
+// Copyright 2026 Google LLC
+// SPDX-License-Identifier: Apache-2.0
+
+// Package jsonl decodes and encodes JSON Lines streams (one JSON object per line).
```
pohlarized: I believe at least in JSON lingo it's one JSON _value_ a line, not restricted to the JSON "object" datatype.

pohlarized (changes requested):

internal/jsonl/jsonl.go
msuozzo: done

pohlarized (approved):

PR #1512, opened 2026-08-27
Add a GCS billy.Filesystem and a URI resolver to billyx

msuozzo:
GCS serves one bucket prefix through billy's interface with an modest
capability subset: files are readable or writable but never both,
writers are sequential, directories are merely name prefixes, and there
are no temp files, symlinks, or locks. billy carries no contexts, so
operations run under the context the filesystem was constructed with.

However to improve upon this last point, we introduce Resolver which
does just-in-time mapping from locations to a filesystem that owns it,
local paths to the host filesystem and gs:// URIs to a bucket-rooted GCS
filesystem. The client is dialed on first gs:// resolution which binds
the fs to the context at first use (which is often what we're looking
for anyway).

### DIFF
--- a/internal/billyx/resolver.go
+++ b/internal/billyx/resolver.go
@@ -1,0 +1,55 @@
+// Copyright 2026 Google LLC
+// SPDX-License-Identifier: Apache-2.0
+
+package billyx
+
+import (
+	"context"
+	"path/filepath"
+	"strings"
+	"sync"
+
+	"cloud.google.com/go/storage"
+	"github.com/go-git/go-billy/v5"
+	"github.com/go-git/go-billy/v5/osfs"
+	"github.com/google/oss-rebuild/internal/gcsx"
+	"github.com/pkg/errors"
+)
+
+// Resolver maps a location to the local or GCS filesystem that owns it.
+type Resolver struct {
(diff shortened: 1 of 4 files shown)


internal/billyx/resolver.go
```diff
@@ -0,0 +1,50 @@
...
+
+// NewResolver returns a Resolver with nothing dialed.
+func NewResolver() *Resolver { return &Resolver{} }
+
+// ForURI returns the filesystem owning uri and uri's path within it.
+// For GCS, ctx governs GCS initialization and every subsequent operation on
+// the filesystem. file:// URIs and bare paths are local; relative paths
+// resolve against the working directory.
+func (r *Resolver) ForURI(ctx context.Context, uri string) (billy.Filesystem, string, error) {
+	if bucket, object, ok := strings.Cut(strings.TrimPrefix(uri, "gs://"), "/"); ok && strings.HasPrefix(uri, "gs://") {
```
wbxyz: Why not parse uri using the url lib?

wbxyz (approved):

internal/billyx/resolver.go
msuozzo: this should have used our existing parser. done

PR #1523, opened 2026-08-31
Conclude tool loops gracefully and raise the turn budget

msuozzo:
Hitting the tool-loop bound errored the whole turn which lost the
iteration and proposed nothing. The second-to-last turn now tells the
model the budget is gone so it concludes from what it has gathered, and
only a model that still insists on calling tools hits the exceeded
error.

Additionally, though, we should essentially never hit this constraint
since the agent's whole purpose is to explore and test configurations
interactively, meaning it should really only ever return with a proposal
once it's confirmed functional. As such, we raise it to 250 which,
functionally, shouldn't be reached.

### DIFF
--- a/internal/agent/session.go
+++ b/internal/agent/session.go
@@ -30,7 +30,7 @@ type AgentDeps struct {
 	MetadataBucket string
 	LogsBucket     string
 	GCSClient      *gcs.Client
-	MaxTurns       int
+	MaxTurns       int // Bounds each model exchange's tool uses (ChatOpts.MaxToolIterations)
 	GenaiClient    *genai.Client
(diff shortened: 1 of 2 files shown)


internal/agent/session.go
```diff
@@ -155,7 +155,7 @@ func doSession(ctx context.Context, req RunSessionReq, deps RunSessionDeps) (com
 		MetadataBucket: deps.MetadataBucket,
 		LogsBucket:     deps.LogsBucket,
 		GCSClient:      deps.GCSClient,
-		MaxTurns:       10,
```
wbxyz, line 158: yeah I think bumping max turns, or completely disabling it, is reasonable. In my experiments I've completely removed turn limits. The real concern is tokens, or CPU time. Turns is an awkward middle ground that measures neither.

wbxyz (approved):

PR #1525, opened 2026-08-31
api: expose /infer for inference-only benchmark runs

msuozzo:
run-bench's infer mode had a stubbed, unimplemented remote path and no
viable path that exposed the remote inference service. This change
exposes /infer from api so as to minimize changes necessary to
client-side configuration and IAM settings.


wbxyz:
This is to resolve this discussion? https://github.com/google/oss-rebuild/pull/1021#discussion_r2624833659

Seems like a reasonable solution.

wbxyz (approved):

PR #1528, opened 2026-09-01
pypi: dedupe repeats within build-system.requires

msuozzo:
Extract the name-keyed merge of metadata and build-file requirements
into mergeRequirements defining a helper for the name alone. Duplicates
inside build-system.requires are also collapsed

### DIFF
--- a/pkg/rebuild/pypi/infer.go
+++ b/pkg/rebuild/pypi/infer.go
@@ -267,6 +267,33 @@ func hasZipDir(dir string, zr *zip.Reader) bool {
 	return false
 }
 
+// requirementName extracts the distribution name from a PEP 508 requirement,
+// dropping any version specifier, so "setuptools<=67.7.2" yields "setuptools".
+func requirementName(req string) string {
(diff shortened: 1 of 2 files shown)


pohlarized (approved):
Do wheels even have any of the manifest files (setup.py / setup.cfg / pyproject.toml)? I thought they shouldn't carry them (and if there happens to be a file with that name in the wheel, i guess we're not too confident that it is actually the manifest describing the project but rather e.g. a test file?)

If they don't, then don't we always have exclusively either `reqs` or `buildReqs` populated, but never both? 
`reqs` is only populated for `.whl` files, which (I believe) doesn't have a manifest, so `buildReqs` would be empty.
And then `buildReqs` will only be populated for sdists, but there we explicitly skip populating reqs.

So I believe in the end we never actually get to merge anything?

If I'm wrong with that, the rest looks good to me 👍

msuozzo:
> Do wheels even have any of the manifest files (setup.py / setup.cfg / pyproject.toml)? I thought they shouldn't carry them (and if there happens to be a file with that name in the wheel, i guess we're not too confident that it is actually the manifest describing the project but rather e.g. a test file?)

ah this is pulling from the repo, not the whl, so we should have the manifest files available.

PR #1529, opened 2026-09-01
pypi: build with a modern PEP 517 frontend

msuozzo:
Install the build frontend from the real index before pointing pip at the
timewarp'd historical index. The frontend does not empirically impact
reproducibility. So the backend selection, defined by timewarp'd
requirement constraints, determines the output shape.

Notably, an older build package lacked --no-isolation (the -n short form
arrived in build 0.7.0), so timewarping the frontend failed pre-2021
packages. This approach mirrors npm fetching its modern CLI while still
timewarping dependencies.

Measured on a small representative sample of tracked package versions, this
raises the deterministic rebuild success rate from roughly 61% to 67% and lowers
the build-failure rate from roughly 30% to 23%.

### DIFF
--- a/pkg/rebuild/pypi/strategy.go
+++ b/pkg/rebuild/pypi/strategy.go
@@ -145,9 +145,8 @@ var toolkit = []*flow.Tool{
 		Name: "pypi/install-deps",
 		Steps: []flow.Step{{
 			Runs: textwrap.Dedent(`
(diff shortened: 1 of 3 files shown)


pohlarized (approved):
I'll have to note that I do not know this part of the code very well, so I cannot really certify that there is no additional context that requires changes.
However, the changes itself appear correct to me 👍

PR #1531, opened 2026-09-02
pypi: resolve refs from archive content when no tag matches

msuozzo:
A pure wheel built from a commit carries that commit's source files
verbatim, so their git blob hashes appear in its tree. When no tag
matches the version, hash the wheel's files and count how many each
commit's tree contains, through the gitscan blob hashing and tree scan
already used in maven inference. Blobs absent from the repository, such
as the generated dist-info files, drop out of the count. Only a unique
best match is accepted here.

Measured on a small representative sample of tracked package versions
with the tie-break stacked above, this resolves a ref for roughly 3% of
versions that otherwise have none, and over half of those then rebuild,
raising the deterministic rebuild success rate by ~2 percentage points.


pkg/rebuild/pypi/refsearch.go
```diff
@@ -0,0 +1,72 @@
...
+// matchArchiveBlobs returns the commit whose tree contains the most of the
+// given blobs, via gitscan.ExactTreeCount. Ties are ambiguous and rejected.
+func matchArchiveBlobs(ctx context.Context, hashes []plumbing.Hash, pkg, version string, repo *git.Repository) (string, error) {
+	closest, matched, total, err := gitscan.ExactTreeCount{}.Search(ctx, repo, hashes)
+	if err != nil {
+		return "", errors.Wrap(err, "searching trees for blob overlap")
+	}
+	if len(closest) != 1 {
+		return "", errors.Errorf("ambiguous blob overlap [ties=%d,best=%d,total=%d]", len(closest), matched, total)
+	}
```
pohlarized, line 65: Why is it a good idea to error here?
Even if it may be hard to tell which of these commits is the best, shouldn't we at least select any commit (potentially with logs if analysis of these cases is what you're after) and just try instead of straight up refusing?

I feel like in this situation, if we have matched the commit against the contents of an artifact from the release we want to rebuild, we have reasonably good chances that one of the commits is going to work.

pohlarized (approved):

pkg/rebuild/pypi/refsearch.go
pohlarized: Ah nvm, I see this is addressed in #1532 so it's fine to leave it breaking here.

PR #1532, opened 2026-09-02
pypi: break archive-match ties by declared version

msuozzo:
Wheels ship no build files, so a commit that only bumps the declared
version carries every blob its parent does and the archive match ties.
Break such ties with the version each candidate's build file declares,
read through the parsing verify parsers, which now surface the version
they find. Each candidate's build file is looked for where the previous
candidate's was found, then anywhere in its tree, so a package that has
moved directories is still read rather than assumed to sit where HEAD
has it. Prefer a candidate that confirms the version, drop one that
contradicts it, and among candidates that declare none take the
earliest, where the artifact content was introduced. Declared and
release versions are compared with the approximate version ordering,
since PyPI canonicalizes the spelling a build file uses.


pkg/rebuild/pypi/refsearch.go
```diff
@@ -54,19 +59,59 @@ func archiveContentRef(ctx context.Context, mux rebuild.RegistryMux, pkg, versio
...
 		return "", errors.Wrap(err, "searching trees for blob overlap")
 	}
-	if len(closest) != 1 {
-		return "", errors.Errorf("ambiguous blob overlap [ties=%d,best=%d,total=%d]", len(closest), matched, total)
+	best := make([]*object.Commit, 0, len(closest))
+	for _, h := range closest {
+		if c, err := repo.CommitObject(plumbing.NewHash(h)); err == nil {
+			best = append(best, c)
+		}
 	}
```
pohlarized, line 74: This may be somewhat nit picky but hopefully not bikeshedding territory, but I feel like `best` when `closest` exists doesn't convey the semantics of what this variable holds properly.
Maybe something like `existing` (or `existingCommits` but I guess that would be too long for your taste)?

I guess best is not technically wrong but very generic.

pkg/rebuild/pypi/refsearch.go
```diff
@@ -54,19 +59,59 @@ func archiveContentRef(ctx context.Context, mux rebuild.RegistryMux, pkg, versio
...
+	ref := pickArchiveCommit(ctx, best, pkg, version)
+	if ref == "" {
+		return "", errors.Errorf("no version-consistent blob overlap candidate [best=%d,total=%d]", matched, total)
 	}
-	ref := closest[0]
 	log.Printf("archive-content match [pkg=%s,ver=%s,blobs=%d/%d,ref=%s]\n", pkg, version, matched, total, shortHash(ref))
 	return ref, nil
 }
+
+// pickArchiveCommit breaks a blob-overlap tie by declared version: the most
```
pohlarized, line 83: I feel like this is a bit misleading.
Declared version becomes an additional requirement that has to be met by which we filter commits, and then the actual tie breaker is recency, isn't it?

pohlarized: Also why is it a good idea to reject commits solely based on the version? Do we consistently do this everywhere all the time?

I thought it would be a semi-common pattern that e.g. the CI changes the version field before release, so that the source repo commit that the artifact is actually built from doesn't have the correct version?
I would think that if we have a high blob overlap it's still worth trying to rebuild even if the version doesn't match.

pkg/rebuild/pypi/parsing/discovery.go
```diff
@@ -103,14 +104,24 @@ func normalizeName(name string) string {
 	return strings.ToLower(normalized)
 }
 
-// Recursively check for build files. Doesn't recurse if hintDir isn't empty.
+// findRecursively returns the build files named fileType: the one at hintDir
+// when given, else every one in the tree.
 func findRecursively(fileType string, tree *object.Tree, hintDir string) ([]*object.File, error) {
```
pohlarized, line 109: I feel like this should now rather be two different functions.
The only code path they share are lines 110-112, otherwise we have completely distinct execution paths depending on the condition whether `hintDir` is "" or not.

pkg/rebuild/pypi/parsing/resolver.go
```diff
@@ -46,17 +46,19 @@ func ExtractRequirements(ctx context.Context, tree *object.Tree, searchDir strin
 
 // DiscoverBuildDir searches for the best directory for requirement extraction.
 // Returns the directory path relative to the tree root, with "" representing root.
```
pohlarized, line 48: Looks like this doc comment was moved away from its function

pohlarized (commented):
Looks good, again mainly some comments on style but seems like everything is working as intended.

pkg/rebuild/pypi/refsearch.go
msuozzo: yeah this slipped through. "candidates" makes more sense

pkg/rebuild/pypi/parsing/resolver.go
msuozzo: oops!

pkg/rebuild/pypi/parsing/discovery.go
msuozzo: yeah this was the expedient option but i can change it

pkg/rebuild/pypi/refsearch.go
msuozzo: i recall this constraint improving our results (another change upstream applies the same constraint to tag commits) but i'll revisit this and loosen this check to a ranking (e.g. match -> unknown -> contradiction) instead of an exclusion if it proves effective.

spencerschrock (approved):

pkg/rebuild/pypi/refsearch.go
msuozzo: yeah to follow up on this: adding back contradicting versions gets 60 changes with 59 failures (wrong artifact version) and 1 success. so i think this is the better policy.

pohlarized: Yeah i guess rejecting it here leads to another method as fallback that works better than a commit with the wrong version 👍

PR #1545, opened 2026-09-02
db: move session and iteration writes onto Resource

msuozzo:
This adds internal/db orm entries for Sessions and iterations. New
Sessions and Iterations resources replace (most of) the direct Firestore
writes in the agent services and ctl local-agent, and a Mutate primitive
(transactional read-modify-write, returning false to skip the write)
absorbs the completion transaction. The iteration create keeps its
bespoke transaction pending a functional fix as the in-transaction query
emulates a uniqueness constraint on (session_id, number) that the
timestamp document key cannot express.


tools/ctl/command/localagent/localagent.go
```diff
@@ -129,13 +128,10 @@ func Handler(ctx context.Context, cfg Config, deps *Deps) (*act.NoOutput, error)
...
 		Updated:       sessionTime,
 	}
-	// Create session in Firestore
-	err = fire.RunTransaction(ctx, func(ctx context.Context, t *firestore.Transaction) error {
-		// NOTE: This would fail if the session record already exists.
-		return t.Create(fire.Collection("agent_sessions").Doc(sessionID), session)
-	})
+	sessions := db.NewFirestoreSessions(fire)
+	err = sessions.Insert(ctx, session)
 	if err != nil {
```
jonjohnsonjr: We could reasonably join these lines now that there's not a callback here.

internal/db/sessions_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+	"context"
+	"testing"
+
+	"github.com/google/oss-rebuild/pkg/rebuild/schema"
+)
+
+func TestMutate(t *testing.T) {
+	ctx := context.Background()
```
spencerschrock, line 14: t.Context()

internal/db/sessions_test.go
```diff
@@ -0,0 +1,67 @@
...
+	if got.Status != schema.AgentSessionStatusCompleted {
+		t.Errorf("skipped write changed Status to %s", got.Status)
+	}
+	if err := sessions.Mutate(ctx, "missing", func(*schema.AgentSession) (bool, error) { return true, nil }); err != ErrNotFound {
+		t.Errorf("Mutate(missing) = %v, want ErrNotFound", err)
+	}
+}
+
+func TestIterationsRoundTrip(t *testing.T) {
+	ctx := context.Background()
```
spencerschrock, line 54: same

spencerschrock (approved):

internal/db/sessions_test.go
msuozzo: will need to do a repo-wide cleanup in a follow-on

msuozzo: gonna do a repo-wide cleanup

internal/db/sessions_test.go
msuozzo: same

tools/ctl/command/localagent/localagent.go
msuozzo: done

PR #1546, opened 2026-09-02
db: migrate iteration creation to Resource.Mutate

msuozzo:
This change modifies the fully transactional cross-resource change into
a two-phase acquire txn where the Sesssion is reserved, then the
iteration written. This is no worse than the existing behavior and
enables us to fully remove the coupling to firestore.

The new procedure is:
1) AgentCreateIteration assigns a number server-side
2) Asserts Session is RUNNING
3) Asserts Session is within MaxIterations
4) Increments Session.IterationCount

Once that completes, we have 'acquired' the right to mint the interation
so we call Iterations.Insert.

The document key stays the server-chosen timestamp so we maintain the
invariant that the calling agent is not able to pick record identity,
sequence, or bounds. Additionally, this moves the iteration cap to be
enforced on the server instead of relying on the agent's self-restraint.


internal/agent/session_test.go
```diff
@@ -177,8 +177,8 @@ func TestDoIterationScratchSuccessConfirms(t *testing.T) {
 	if confirmed == nil {
 		t.Fatal("IterationStub not called for locally-verified success")
 	}
-	if confirmed.Strategy != strategy || confirmed.IterationNumber != 2 {
-		t.Errorf("confirmation request = %+v, want proposed strategy and iteration number 2", confirmed)
+	if confirmed.Strategy != strategy || confirmed.SessionID != "sess" {
+		t.Errorf("confirmation request = %+v, want proposed strategy for session sess", confirmed)
```
spencerschrock, line 181: intentional?

internal/api/agentapiservice/iteration.go
```diff
@@ -39,70 +35,75 @@ type AgentCreateIterationDeps struct {
...
+	default:
+		return session, api.AsStatus(codes.Internal, errors.Wrap(err, "reserving iteration"))
+	}
+}
+
+func AgentCreateIteration(ctx context.Context, req schema.AgentCreateIterationRequest, deps *AgentCreateIterationDeps) (*schema.AgentCreateIterationResponse, error) {
+	if req.SessionID == "" {
+		return nil, api.AsStatus(codes.InvalidArgument, errors.New("session_id required"))
 	}
-	// Extract strategy and use GCB executor to plan and execute build
```
spencerschrock, line 98: intentional?

internal/api/agentapiservice/iteration_test.go
```diff
@@ -0,0 +1,63 @@
...
+	"time"
+
+	"github.com/google/oss-rebuild/internal/db"
+	"github.com/google/oss-rebuild/pkg/rebuild/schema"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+func TestReserveIteration(t *testing.T) {
+	ctx := context.Background()
```
spencerschrock, line 18: t.context

spencerschrock (approved):

internal/agent/session_test.go
msuozzo: yeah iterationnumber was removed

internal/api/agentapiservice/iteration_test.go
msuozzo: yeah there are a ton of these. will clean up in a follow-on

internal/api/agentapiservice/iteration.go
msuozzo: eh yeah not sure that really adds anything

PR #1562, opened 2026-09-09
pypi: recognize the uv build backend

msuozzo:
Wheels stamped "Generator: uv X" come from uv_build, whose PyPI project
is uv-build and whose version tracks uv. Pin it the way the other
generators are pinned. uv_build finds its uv-build binary with a PATH
lookup, as an isolated PEP 517 build would arrange, rather than beside
the interpreter, so a build with the uv backend leads PATH with the venv
bin directory.

Measured on the uv versions the tracked set had failed on, roughly
60% rebuild once inferred.


pkg/rebuild/pypi/infer.go
```diff
@@ -438,6 +444,8 @@ var poetryPat = re.MustCompile(`^Generator: poetry ([\d\.]+)`)
 var poetryCorePat = re.MustCompile(`^Generator: poetry-core ([\d\.]+)`)
 var pdmBackendPat = re.MustCompile(`^Generator: pdm-backend \(([\d\.]+)\)`)
 
+var uvBuildPat = re.MustCompile(`^Generator: uv ([\d\.]+)`)
```
pohlarized, line 446: Just curious whether there's a pattern in the spacing between the different build patterns

pkg/rebuild/pypi/infer.go
```diff
@@ -269,6 +270,11 @@ func requirementName(req string) string {
 	return fields[0]
 }
 
+// pins reports whether the requirements pin any of the named packages.
+func pins(reqs []string, names ...string) bool {
+	return slices.ContainsFunc(reqs, func(r string) bool { return slices.Contains(names, requirementName(r)) })
+}
```
pohlarized: What does "pin" mean in this case?
To me, "pinning" a dependency means that you explicitly set a static version.
This function appears to just check whether the requirements contain any of the given package names.

I guess the implemented functionality works for what we want to achieve in this MR and we probably explicitly don't want to pin (as I defined it above), I was just confused by the wording here.

pkg/rebuild/pypi/strategy.go
```diff
@@ -61,6 +62,14 @@ func (b *PureWheelBuild) GenerateFor(t rebuild.Target, be rebuild.BuildEnv) (reb
 	return b.ToWorkflow().GenerateFor(t, be)
 }
 
+// venvOnPath marks a build whose backend shells out to a binary in the venv (e.g. uv-build).
+func venvOnPath(reqs []string) string {
```
pohlarized: I found it a little confusing that this function is called `venvOnPath`, as it seems like that is not describing what the function does, but what its consequence is within the `flow.Tool` steps when this function is called.
Something like `usesBinaryInVenv` might be more intuitive (this would also match with what the doc comment describes).

pohlarized (approved):
LGTM! Functionality wise it looks very good, I have just a few comments on understanding the code.

pkg/rebuild/pypi/infer.go
msuozzo: it's mainly so the above comment groups just the poetry ones

pkg/rebuild/pypi/infer.go
msuozzo: yeah "requirement" is probably better. updated

pkg/rebuild/pypi/strategy.go
msuozzo: yeah i think adding a "needs" prefix should help out here

PR #1564, opened 2026-09-09
pypi: build setuptools wheels with the tag the artifact names

msuozzo:
The target artifact name includes the python language tag which, for
non-py3, would be either universal (py2.py3) or Python 2 (py2).
setuptools decides which to apply from the [bdist_wheel] option.
Inference records the tag on the strategy when the artifact carries one
and the build backend is setuptools, and the build hands it to
setuptools through the user-level distutils config, which every
setuptools reads alongside setup.cfg.

Measured across the tracked PyPI package versions, this raises the
number of deterministic rebuild successes by about 1.5%.


pkg/rebuild/pypi/strategy.go
```diff
@@ -46,6 +47,7 @@ func (b *PureWheelBuild) ToWorkflow() *rebuild.WorkflowStrategy {
 				"dir":        b.Location.Dir,
 				"locator":    "/deps/bin/",
 				"venvOnPath": venvOnPath(b.Requirements),
+				"wheelTag":   b.PythonTag,
```
pohlarized: Why is it "PythonTag" in some places but "wheelTag" here?
If it's just an oversight, feel like "PythonTag" might be more fitting since wheels have multiple kinds of tags (unless you'd want to expand this single field to contain them all).

pohlarized (approved):

pkg/rebuild/pypi/strategy.go
msuozzo: the difference between the flow step and the strategy. i think i updated the strategy last but didn't update this so i'd be fine making them consistent.

pohlarized: Yes, if there is no specific reason that wheelTag carries a special meaning in the context of the flow step I believe that would be a good idea.

PR #1602, opened 2026-09-17
api: carry structured details on a status error

msuozzo:
A handler that wants to pass partial or additional data back in an error
case has no current way to pass it back that survives both stubs: a
typed Go error is lost over HTTP and a status message is only text.
Status details already cross both stubs, so Detail is a form-tagged
struct carried as an ErrorInfo under the oss-rebuild.dev domain with its
Reason as the key. AsDetail renders one for AsStatus and DetailOf
decodes it from a status anywhere in an error chain. AsStatus skips a
nil detail, which is what AsDetail returns for a value the form codec
cannot encode.

> Children: [#1603](https://redirect.github.com/google/oss-rebuild/pull/1603)


pkg/act/api/http.go
wbxyz, line 1: Do the details need to be written on `wr` in the `writeStatusError` function to make this travel across http connections?

wbxyz (approved):

PR #1623, opened 2026-09-19
proxy: enforce policy and log activity for non-80/443 CONNECTs

paracoding-ai:
## What

`NewTransparentProxyServer` installs TLS/HTTP interception (MITM) only for CONNECT targets on ports **80 and 443**. A CONNECT to any other port falls through to goproxy's default `ConnectAccept` action, which forwards raw bytes without producing an `OnRequest` event. Both the network policy (`ApplyNetworkPolicy`) and the activity log (`netlog.CaptureActivityLog`) are attached to `OnRequest`, so for such a connection:

- in `-policy_mode=enforce`, the policy is **not applied** — the connection is tunneled regardless of the allowlist; and
- the connection is **absent** from the network activity log that backs the `NetworkRebuild` attestation.

The second point is the one I'd flag most: a signed network record can certify a clean network profile while the build actually communicated over an unrecorded channel on a non-standard port.

## The change

- Add `TransparentProxyService.HandleUninterceptedConnect`, a catch-all CONNECT handler that (1) records the connection to the activity log and (2) in enforce mode applies the policy at **host granularity** — a raw tunnel exposes no HTTP path — and rejects targets the policy does not permit. It is registered in `cmd/proxy` **after** the 80/443 MITM matchers, so intercepted traffic keeps its existing handling (path-scoped rules still apply there) and only otherwise-raw tunnels reach it.
- Add `policy.Policy.AllowsHost` / `policy.URLMatchRule.AllowsHost` for the host-level check, refactoring `URLMatchRule.Allows` to share the host-matching logic (no behavior change to `Allows`).
- Fix `cmd/proxy/README.md`: the example policy JSON did not match the implemented schema (`anyOf`/`allOf` + `ruleType` + `URLMatchRule{host, matchHostBy, path, matchPathBy}`); and document the port scope under Limitations.

## Design note / open question

Raw CONNECT tunnels have no HTTP path, so this enforces at host granularity. Two reasonable alternatives, happy to switch if maintainers prefer:

1. **Host-level enforce (this PR):** allow a non-80/443 CONNECT iff a rule permits the host; record it. Keeps legitimate services on non-standard ports working while closing the fail-open.
2. **Fail-closed in enforce mode:** reject *every* non-intercepted CONNECT (log as blocked). Simpler and strictest, but breaks legitimate non-80/443 egress.

In the reference GCB deployment the sandbox iptables only DNAT 80/443 to the proxy, so non-80/443 traffic doesn't reach the proxy there anyway; this change is aimed at `cmd/proxy` adopters who route all traffic through it, and at making the activity log complete for whatever does reach the proxy.

## Testing

- `go build ./pkg/proxy/... ./cmd/proxy/...`, `go vet` — clean.
- `go test ./pkg/proxy/...` — existing tests pass.
- New tests:
  - `pkg/proxy/proxy`: `TestHandleUninterceptedConnect` (enforce rejects a non-allowlisted host on `:9443`, allows an allowlisted host on `:9443`, disabled mode tunnels anything; every case is recorded in the activity log) and `TestHandleUninterceptedConnectSkipLogging`.
  - `pkg/proxy/policy`: `TestAllowsHost` (anyOf/allOf, suffix vs partial-domain, path-ignored-at-host-level, port stripping).

I'm happy to iterate on the approach and to test any alternative patch.


paracoding-ai:
@googlebot I signed it!

